PR #37, opened 2015-06-08
README: libglm-dev is not installed as a dep

steveschnepp:


JonnyH:
Does the ubuntu libglm-dev work for you? I tested it on a freshly-updated 14.04 and it was too old (0.9.5) hence the documentation describing how to install the version from dependencies/glm

steveschnepp:
Well, I'm using debian 8...

JonnyH:
Feel free to add a Debian section to the readme. I only added Ubuntu under the assumption that the majority of Linux users seem to start there, and those that don't are more likely to be able to figure out what the instructions are trying to do and do the equivalent for their distro.

As an aside, I believe we need glm 0.9.6+ to allow implicit vector<int> to vector<unsigned int> conversions. It just happened to work on my PC as the gentoo glm package happens to be newer rather than a conscious decision. A long term fix would be to support the older glm by removing the implicit conversions, or adding the equivalent constructor to our Vec2/Vec3 wrappers.

pmprog:
Does this want to be merged then? What's glm used for?

JonnyH:
Not right now - this will likely break ubuntu 14.04. And glm is used to implement the Vec2<>/Vec3<> classes (and soon transforms/rotations/even more), and therefore used all over the place.

PR #42, opened 2015-06-10
German language strings

Blackwolf-Kuzoku:
German language strings has been added to STRINGS.XML and LANGUAGE.XML


pmprog:
Thanks

PR #66, opened 2015-11-26
Play music correctly.

sfalexrog:
The second array is actually an array of ending offsets in the MUSIC file.
This removes music cutting mid-playing as well.

Note that, as of now, playing the same track again will not work - looks like the MusicTrack does not get rewinded.


sfalexrog:
Now the tracks should "rewind" as well. I'm storing the starting track position and seekg()ing to it upon track end.

PR #67, opened 2016-04-08
update build doc for Ubuntu 14.04.3

jgatkinsn:
Made this into a local branch (feel free to delete the previous).  This is what I had to do to get it to build with latest Ubuntu 14.04


JonnyH:
Thanks

The only testing I do with ubuntu is the travis-ci, maybe I remember to change the build instructions every time I have to fix the build in that....

PR #68, opened 2016-05-04
Cuesheet parser/image reader

sfalexrog:
I'm not sure if that's ready to be merged, it *seems* to work on x86 Linux, Windows requires project files update, and I haven't tried it on anything else.

The archiver produces a lot of output, highlighting some of possibly problematic reading patterns (like readule* functions causing 512 byte reads no matter what).


pmprog:
Why do you think it's not ready? Also, IIRC, you've got permissions this to merge this

sfalexrog:
Merged as a squash commit.

PR #69, opened 2016-05-06
Add SMK video player, Rewrite SDLRaw sound backend

JonnyH:
BIG CHANGE: This now puts a hard requirement on Visual Studio 2015. The <chrono> usage in the video playback seems to be broken on 2013 - it seems to be missing the conversions between different units of time.

- Re-write the SDLRaw audio backend to support audio streaming
-- Some other small improvements at the same time (Moves audio conversion out of 'critical' realtime callback thread, store converted Samples instead of re-converting, call MusicTrack buffer filler in a ThreadPool task)
- Adds SMK video playing VideoScreen - used at startup to show the intro video
-- Has an async task, allows it to 'hide' loading - falls back to the same as the LoadingScreen logo & spinner if the video ends or is cancelled (ESC key) before the task is complete.


JonnyH:
@pmprog @sfalexrog @SupSuper 

Comments? MSVC2015 requirements? Should I shut up?

pmprog:
Sure, go with 2015 if you want. I've been thinking of installing that on my Win box so I can try the Xamarin Android stuff as well... Should I ever boot back into Windows (and if VS2015 runs on Win7)...

PR #72, opened 2016-07-02
Gameplay: vehicle altitude and attack mission

idshibanov:
- Altitude and attack modes should work now.
- Added basic vehicle attack mission, makes it follow the target.
- Fixed vehicle accuracy value.

Still in TODO: evasive maneuvering, make attack modes matter, fix weapon range calculations.


JonnyH:
Looks good to me! Thanks for your help.

PR #75, opened 2016-07-15
Save screen - gui

ShadowDancer:
Created new save/load screen.
Currently missing functionality to type save name or save game.

To test loading You need specific directory structure:
\saves\test.save
should contain `save_manifest.xml`:
```
<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
<save>
	<name>TestSave</name>
	<savedate>393923</savedate>
	<difficulty>SUPERHUMAN</difficulty>
</save>
```
and `save_data.zip` (save state spit out by gamestate)

After merging this you need to change
`--load-save=save` to `--load-save=save.zip`


game/state/savemanager.cpp
```diff
@@ -25,8 +24,8 @@ sp<Stage> SaveManager::loadGame(const SaveMetadata &metadata) const
 		}
 		state->initState();
 	});
-	return mksp<LoadingScreen>(std::move(loadTask),
-	                           [state]() -> sp<Stage> { return mksp<CityView>(state); });
+
+	return GameStateLoadTask(std::move(loadTask), state);
```
ShadowDancer: I don't really know what I did there.

framework/serialize.cpp
```diff
@@ -284,13 +312,11 @@ static bool readDir(const UString &name, std::map<UString, std::string> &content
...
 }
 
-static bool writePack(const UString &name, const std::map<UString, std::string> &contents)
+static bool writePack(const UString &path, const std::map<UString, std::string> &contents)
 {
 	mz_zip_archive archive;
 	memset(&archive, 0, sizeof(archive));
-
-	auto path = name + ".zip";
-
```
ShadowDancer: I need quick way of telling if file is saved game, so user do not have to wait for save screen, so I am detecting saves by .save extension.

game/state/savemanager.h
```diff
@@ -0,0 +1,96 @@
...
+#include "game/state/gamestate.h"
+#include <future>
+#include <tuple>
+
+class SerializationNode;
+class GameState;
+namespace OpenApoc
+{
+
+enum SaveType
```
JonnyH: Any chance we can prefer 'enum class' to keep the global namespace cleaner?

game/state/savemanager.h
```diff
@@ -0,0 +1,96 @@
...
+	             const sp<GameState> gameState);
+	SaveMetadata(const SaveMetadata &metdata, time_t creationDate, const sp<GameState> gameState);
+
+	/* Deserialize given manifest document	*/
+	bool deserializeManifest(const sp<SerializationArchive> archive, const UString &saveFileName);
+
+	bool serializeManifest(const sp<SerializationArchive> archive) const;
+
+	/* Creation date of save file (unix epoch)	*/
+	const time_t getCreationDate() const;
```
JonnyH: 'const' return types doesn't mean anything for non-pointer/reference - maybe remove that?

game/state/savemanager.h
```diff
@@ -0,0 +1,96 @@
...
+	/* Path to file containing (or that will contain) save	*/
+	const UString &getFile() const;
+
+	/* Name of difficulty	*/
+	const UString &getDifficulty() const;
+
+	/* What kind of save is it */
+	const SaveType &getType() const;
+
+	const unsigned int getGameTicks() const;
```
JonnyH: again const return for non-pointer/ref

framework/serialization/providers/filedataprovider.cpp
```diff
@@ -0,0 +1,47 @@
+#pragma once
```
JonnyH: #pragma once doesn't seem to make sense to me in cpp files (and gcc complains) - maybe remove these?

framework/serialization/serialize.cpp
```diff
@@ -0,0 +1,385 @@
+#define PUGIXML_NO_XPATH
+#define PUGIXML_HEADER_ONLY
+#include "dependencies/pugixml/src/pugixml.hpp"
```
JonnyH: As we're including the pugixml.hpp header lots of times with PUGIXML_HEADER_ONLY it probably makes sense to disable that and add pugixml.cpp to the build instead.

game/state/gamestate_serialize.cpp
```diff
@@ -814,6 +814,16 @@ template <> void serializeIn(const GameState *state, sp<SerializationNode> node,
...
 }
 
+template <> void serializeIn(const GameState *state, sp<SerializationNode> node, GameTime &t)
+{
+	if (!node)
+		return;
+
+	unsigned int ticks;
+	serializeIn(state, node->getNode("ticks"), ticks);
+	t = GameTime(ticks);
```
JonnyH: If the node doesn't actually have a "ticks" this will have an unset 'ticks' value - I guess you either want to set 'ticks' to a sane default, or use node->getNodeReq("ticks") if it should error if the node is missing

framework/framework.cpp
```diff
@@ -274,6 +281,8 @@ Framework::Framework(const UString programName, const std::vector<UString> cmdli
 
 	this->threadPool.reset(new ThreadPool(threadPoolSize));
 
+	LogWarning(
+	    ("Current working directory: " + boost::filesystem::current_path().string()).c_str());
```
JonnyH: This doesn't look right - a char[] + const char* won't do what you want - you probably want ("Current working directory: %s", boost::filesyste...whatever.c_str());

JonnyH: ...But it does happen to work in this case, as the const char[] at the beginning is constructed to a UString, then the operator+(const char*) called on that.

But anyway I prefer "%s" style for now, as we use it in most other places, and I like consistency.

Plus, GCC warns about non-const strings passed into printf strings (e.g if your path happened to contain a '%' printf specifier it'll probably do bad things

game/ui/general/savemenu.cpp
```diff
@@ -0,0 +1,304 @@
...
+}
+void SaveMenu::Pause() {}
+
+void SaveMenu::Resume() {}
+
+void SaveMenu::Finish() {}
+
+void ClearTextEdit(sp<TextEdit> textEdit)
+{
+	auto e = new OpenApoc::FormsEvent();
```
JonnyH: This leaks 'e', you probably want to wrap this in a unique_ptr<> (up<> in openapoc namespace)

game/ui/general/savemenu.cpp
```diff
@@ -0,0 +1,304 @@
...
+}
+
+void BeginEditing(sp<TextEdit> textEdit, sp<TextEdit> activeTextEdit)
+{
+	if (activeTextEdit != nullptr)
+	{
+		ClearTextEdit(activeTextEdit);
+	}
+
+	auto e = new OpenApoc::FormsEvent();
```
JonnyH: another 'e' leak

JonnyH:
After this it seems that trying to call getSection() not spits out an error:

virtual OpenApoc::sp<OpenApoc::SerializationNode> OpenApoc::XMLSerializationArchive::getRoot(const OpenApoc::UString&, const OpenApoc::UString&): Failed to parse "gamestate/vehicle_types.xml" : "No document element found" at "0"

instead of just returning nullptr? This seems to cause excessive error messages when loading a 'overlay' patch (e.g. running the common_patch importer in the extractor tool)

game/state/gametime.h
```diff
@@ -0,0 +1,48 @@
...
+class GameTime
+{
+  private:
+	unsigned long ticks;
+
+	bool dayPassedFlag;
+	bool weekPassedFlag;
+
+  public:
+	GameTime(unsigned long ticks) : ticks(ticks){};
```
JonnyH: You need to initialise the dayPassedFlag/weekPassedFlag here otherwise they're undefined

PR #80, opened 2016-08-19
Manufacturing almost done, and Research bugfixes v2

Istrebitel:


game/state/gamestate_serialize.cpp
```diff
@@ -1425,17 +1509,30 @@ template <> void serializeOut(sp<SerializationNode> node, const ItemDependency &
 
 template <> void serializeOut(sp<SerializationNode> node, const ResearchTopic &r)
 {
+	LogWarning("outputting topic %s", r.name.c_str());
```
JonnyH: If this is normal, this shouldn't be a warning. Either remove it if it was just temp debugging, or change to LogInfo() if you think it'll be useful to keep around

JonnyH:
Looks good, thanks again :)

game/state/gamestate_serialize.cpp
Istrebitel: I forgot to remove it. I was debug output

PR #81, opened 2016-08-22
Manufacturing done, basescape link to ufopaedia, ordered ResearchTopics

Istrebitel:


JonnyH:
Thanks again :D

We even manged to get git to do what we wanted for a change...

PR #82, opened 2016-08-26
Battlescape initial work, a display of a static map (58UFO8)

Istrebitel:
- read map and tile information from 58UFO8 folder
- display static battlescape map
- buttons working: layer buttons (up/down, numbers), layer mode (up to current, all layers, only current), strategic view
- available via ingame menu's button


JonnyH:
Thanks once again.

I cleaned up a few unnecessary files, and added the new stuff to the linux CMakeLists.txt too

PR #83, opened 2016-08-26
Battlescape now can load any map, fixed tiles trying to load nonexistant strategy images

Istrebitel:


JonnyH:
Thanks again.

I checked, and it *should* use PCKSTRAT for the strategy sprites
It ends up being the 'same' if the pck happens to start at x=0 for every line, but where it doesn't it does bad things and you get crashes and visual corruption. You have to look carefully to see it though :)

PR #84, opened 2016-08-28
Battlescape collision and voxel data fixed and working, battlescape tile selection, beginning work at agent data extraction

Istrebitel:


game/state/gamestate_serialize.cpp~RF3c74b548.TMP
```diff
@@ -0,0 +1,1840 @@
+#include "game/state/gamestate.h"
```
JonnyH: Another clang-format temporary file accidentally committed?

data/difficulty5_patched/gamestate/vehicle_types.xml
```diff
@@ -676,10 +676,10 @@
               <z>16</z>
             </size>
             <slices>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
+              <entry>LOFTEMPS:xcom3/UFODATA/LOFTEMPS.DAT:xcom3/UFODATA/LOFTEMPS.TAB:122</entry>
```
JonnyH: This seems to have changed, no idea why - regenerating this locally gives 'loftemps.DAT' like the previous commit?

Anyway, it makes sense to rewrite this commit without this, as it's nothing to do with this change and will pollute the history changing it back in another commit

data/difficulty3_patched/gamestate/cities/CITYMAP_HUMAN/tile_types.xml
```diff
@@ -3438,22 +3438,22 @@
...
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
-              <entry>LOFTEMPS:xcom3/UFODATA/loftemps.DAT:xcom3/UFODATA/loftemps.TAB:122</entry>
+              <entry>LOFTEMPS:xcom3/UFODATA/LOFTEMPS.DAT:xcom3/UFODATA/LOFTEMPS.TAB:122</entry>
```
JonnyH: It seems weird that only some difficulty's gamestate's have changed case here?

tools/extractors/extractors.cpp
```diff
@@ -35,12 +35,21 @@ void InitialGameStateExtractor::extract(GameState &state, Difficulty difficulty)
 	this->extractCityScenery(state, "ALIENMAP_", "alienmap", "alien", "stratmap", "loftemps",
 	                         "cityovr", state.cities[humanMapId]);
 
+	/*UString map = "08PORT";
```
JonnyH: If this is debug/testing code that doesn't need to be in the git history maybe remove it?

tools/extractors/main.cpp
```diff
@@ -60,6 +60,9 @@ int main(int argc, char *argv[])
 
 		std::list<std::future<void>> tasks;
 
+		// auto dpair = std::pair<UString,
```
JonnyH: More testing code? Maybe remove?

tools/extractors/extract_battlescape_stuff.cpp
```diff
@@ -89,6 +150,13 @@ static void readBattleMapParts(GameState &state, BattleMapPartType::Type type,
 		object->imageOffset = {24, 48};
 
 		state.battle.map_part_types[id] = object;
+
+		if (false) // debug output to get images into files
```
JonnyH: I guess this is for local-debugging, it probably won't be too useful for other users, so maybe remove before merging into git?

game/state/gamestate_serialize.cpp~RF3c74b548.TMP
Istrebitel: erm, what is "clang format termporary file" and how do I get rid of those?

JonnyH: I'm just saying this while .TMP file seems like it doesn't need to be committed, and possibly only there due to being left from a cancelled/crashed clang-format run, as I believe that uses similarly named temporary files.

You should be able to remove it by deleting the file, then it should show up as 'removed' in github, so it will be removed in the next commit.

If you fix this and the other things mentioned below in another commit, I can squash-merge in, so no need to worry about fancy git commands.

JonnyH:
Thanks, merged

PR #108, opened 2016-11-21
build: adding a brainless shell script

steveschnepp:
Using a brainless script is easier than reading README.md 😄


brainlessbuild.sh
```diff
@@ -0,0 +1,26 @@
+#! /bin/sh
+
+git submodule init
+git submodule update
+
+(
+	cd dependencies/physfs
```
JonnyH: We should no longer need a 'system' version of physfs, the cmake build system should build and link our patched version. If we still end up using the system version, that's a bug and needs to be fixed.

brainlessbuild.sh
```diff
@@ -0,0 +1,26 @@
...
+	cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo .
+	make
+	make install
+)
+
+(
+	cd dependencies/glm
```
JonnyH: It may be easier to add the local GLM header paths to cmake directly too, as it's just a header-only library it shouldn't need to install anything, just setup the include paths...

steveschnepp:
I added a `Vagrantfile` to be able to do a simple `vagrant up` to compile it :)

brainlessbuild.sh
steveschnepp: I don't really know how to do CMake hacking ...

steveschnepp:
So, now I managed to make it work flawlessly.

steveschnepp:
I did broke this into #110 and #111 .

So, as this script isn't useful anymore, i'll just close the PR.

PR #130, opened 2017-09-11
Relicense non-library (game/* tools/* and tests/*) under GPL3+

JonnyH:
As far as I can see, the only way of directly using this code is to
effectively implement an apoc clone. I don't see it being useful to pull
into other projects as library code, so I see a stricter copyleft
license as a better fit.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,9 @@
+The files under the "forms", "framework" and "library" directories are licensed under the MIT license.
+
+The All other files - the "game", "tests" and "tools" directories, are licensed under the GPL3+ license.
+
+------
+
 The MIT License (MIT)
 
 Copyright (c) 2014 Marq Watkin / Polymath Programming
@@ -19,3 +25,22 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
+
+------
+
+The GPL3 (or later) license (GPL3+)
+
+Copyright (C) 2017 "The OpenApoc Team"
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.


JonnyH:
OTHER DEVS - PLEASE COMMENT ON THOUGHTS!!!

If going forward we actually get a game released, I don't want others to take our implementation, rename it and (for example) sell it on the app store.

git log suggests the following devs have modified these directories:

Author: IstrebiteI <no.ads.starsabre@gmail.com>
Author: Istrebitel <starsabre@gmail.com>
Author: Ivan Shibanov <shibanov.ivan@gmail.com>
Author: Jonathan Hamilton <jtrhamilton@gmail.com>
Author: pmprog <devhead@pmprog.co.uk>
Author: Przemyslaw Onak <przemekonak@gmail.com>
Author: sfalexrog <sfalexrog@gmail.com>
Author: sparkstar <sparkstar@empal.com>
Author: SupSuper <supsuper@gmail.com>


I guess we'll either need sign off from all of them, or actually check their contributions to see if they're trivial and/or been replaced anyway?

SupSuper:
I don't see the point of complicating the licensing for the sake of being stricter. Personally I don't like GPL and anyone that wants to rip us off is gonna have to contend with the IP holders anyways.

JonnyH:
I think it's more I "default" to permissive for library code, more restrictive copyleft for "app-specific" code. I have no good specific reasons to relicense outside of me having seen dodgey android-app-store ripoffs that charge $$$ and inject spyware of other apps. But most of them wouldn't stick to a license anyway, as they're already breaking them and/or (ab)using the same names to squat searches and the like.

I wanted to poll other devs on their thoughts - so I guess that's a "no" from sup :)

sfalexrog:
I don't really understand all those licenses, and I'm fine with any decision. But I'd like to know what that would mean for the code that uses OpenApoc (say, a port that uses OpenApoc as a submodule, as it's done now) - does that mean that the port-specific code (think java parts of an Android port) should be GPL'd as well? (I'd be totally fine with that, those source files are unusually license-less :) )

JonnyH:
@sfalexrog Currently, MIT allows you to make your port closed-source. GPL would mean you have to make the code available to anyone you ship it to (e.g. have it on github).

And license-less source files are evil - 'no license' effectively means 'you have no rights to even /look/ at this file'.

hartmark:
The opposite is also true, MIT allows you to fork the repo and make it GPL3, however all previous commits will keep the MIT license also. Unless the previous commiter (copyrightholder) releases it as GPL3 instead.

Yeah, and having the license in top of all source files is also good practice.

SupSuper:
I believe GPL also conflicts with a lot of other restrictions like app stores, which is why SDL and such switched to MIT, so if you plan on ever putting OpenApoc on those platforms, something to think about.

As for the license preamble on code, I'm pretty sure it's more ancillary tradition than anything and there's nothing specially binding about it. If you do go that route please avoid cluttering the source with paragraphs of mumbo jumbo.

pmprog:
I do think MIT is okay for things like SDL, where it's a library designed to be reused.
The MIT licence in OpenApoc merely stemmed from my framework licence, which again is a library-esque.
As OpenApoc is a end product I'm quite happy for OpenApoc to get a more restrictive licence

Istrebitel:
I trust whatever you guys think. So you can write whatever non-commercial license you want under my commits. But I vote against putting license into every source file unless it's absolutely necessary.

PR #133, opened 2017-09-12
more fixes

Istrebitel:


Istrebitel:
Okay so I fixed the conflicts and it said "commit merge" as the only option. 
I didn't and I lost all changes (everything I resolved in conflicts)
So I fixed them again and clicked commit because there was no other way to save changes it seemed, only merge or not.

But it seems it didn't commit but actually saved my edits this time!

Is "commit merge" supposed to mean "apply conflict resolution changes"??

Anyways, as we discussed, not merging myself until everything checks out.

JonnyH:
@Istrebitel I pushed some travis changes on top (a linux debug build fix about Base::SIZE, and a few outstanding clang-format changes)

JonnyH:
One of the tests is failing intermittently - the serialization round trip - maybe there's something with a random order being serialized so it doesn't always match?

I'm not 100% sure on if "container order" is something needed to be preserved - as there are some gameplay situations where determinism requires the order to the kept, as stuff like the order in which update() ticks on gameobjects are calculated could affect behaviour (e.g. bullet A hits scenery, destroys it, allowing bullet B to pass through - if that's reversed bullet B could be the one that hits the scenery, allowing bullet A to continue)

PR #170, opened 2017-10-03
new stuff

Istrebitel:


Istrebitel:
- Options (notification toggles and togglable mouse scroll)
- Agent equipment fully implemented (multi-select, multi-operation)
- Fatal wound and Incoming/Outgoing Psi icons in battle

PR #193, opened 2017-10-17
economy, transfers, fixes

Istrebitel:
- economy
- base transfers (everything sans agents)
- alien containment screen
- ufo spawing at proper times
- vehicles advance on target if no los
- vehicles no longer bunch up in one tile occasionally when maneuvering
- vehicle chance to dodge projectile changes according to behavior (still not proper vanilla behavior - will implement when we know it)
- better functionality for scrollbars and buttons
- proper starting vehicle equipment, ammo and vehicles
- multiple fixes with cargo, vehicles and others


Istrebitel:
- economy
- base transfers (everything sans agents)
- alien containment screen
- ufo spawing at proper times
- vehicles advance on target if no los
- vehicles no longer bunch up in one tile occasionally when maneuvering
- vehicle chance to dodge projectile changes according to behavior (still not proper vanilla behavior - will implement when we know it)
- better functionality for scrollbars and buttons
- proper starting vehicle equipment, ammo and vehicles
- multiple fixes with cargo, vehicles and others

PR #212, opened 2017-10-24
Display item in transaction screen

FranciscoDA:
- Implements displayItem() in transactionscreen
 - Implements getWeight(), getAccuracy(), getArmor(), etc. in VehicleType (accepts an iterator pair of VEquipmentType to calculate the stats)


FranciscoDA:
I had to comment out the call to LogError at game/state/rules/city/vehicletype.h:203
I'd rather not leave it unchecked so if somebody can fix this before merging it'd be much appreciated.

game/state/rules/city/vehicletype.h
```diff
@@ -14,7 +14,6 @@
```
JonnyH: Adding
#include "framework/logger.h"

here fixes the LogError() issues below for me.

Istrebitel:
I merged this because git started going crazy already (it showed conflict with two identical code pieces that I had to resolve) and since Francisco confessed he's not very pro (as I am too), I think this would only get more and more confusing (as I remember it got for me when I started doing it) and I don't want to discourage him from contributing.

PR #217, opened 2017-10-26
Implement <subform> tag in forms

FranciscoDA:
Allows to reuse forms like: agent item view, vehicle item view, agent view, etc.

<subform> properties can override the style properties defined in the source form

It seems that the subform body captures hover and click events, so it will shadow controls that are under it (possibly an undesired effect?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,6 +93,8 @@ openapoc_log.txt
 version.h
 saves/
 
+# Vim swap files
+.*.swp
 
 # MSVC generated files end up in the source dir, we don't want to add them into git
 game/state/gamestate_serialize_generated.cpp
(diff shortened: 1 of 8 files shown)


JonnyH:
Looks good - maybe this can be used to cleanup the city/battlescape "tabs" too?

IE have the tabs as the main form, then <subform> for the "shared" parts - as we've had a number of issues due to the form "tree" not being strict subsets of the region defined by their parents, causing some events to be mis-delivered.

FranciscoDA:
I'm not familiar with that issue, but it does seem that mouse events are limited to the subform's bounding box. I think it should be possible to set the subform's position and size in the main form so that it won't overlap with other controls.

JonnyH:
Still it's an improvement, we can build on it further later,

Thanks!

PR #230, opened 2017-10-31
refactoring of RecruitScreen

redv:
Added ability to display agent's stats and skills

recruitscreen1

recruitscreen2

### DIFF
--- a/data/forms/recruitscreen.form
+++ b/data/forms/recruitscreen.form
@@ -153,7 +153,7 @@
         <size width="121" height="389"/>
         <orientation>vertical</orientation>
 		<item size = "32" spacing = "0"/>
-        <hovercolour r="120" g="120" b="120" a="255"/>
+        <hovercolour r="255" g="255" b="255" a="255"/>
       </listbox>
  
 	  <scroll id="LIST2_SCROLL">
@@ -177,7 +177,7 @@
         <size width="105" height="389"/>
         <orientation>vertical</orientation>
 		<item size = "32" spacing = "0"/>
-        <hovercolour r="120" g="120" b="120" a="255"/>
+        <hovercolour r="255" g="255" b="255" a="255"/>
       </listbox>
  
  
--- a/data/forms/recruitscreen.form
+++ b/data/forms/recruitscreen.form
@@ -152,7 +152,7 @@
         <position x="259" y="76"/>
         <size width="121" height="389"/>
         <orientation>vertical</orientation>
-		<item size = "32" spacing = "0"/>
+		<item size = "30" spacing = "2"/>
         <hovercolour r="255" g="255" b="255" a="255"/>
       </listbox>
  
@@ -176,7 +176,7 @@
         <position x="460" y="76"/>
         <size width="105" height="389"/>
         <orientation>vertical</orientation>
-		<item size = "32" spacing = "0"/>
+		<item size = "30" spacing = "2"/>
         <hovercolour r="255" g="255" b="255" a="255"/>
       </listbox>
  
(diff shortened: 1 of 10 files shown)


makus82:
nice =]

JonnyH:
Looks good! - fixup the small formatting/lint issues (https://travis-ci.org/OpenApoc/OpenApoc/jobs/295513050) and I'll merge

For future lint stuff, make sure you install clang version 4.0, and you can use the clang-format from that to automatically make your code fit the openapoc standard :)

redv:
I use ClangFormat from this source: https://marketplace.visualstudio.com/items?itemName=LLVMExtensions.ClangFormat
And I formatted whole file before last commit. As result, format changed in five parts of code that i even did not touch.

Please, look at this last failed build: https://travis-ci.org/OpenApoc/OpenApoc/jobs/295884881

The .clang-format in the project contains parameter BinPackArguments: true

BinPackArguments (bool)
If **false**, a function call’s arguments will either be all on the same line or will have one line each.
```
BinPackArguments: true
  f(aa, bb, cc);   // case 0
  f(aaaaaaaaaaaaaaaaaaaa, bbbbbbbbbbbbbbbb,
    cccccccccccccccccccccccccccccccccccccc);   // case 1

BinPackArguments: false
  f(aa, bb, cc);   // case 2
  f(aaaaaaaaaaaaaaaaaaaa,
    bbbbbbbbbbb,
    cccccccccccccccccccccccccccccccccccccc);   // case 3
```
__Point 1__
If the **BinPackArguments** set as **true** (as is), the code will be formatted as:
pic1
but if the **BinPackArguments** set as **false** then the code will be formatted as:
pic2

__Point 2__
If the **BinPackArguments** set as **true** (as is), the code will be formatted as:
pic3
but if the **BinPackArguments** set as **false** then the code will be formatted as:
pic4

__Point 3__
If the **BinPackArguments** set as **true** (as is), the code will be formatted as:
pic5
but if the **BinPackArguments** set as **false** then the code will be formatted as:
pic6

Is it possible that the travis-ci.org uses option **BinPackArguments: false** ? I dont think so.
Because of __Point 4__
The travis-ci.org offers even better match for the **BinPackArguments: true** option (looks exactly as case 1):
pic7

Whats conclusions?
1. Formatter which use on the travis-ci.org is bugged actually.
2. Formatters of different builds may lead to different results.
3. It is not so easy pass this check. Still, possible to copy result from the travis-ci.org's output.
Are you sure this checking is better than conventional code review?

JonnyH:
That looks like it's using a slightly different version of clang format.

We know that different versions produce slightly different output with the
same .clang-format config, so I decided to standardise on 4.0 (as that was
the most recent release at the time)

I do not know what version that visual studio extension uses, it says
6.0.something, and I don't know of that's just the revision of the
extension itself and not the clang-format version it uses, as I believed
that clang 6 was not released yet.

You can get the 4.0 revision of clang from here
http://releases.llvm.org/download.html

The patch revision shouldn't matter, so 4.0.1 is fine.

You can either use the clang-format.exe that's in that package directly, or
it may be possible to replace the clang-format version shipped with the
extension, depending on how it's integrated.

Hopefully that should then produce the same results.

I personally find it really annoying that clang doesn't produce the same
output over different versions with the same config. It makes it a pain to
try to setup different developers systems, causing this exact sort of
problem. But I feel that keeping a consistent codebase is worth the pain in
the long run.

redv:
Yes, version 4.0.1 allow to pass all checks.
Thanks.

JonnyH:
@redv I think you and @FranciscoDA have ended up working on some of the same things - any chance you can rebase against master again and see if the changes here are still valid or other fixes required?

FranciscoDA:
If you could reimplement the scientists' skill view and the highlight effect in the recruit screen that'd be great. You can find the code used to render agent stats in game/ui/general/agentsheet.cpp

redv:
done #236

PR #231, opened 2017-11-01
fix audio sample conversion issues in some systems

FranciscoDA:
also be more verbose about selected audio device ;)


JonnyH:
Looks good to me

PR #232, opened 2017-11-02
refactoring Researchscreen

redv:
- created two separate lists for small and large labs;
- added quantity number for lab icon;
- the state will remember selected lab for each base separately.

researchscreen1


redv:
- fixed deleting labs. Scientists returns to the unassigned list.
- fixed NewProject and CancelProject buttons behavior. Topic name updates properly.

PR #233, opened 2017-11-02
fixes vehicles being bought twice in transaction screen

FranciscoDA:
same for vehicle ammo


makus82:
https://github.com/OpenApoc/OpenApoc/issues/234

PR #236, opened 2017-11-03
refactoring Recruitscreen

redv:
- added skill label for scientists;
- added forms attribute "largechange" to scroll and "scrolllarge" to graphicbutton.

recruitscreen3


JonnyH:
BTW you can always update pull requests by force-pushing to the same branch - no need to close the old one and create a new one ;)

PR #238, opened 2017-11-06
refactoring Agentassignment

redv:
UPDATED

Changes:
1. Added the MultilistBox class.
- extends the Control class.
- allow to multiselect items.
- allow to build hierarchical trees. Hierarchical "hover" and "select" can works too.
- uses strategy pattern for "hover" and "select" rendering.
- uses strategy pattern to make a decision about visible/invisible items.
- uses strategy pattern for behaviour during selection actions.
- can be use either with ScrollBar or without ScrollBar.
- works only for vertical orientation (yet).
2. Several changes in the Control class.
- added SelectionSize to the Control class. It's an area which can be hovered and selected. The old Size is the area where the Control rendered. In most cases they're the same but not for the MultilistBox.
- changed meaning of the Control::preRender(). From now this method implements any operations other than graphical. It should call before the render() method if used controls which should adjust theyself just before rendering. I.e. if controls resize because of other controls.
- uses strategy pattern for ~~Control::update()~~ Control::preRender() performing.
3. Refactored the AgentAssignment.
- added drag&drop transfering between base and vehicle lists.
- generates the selected agents list for Raid and Extermination actions.
- generates the selected vehicle list (will be used for the Alert screen);
- click on agent's icons starts the agent's equip screen;
- click on vehicle's icons starts the vehicles screen.

Disadvantages:
1. Buildings icons is not vanilla. I don't know which image resources should be used.
2. ~~The right tree without scroll bar and not syncronised with the left tree.~~
3. The "Raid" action could not be started because cannot to load some image item.

Pictures (outdated):
1. Xcom base (has been opened by the agent button)

assignment10_base

2. On the fly (has been opened by the vehicle button)

assignment11_fly

3. Ready for raid (has been opened by the click on a building)

assignment12_raid

Outdated pictures:
1. Multiselection

assignment1_select

2. Drag

assignment2_move

3. Drop

assignment3_drop

4. Agents assigned to vehicle

assignment4_hover1

5. New selection

assignment5_new_selection

6. Drag to another vehicle

assignment6_move2

7. Drop and agents assigned to another vehicle

assignment7_drop2

### DIFF
--- a/data/forms/city/agentassignment.form
+++ b/data/forms/city/agentassignment.form
@@ -35,19 +35,19 @@
     </graphicbutton>
     <multilistbox id="AGENT_SELECT_BOX" scrollbarid="AGENT_SELECT_SCROLL">
         <position x="16" y="32"/>
-        <size width="228" height="328"/>
+        <size width="506" height="328"/>
         <item size="0" spacing="2"/>
         <hovercolour r="255" g="255" b="255" a="255"/>
         <selcolour r="128" g="180" b="255" a="255"/>
     </multilistbox>
 
-	<multilistbox id="VEHICLE_SELECT_BOX"> <!--scrollbarid="AGENT_SELECT_SCROLL"-->
+	<!--multilistbox id="VEHICLE_SELECT_BOX">
         <position x="271" y="32"/>
         <size width="228" height="328"/>
         <item size="0" spacing="2"/>
         <hovercolour r="255" g="255" b="255" a="255"/>
         <selcolour r="128" g="180" b="255" a="255"/>
-    </multilistbox>
+    </multilistbox-->
 
     <multilistbox id="DRAGGED_AGENT_BOX">
         <item size="0" spacing="2"/>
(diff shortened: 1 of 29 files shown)


makus82:
NICEEEEEE !!! Redv thats really good addon =]

makus82:
strange why vehicles list on top hover xcom logo ?
move little bit down =]

redv:
1. Yes :) but need optimization. At least one more commit.
2. It because of wrong size of XCom control or vehicle control. Will be fixed.

FranciscoDA:
Hey, great job!
BTW: it seems the travis (linux) build is failing because you forgot to add the new files in the following file: game/ui/CMakeLists.txt

redv:
Thanks
It will be fixed.

redv:
Updated.

makus82:
Отлично работает !
скролинг кстати в окнах не работает

redv:
I believe that original implementation is more complex. The left and the right trees should be placed inside one item. One item per building. These items should be inside a list which controlled by scroll bar on the right side.

redv:
Added:
- scrolling support;
- properly bases names;
- method which returns selected vehicles;
- tree for alert screen rendered properly.

assignment30_alert_and_names

makus82:
Chic! =D

redv:
thanks :)

Istrebitel:
This is so fucking cool! Great work

redv:
thanks :)

PR #258, opened 2017-11-23
Update Physfs to 3.0.1

JonnyH:
Adds cleanups, fixes some warnings, "official" iso9660 support allows us to drop some openapoc-specific hacks


JonnyH:
Dropped in favor of #275

Issue #253, opened 2017-11-22
Passenger Module does not add additional passengers to craft


FilmBoy84:
Tested with the Valkyrie and one passenger module, the capacity of the craft does not raise to 12 as it should, instead staying at 8.
The passenger module should add four passengers per module

FranciscoDA:
Whoops, I screwed up in a previous commit. Will fix ASAP

FilmBoy84:
This still appears bugged under 0.1-83.

Now any number of agents can be added to a craft, regardless of number of passenger modules!
For example i can currently fit all 15 of my agents in the Valkyrie with no extra modules.
Passengers are listed as 15/8! :D

redv:
Whoops! It will be fixed.

FilmBoy84:
Issue fixed by @redv
Closing

PR #270, opened 2017-11-26
fixed enterVehicle()

redv:
- fixed issue #253 "Now any number of agents can be added to a craft, regardless of number of passenger modules!"
- pixelperfection for AgentAssignment


data/forms/city/agentassignment.form
```diff
@@ -34,21 +34,13 @@
         <imagedepressed>BUTTON_SCROLL_DOWN_DEPRESSED</imagedepressed>
     </graphicbutton>
     <multilistbox id="AGENT_SELECT_BOX" scrollbarid="AGENT_SELECT_SCROLL">
-        <position x="16" y="32"/>
```
JonnyH: Was this change intended to be in this commit?

If it's fixing a different issue (e.g. misaligned form), it probably makes sense to have a separate commit

makus82: Hey Jonny offtopic but look pls i create PR wich readme)

Issue #254, opened 2017-11-22
Player does not start with correct vehicle equipment


FilmBoy84:
The Valkyrie should already have a cargo module and a passenger module at the start of play.
Currently this is not the case under OpenApoc.
Will add further vehicle specs to this as i find them.

Valkyrie In OpenApoc; missing Cargo Module and Passenger Module
![image](https://user-images.githubusercontent.com/33913088/33224503-ec1de8f0-d161-11e7-8fd0-ddb254bf18a4.png)

makus82:
check all good =]

FilmBoy84:
Excellent, Valkyrie does have a passenger module and cargo module now. Thanks @makus82 and @FranciscoDA

makus82:
So we should close this ?)

FilmBoy84:
Yep, close it, everything appears to start with the correct modules at the moment.

PR #276, opened 2017-11-27
set correct initial equipment for vehicles

FranciscoDA:
Now all the initial equipment for vehicles must be defined explicitly. Don't assume that the initial equipment is the same as the vehicle's default.
See issue #254


JonnyH:
I somewhat suspect there must be a table somewhere in the original .exe that lists this, that we *could* extract, but seems a rather large amount of effort for very little gain so I'm happy with this.

Issue #274, opened 2017-11-27
Game crash/use-after-free of Vehicle in Vehicle::update() with DepartToSpace mission


JonnyH:
The DepartToSpace mission calls Vehicle::die in the mission update function, which removes the vehicle object to be destroyed. This then returns into Vehicle::update(), which then continues to use the (now deleted) Vehicle object.

This can cause crashes, "unexpected" behaviour, and all kinda of weirdness. Plus, it makes valgrind/sanitizers unhappy.

PR #293, opened 2017-12-02
workaround for vehicle update & die

redv:
fixed issue #274 "Game crash/use-after-free of Vehicle in Vehicle::update() with DepartToSpace mission"

BTW need a refactoring actually


JonnyH:
A workaround is better than a crash - even if it's not particularly clean...

It may be a good idea to implement something similar for all state objects used by the Tile Engine - it would be interesting to test if the original destroyed objects instantly, or waited until the end of the tick.

As far as I can see, destroying stuff instantly could cause slight non-deterministic behavior, for example if two projectiles would hit an object in the same tick, the one that happened to be processed first could 'destroy' the object allowing the second to pass through the now-empty tile - and I don't like non-determinism.

If objects aren't destroyed until the end of the tick, /both/ would hit, removing both projectiles form play.

Probably not a massive deal - the average player likely wouldn't notice - but as I said I like determinism - it makes debugging easier :)

PR #296, opened 2017-12-03
Reduced the number of calculations for the cityview interpolated palette

redv:
- Reduced the number of calculations for interpolated palette.
- Also decreased the number of allocations/deallocations.

### DIFF
--- a/game/ui/tileview/citytileview.cpp
+++ b/game/ui/tileview/citytileview.cpp
@@ -67,6 +67,8 @@ CityTileView::CityTileView(TileMap &map, Vec3<int> isoTileSize, Vec2<int> stratT
 		mod_day_palette.push_back(newPal[0]);
 		mod_twilight_palette.push_back(newPal[1]);
 		mod_night_palette.push_back(newPal[2]);
+		mod_interpolated_palette.push_back(mksp<Palette>());
+		interpolated_palette_minute.push_back(0);
 	}
 
 	selectedTileImageBack = fw().data->loadImage("city/selected-citytile-back.png");
@@ -990,11 +992,11 @@ void CityTileView::update()
 	counter = (counter + 1) % COUNTER_MAX;
 
 	// Pulsate palette colors
-	colorCurrent += (colorForward ? 1 : -1);
+	colorCurrent += colorForward;
 	if (colorCurrent <= 0 || colorCurrent >= 15)
 	{
 		colorCurrent = clamp(colorCurrent, 0, 15);
-		colorForward = !colorForward;
+		colorForward = -colorForward;
 	}
 
 	// The palette fades from pal_03 at 3am to pal_02 at 6am then pal_01 at 9am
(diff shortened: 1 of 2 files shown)


JonnyH:
Possibly of interest:
https://github.com/JonnyH/OpenApoc/tree/WIP/PR-296-int-palette-interpolation

I had a quick play around, and it looks like gcc generates vectorised code at -O3, and it's actually about equally as fast between int/float if you enable avx2 (tested on a 6700hq).

At -O2 and below, integer stuff is a bit faster.

MSVC seems to fail completely at the float path, trying to unroll but failing to vectorise - possibly as it can't figure out if the pointers alias (gcc, for example, checks if the pointers overlap, falling back to the 'slow' version if they do, or declaring the input __restrict__ allows it to assume they don't overlap).

MSVC's __restrict keyword on the interpolate() inputs doesn't seem sufficient to workaround this, possibly as the docs suggest it's not propagated down, so we'll need a __restrict version of Palette::colours[] and Colour::r,g,b,a too?

Not compared directly to the previous openapoc version, but pulling out the get/set stuff and implementing it directly in the Palette class will also probably make the compiler's job easer.

redv:
Yes, avx works fast, but leads to increased power consumption. What may reduce cpu frequency. Especially for popular CPUs with "U" suffix. What may reduce overall performance for non-avx instructions and leads to cooler's noise (big issue for my notebook :)
https://computing.llnl.gov/tutorials/linux_clusters/intelAVXperformanceWhitePaper.pdf
> Workloads using Intel AVX instructions may reduce processor frequency

and
> Workloads that utilize a very high percentage of Intel AVX2 may operate closer to the AVX base frequency

I think, if possible to use integer calculations, then better to use integer calculations.

JonnyH:
I played around with a few integer implementations and vectorised examples here: https://gist.github.com/JonnyH/4e3a57d8ee5fb763b0430ab2cfcd5612

(Also fixed the integer implementations compared to my previous comment - sorry about that! Plus, it doesn't round but truncates, which matches the current implementation too)

My takeaway: On an i7-6700hq (laptop so if there was some throttling/power limit going on I'd expect it to be hit):
There's no advantage over any of the 3 integer implementations I could think of over the float version (1 was actually noticeably slower) without hand-vectorizing.

Hand-vectorising, however, gives a good speedup with an integer implementation with ssse3, processing 16 elements at a time gives over 2x speedup compared to the compiler-vectorized implementations.

32 elements at a time is even faster, nearly 2x over that, but required avx2. As this isn't really "dense" avx2 code, it likely hasn't hit the limits mentioned. Plus the fact that a palette interpolation takes time to the order of 80 nanoseconds it probably won't even measure on any performance/power usage tools as it will be lost in the noise.

redv:
Really interesting results.
Interpolate1 so slow just because of division.
I used to think that float based algorithms always slower because any separate float instruction slower than the same integer.

JonnyH:
The floating point stuff is normally done in a different pipeline, so long as you don't need the result before the latency it probably won't affect things - or even be faster if you're blocked on the integer pipeline (for address generation or other ops)

And most of the time it's more efficient to complete operations quicker and idle for longer than use a slower path that uses fewer chip resources, even if using those fewer resources causes it to draw less power, as the sum power used is normally power.

JonnyH:
FYI I just updated that gist above with a version of "vec16" that directly uses sse intrinsics (ssse3 required), allowing MSVC support. Should be pretty much the same speed as the vec16 it's based on, asm looks sane, but not actually compared runtime perf....

redv:
The integer version may be faster.
The `(num - factor)/num` and `factor/num` is just a proportions. If the `factor` can change from 0 till `num` then not a big difference what is the `num` exactly.
I.e. `num = 255`, then `factor = f * 255.0f` and `mix_unorm8 = (a * (255 - factor) + b * factor) / 255`
If `num = 345`, then `factor = f * 345.0f` and `mix_unorm8 = (a * (345 - factor) + b * factor) / 346`
Or if the `num = 1000`, then `factor = f * 1000.0f` and `mix_unorm8 = (a * (1000 - factor) + b * factor) / 1000`

But, obviously, most optimal option is `num = 256` and `factor = f * 256.0f`
and `mix_unorm8 = (a * (256 - factor) + b * factor) >> 8` as in the very first test case. Which should be a bit faster than other integer cases but, I think, cannot bit ssse and avx.

JonnyH:
I don't know how much of an advantage changing the scale would benefit - as already the compiler lowers the integer divide into a multiply and shift (and I did the same in the ssse3 intrinsic version) - there may be some opportunity to interleave more than 1 to get the best out of the CPU's pipelines and avoid stalls, but that's starting to get cpu-implementation specific, and *really* not worth the time for something that takes perhaps 1/80000th of a frame (assuming 60fps = 16ms, and it took ~20 ns for a single interpolation)

JonnyH:
We can push an optimized palette interpolate function as a separate change - I'll probably try to figure out what the best combination of (supported by compiler ) x (supported by processor) x (speed) is - probably a "base" integer version, an ssse3 intrinsics version for msvc, and the vec16 vector_size extension for clang/g++ (as it generates the same code as the ssse3 intrinsic version on x86, but also supports other processors, like arm)

Plus I suspect my playing around with this is more "fun" than "useful" - as I said in a comment above it's such a tiny part of the frame time - it's likely the cpu overhead for passing a new Palette to the renderer is an order of magnitude more expensive than the interpolation itself :)

redv:
I think this work is more "useful" than "fun" because allow to use optimal decision in similar cases.

PR #313, opened 2017-12-12
implemented the BribeScreen

redv:
- Implemented the BribeScreen.
- Since the relation is float instead of int, changed the calculation of thresholds.
`	float x = this->getRelationTo(other);`
`	if (x < -50)	{ return Relation::Hostile; }`
`	else if (x < -25)	{ return Relation::Unfriendly; }`
`	else if (x < 25)	{ return Relation::Neutral; }`
`	else if (x < 75)	{ return Relation::Friendly; }`
`	else	{ return Relation::Allied; }`
It is not exactly like vanilla, but very close. This approach allow to simplify calculations and reduce the number of bugs. Another approach is use the int relation like vanilla.

The screen:
bribe_openapoc

### DIFF
--- a/data/forms/city/bribe.form
+++ b/data/forms/city/bribe.form
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openapoc>
+  <form id="FORM_SCORE_SCREEN">
+    <style minwidth="640" minheight="480">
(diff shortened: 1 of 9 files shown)


makus82:
be au ti ful

redv:
Thanks
tumblr_inline_mmr7u5q6481qz4rgp

Issue #260, opened 2017-11-23
No vehicle matching ID "VEHICLE_794" CTD


FilmBoy84:
From the Log following a CTD...

`W 52740355833 bool __cdecl OpenApoc::Vehicle::popFinishedMissions(class OpenApoc::GameState &): No next vehicle mission, going idle
E 52741482678 class std::shared_ptr<class OpenApoc::Vehicle> __cdecl OpenApoc::Vehicle::get(const class OpenApoc::GameState &,const class OpenApoc::UString &): No vehicle matching ID "VEHICLE_794"
  0x00000001402F9D30 PHYSFS_writeSLE16+0x15f400
  0x000000014020208A PHYSFS_writeSLE16+0x6775a
  0x00000001401FFD3A PHYSFS_writeSLE16+0x6540a
  0x00000001400A784C PHYSFS_swapULE64+0x6600c
  0x000000013FFA3D32 PHYSFS_swapULE64+0xfffffffffff624f2
  0x000000013FF7F210 PHYSFS_swapULE64+0xfffffffffff3d9d0
  0x000000014017E015 PHYSFS_swapULE64+0x13c7d5
  0x0000000077A159CD BaseThreadInitThunk+0xd
  0x0000000077C4A561 RtlUserThreadStart+0x21
`

FilmBoy84:
Just a heads up
This CTD has now bugged two games.
Both on the "Easy Map"
Both with vehicle ID 794~
Clicking "Limp Along" takes you straight back to desktop.

JonnyH:
That backtrace looks incorrect, like it can't find the debug symbols (so it's just using the closest symbol it can find, which happens to be physfs).

If you're using the appveyor builds - can you extract the 'debug' package over the top? So the .pdb files are in the same directory as the .exe?

And "try to limp along" simply won't work for some errors (like this), as the error is effectively saying "We're just about to crash, and here is why:", so trying to continue just crashes :)

As a longer term goal, it may be possible to dump the current 'save' state on error, it might give us useful info about what's wrong... But that may fall into exactly the same problem (if the internal data can't be trusted, we may just crash trying to write the save)

FilmBoy84:
@JonnyH No problem, will aim to get on testing with the debug package over the top when i get back from work on Monday...

Currently up to my eyeballs in young Gymnasts :D

makus82:
how young? ) hope its legal XD

FilmBoy84:
[...]

Issue #317, opened 2017-12-14
Another Vehicle ID based CTD


Darkestaxe1:
[openapoc_log.txt](https://github.com/OpenApoc/OpenApoc/files/1557735/openapoc_log.txt)

[save_DKX.zip](https://github.com/OpenApoc/OpenApoc/files/1557738/save_DKX.zip)

Not sure what exactly what does or doesn't cause this one, but I got it by:
-Made new game (medium)
-Investigated first infested building (preemptively) with Valkyrie
-Sold all my starting craft
-Built a bunch of workshops (debug F10)
-Started construction on an annihilator and two retaliaters (debug F10)
Passing time then caused CTD: No vehicle matching ID "VEHICLE_793"

Using build OpenApoc-v0.1-150-gef071d58

redv:
Its looks like the same bug as two previous. This bug happens after selling vehicles.
You can try build which contains fix #316 
https://ci.appveyor.com/project/OpenApoc/openapoc/build/0.0.45

Darkestaxe1:
https://ci.appveyor.com/project/OpenApoc/openapoc/build/0.0.45/job/h2f3xhnmd9i83fni/artifacts

If you look at the filenames you will see that they are OpenApoc-v0.1-150-gef071d58. Maybe I should have called it build 0.0.45? At any rate it's a clean new game start-to-crash in gef071d58/0.0.45.

I don't know if it's related to the old crashes or if it was introduced by your fix, but my money is on the former. I crashed out of several games with these errors before I got frustrated and started making bug reports, and a lot of the crashes I was getting were in cases like this one.

I started another game after it and was able to manufacture up some craft just fine when I didn't sell off my starting craft first. However I started getting errors when saving after sending an Annihilator through a portal and selling off my vehicles. The errors said something about a mission state and I was able to limp through. Those saves CTD on loading. 

Serialization failed: "Invalid enum value for enum OpenApoc::VehicleMission::MissionType:"" vehicles.xml:/vehicles/entry/entry/entry/entry/entry/entry/entry/entry/entry/entry/entry/entry/entry

I suspect you squashed a worker ant and we're looking for the queen.

FilmBoy84:
It appears this is related to issue #260 as @redv has pointed out this relates to the parts of the code dealing with sales/transactions being a complete mess and most likely needing a refactoring before a fix can be made...

Can this report be merged with Issue #260 ?

redv:
[...]

PR #327, opened 2017-12-31
fixed vehicle.die()

redv:
fixed issue #260 "No vehicle matching ID "VEHICLE_794" CTD"


redv:
also fixed #317 "Another Vehicle ID based CTD" because its the same bug.

FilmBoy84:
This appears to have fixed the vehicle bug, although agents still seem to have trouble being removed if sold/transferred, have updated #260 and post the following

"@redv The vehicle crash seems fixed, however, the problem seems to have extended to Agents too, this bug still occurs...

image
image"

Issue #255, opened 2017-11-22
Emergency Vehicles do not pick up crashed road vehicles


FilmBoy84:
Experiencing a problem with a CoS bike not being picked up by a flying emergency vehicle when it crashed. The emergency vehicle flies around the bike aimlessly diving and rising.
Tested on "Easy" Map.

makus82:
Alex say early that we need finish some ground vechicles mechanics  .. somthing like that

FilmBoy84:
Hopefully this issue will be fixed soon.
After a day or two of things building up, so many recovery missions are underway the game slows to a crawl and crashes...

[save_Medium 1.zip](https://github.com/OpenApoc/OpenApoc/files/1528061/save_Medium.1.zip)

FilmBoy84:
This issue seems to cause MASSIVE slow-down in the cityscape, possibly linked to the performance issues referenced in #247 . Can it be fixed promptly please? Many games are doomed to slowing to a crawl or CTD with the sheer number of unrecovered vehicles on the cityscape after a couple of days.

FilmBoy84:
Between 2345 and 2355 this CTD seems to occur...

![image](https://user-images.githubusercontent.com/33913088/34469901-0751e8f8-ef1f-11e7-9d08-1c53e348fec2.png)
![image](https://user-images.githubusercontent.com/33913088/34469903-0de4d89c-ef1f-11e7-9062-ccde1d3ade1d.png)

FilmBoy84:
Vehicles are cleared at last :) Thanks @redv 
However, for some reason smoke from their crash remains in the game, this probably needs fixing too

![image](https://user-images.githubusercontent.com/33913088/34538380-10d86dd8-f0c4-11e7-8b8b-5a011897a408.png)

redv:
Fixed smoke issue. The fix will work for newly crashed vehicles or a new game.

FilmBoy84:
@redv resolved this issue
can no longer replicate
Closing

PR #331, opened 2018-01-03
fixed the mission RecoverVehicle

redv:
fixed issue #255 "Emergency Vehicles do not pick up crashed road vehicles".

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -359,6 +359,10 @@ class FlyingVehicleMover : public VehicleMover
 		}
 		if (vehicle.crashed)
 		{
+			if (vehicle.smokeDoodad)
+			{
+				vehicle.smokeDoodad->update(state, ticks);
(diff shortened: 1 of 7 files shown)


FilmBoy84:
Many thanks @redv for fixing this, i shall get right into testing this evening and report back on issue #260 / #317 as well as check that #255 is totally resolved before letting you know.

FilmBoy84:
@redv 
Have left an update to #255 regarding smoke not disappearing.
Have also introduced a new issue #332 that has started to occur in the most recent 0.1-169 version.

redv:
fixed issue #332 "No vehicle matching pointer". Its because CoS tries to loot UFOs.
fixed issue #333 "City Organisations recover UFOs before X-COM". Because CoS rescue allied Aliens.

redv:
fixed smoke issue as described in #255 
This fix will work only for newly crashed vehicles or a new game.

PR #337, opened 2018-01-03
fixed destructor of GameState

redv:
fixed issue described in #260 
tumblr_inline_mmr7u5q6481qz4rgp

When quit game the destructor of Lab calls to agents from list, but agents already destroyed.
Simplest way is to clear the Lab's agents list in the GameState destructor.

### DIFF
--- a/game/state/gamestate.cpp
+++ b/game/state/gamestate.cpp
@@ -83,8 +83,11 @@ GameState::~GameState()
 		for (auto &f : b.second->facilities)
 		{
 			if (f->lab)
-				f->lab->current_project = "";
-			f->lab = "";
+			{
+				f->lab->assigned_agents.clear();
+				f->lab->current_project.clear();
+				f->lab.clear();
+			}
 		}
 		b.second->building.clear();
 	}


redv:
Another possible approach is to call die() in the ~Agent()

JonnyH:
At this point I think I've realised that StateRef<>s are not "fit for purpose" using shared_ptr<>s internally - as they don't /really/ implement shared ownership - when they're removed from the game the objects effectively become invalid - even if other things are holding StateRef<>s to those (which is generally what these "No object matching ID/No object matching pointer" errors are about.)

I think it needs a bit of a rethink - either just use a unique_ptr<> in the owning StateRefMap, so we just stop pretending the ownership is shared and allows "normal" destructors and backlinks to detach any references (plus the "normal" native memory debugging methods should be able to be used to track leaks/stale references) - or a weak_ptr<>-like object for references outside the central StateRefMap<> - so when they're destroyed they automatically invalidate themselves...

redv:
Yes. StateRef is not a good pattern. The smart pointer which behaves like a regular C-pointer with dereference issues.

PR #345, opened 2018-01-22
9 bugs fixed. 2 additions made.

AndreyCreator:
--------------------------------------------------------------------
Changes and fixes made:
--------------------------------------------------------------------

- Added: Wolfhound APC can be all-terrain vehicle (optional). Option "ATVAPC" is added to OpenApoc_settings.conf.
- Fixed: Initial vehicles (the ones you recieve in the beginning of the game) were not loaded and fueled.
- Fixed: When ground vehicles did not move they still consumed fuel. Now they don't consume fuel while standing.
- Fixed: Ground vehicles crashed when out of fuel. They don't crash now. They request evacuation.
- Fixed: Recovered ground vehicles which stayed out of fuel could be "irrepairable". They are always ok now.
- Fixed: On vehicle equipment screen highlighting flashed too fast. I made flashing like it was in original game.
- Fixed: Now Multiworm Egg and Chrysalis must not be killed to win in battlescape mode.

--------------------------------------------------------------------
Added later:
--------------------------------------------------------------------

- Fixed: Explosion splash area was completely wrong. It is now like in original game (skin36 reverse-engineered correct formula).

--------------------------------------------------------------------
Added later:
--------------------------------------------------------------------

- Added: Idle vehicles turn towards enemies and shoot automatically (turning was not implemented before).
- Fixed: Agent and vehicle location screens. Equipment screen opens on mouse button release. It allows to drag agents by pushing on their icons.
- Fixed: When mouse cursor was hovering above an item, its attack range was displayed incorrectly. It is like in original game now.


AndreyCreator:
One adition is not completed yet. It is ability to fire multiple shits with one shot - like shotgun. It os called "burst" in code. It works, but openApoc does not read this option from XML files.

game/state/shared/aequipment.cpp
```diff
@@ -702,12 +702,18 @@ void AEquipment::fire(GameState &state, Vec3<float> targetPosition, StateRef<Bat
 	auto unit = ownerAgent->unit;
 	auto payload = getPayloadType();
 	Vec3<float> originalTarget = targetPosition;
+	char number_of_shots;
```
JonnyH: I'm not a big fan of 'char' data types (for anything not a string character)

If you don't really care about the size, use 'int'.

If you do, use one of the stdint.h types (int8_t or whatever)

game/ui/components/equipscreen.cpp
```diff
@@ -3,6 +3,7 @@
 #include "game/ui/components/equipscreen.h"
 #include "framework/framework.h"
 #include "framework/renderer.h"
+#include <SDL.h>
```
JonnyH: The game UI should /never/ get into the level of poking sdl directly.

game/ui/components/equipscreen.cpp
```diff
@@ -60,10 +61,18 @@ void EquipmentPaperDoll::setNonHighlightColour(const Colour &colour)
...
 void EquipmentPaperDoll::update()
 {
+	static unsigned long lasttime = 0;
+	unsigned long currenttime, deltatime;
+
 	// Only draw highlight if one of the colours isn't 100% transparent
 	if (slotHighlightColours[0].a != 0 || slotHighlightColours[1].a != 0)
 	{
-		this->slotHighlightCounter += slotHighlightIncrement;
+		currenttime = SDL_GetTicks();
```
JonnyH: Never use SDL outside the framework. Do you really need the global time here? Can't you just scale the local counter time differently if you want to slow it down?

If you think you /really/ need to have this based on real time (and not based on the number of times the screen is updated, which does depend on frame rate) add a function to the Framework instead of calling sdl directly.

game/state/city/vehicle.h
```diff
@@ -204,6 +204,7 @@ class Vehicle : public StateObject,
 	unsigned int shieldRecharge = 0;
 	int stunTicksRemaining = 0;
 	bool crashed = false;
+	bool evacuation = false;
```
JonnyH: Does this need to be serialized too?

game/state/city/vequipment.cpp
```diff
@@ -34,6 +34,8 @@ bool VEquipment::fire(GameState &state, Vec3<float> targetPosition, Vec3<float>
 	};
 
 	auto muzzle = owner->getMuzzleLocation();
+	char number_of_shots;
```
JonnyH: See comment below on "char" type

game/ui/components/equipscreen.cpp
AndreyCreator: Yes, I need real time. Time between frames will be perfect. But I don't know where to get it. Local counter won't work because mousecursor movement changes framerate 4 times and flashing becomes 4 times slower. Can you explain me where can I get milliseconds last between frames?

game/state/shared/aequipment.cpp
AndreyCreator: No problem.

game/state/city/vehicle.h
AndreyCreator: Yes, really. It needs. Thanks for the tip.

AndreyCreator:
I also have a question about reading weapon data from XML files. How exactly do you read it? I found functions extractAgentEquipment() and extractVehicleEquipment(). But I'm a bit confused with your algorythms. Do you read it into one structure and then copy data to another structure?

game/ui/components/equipscreen.cpp
Istrebitel: Well I thinks it's a good idea to leave it a local var for now. Everything uses it (pulsating minimap icons, unit waypoint icons, healing icon etc.). Yes it may be imperfect but it will be consistent across the game. 

You can plan a task in Trello to rework this system everywhere in the game, so that we don't have a myriad of local vars to track pulsation, maybe it's worth doing eventually.

Istrebitel:
We are reading weapon/vehicle/etc data from vanilla game proprietary format. Then we save it to xml and OpenApoc uses xml. 

Methods like "extractAgentEquipment" extract all the data of similar kind (like, all equipment or all vehicles or all units). To do this, a struct is used to define how data is stored by vanilla (see aequipment.h in extractor). Then this data is actually read from file into an array of sorts (see tacp.cpp:82). Then the method goes through the array and creates new entries into gamestate based on what it finds (loop starts at extract_agent_equipment.cpp:392). 

Sometimes there's a need to create more than one entry and that's what you may be noticing. 

For example, vanilla only has one chrysalis entry. But in OpenApoc the way I implemented them is that they can't move their bodies. If they can't move their bodies they can't be turned two different ways. So I've made two different body types for chrysalis so that there's a x-facing chrysalis and y-facing chrysalis.  Same with egg - can rotate it's "head" but body can be facing in two different directions. Same for android - can have different kinds of 'heads". 

Or for another example, game has one "enzyme" damage type and entropy launcher shares it with other weapons. But it made sense to make the entropy enzyme a damage type property (so you can assign it to a weapon and it will do the effect). So we have to dupe it - one enzyme for all weapons, one for entropy launcher.

As to why things are hardcoded it's because it's faster that way. Most of the things are not hardcoded inside code but taken up into the beginning of relevant files. It will be easy to externalize them when we know what to externalize and how to organize it. Otherwise it's a lot of extra work, and some settings become irrelevant or get grouped together into one or branched into multiple ones. Easier to change when it's hardcoded.

The plan was to make them non-hardcoded when game is fully playable, make an xml file with engine paramters and externalize everything there.

Istrebitel:
On topic of hardcoded stuff and uncomplete solutions, the real thing that needs attention, if you want to get into it and fix that kind of thing, is that we don't distinguish between game rules and game state.

What I mean is for example we have research class, and both "how many hours it takes to research" and "is it complete by player" are its members. But first is a game rule (like damage of a weapon or height of a certain alien) while second is a game state (like current hit points of a certain unit or player money balance). 

This both makes saving take way more time and space on disk and precludes changing mods in an already ongoing game, and both are very big problems. 

Instead, we should have gamerules separate from gamestate. 

Gamerules should contain everything that never changes in the game, and defines the game (list of orgs, unit types, vehicle types etc.). Gamerules do not get saved, they only get loaded - first from base game then from mods in mod load order. Like Elder Scrolls series.

Gamestate should contain everything that defines current state of the game. Hitpoints, positions of objects, player balance and score, etc. Gamestate should get saved and loaded, and after it's loaded it should be checked for consistency with rules (like, does an object refer to an unexisting class?)

game/ui/components/agentassignment.cpp
```diff
@@ -84,7 +84,8 @@ void AgentAssignment::init(sp<Form> form, Vec2<int> location, Vec2<int> size)
 			this->currentAgent = agent;
 		}
 		auto icon = c->findControl(ControlGenerator::AGENT_ICON_NAME);
-		if (agent && icon && c->isPointInsideControlBounds(e, icon))
+		if (agent && icon && c->isPointInsideControlBounds(e, icon) &&
+		    Event::isPressed(e->forms().MouseInfo.Button, Event::MouseButton::Right))
```
redv: It is not vanilla behaviour. In vanilla X-Com work both mouse buttons for load the equip screen.

game/ui/components/agentassignment.cpp
```diff
@@ -103,7 +104,8 @@ void AgentAssignment::init(sp<Form> form, Vec2<int> location, Vec2<int> size)
 
 		auto vehicle = c->getData<Vehicle>();
 		auto icon = c->findControl(ControlGenerator::VEHICLE_ICON_NAME);
-		if (vehicle && icon && c->isPointInsideControlBounds(e, icon))
+		if (vehicle && icon && c->isPointInsideControlBounds(e, icon) &&
+		    Event::isPressed(e->forms().MouseInfo.Button, Event::MouseButton::Right))
```
redv: The same. This is not a behaviour of vanilla.

FilmBoy84:
@AndreyCreator With regards the "burst" shot and scatter. Is this by any chance linked to the "Missile Defence" weapons such as:

Laser Defence Array
Plasma Defence Array
Plasma Multi-System

I recall these having multiple projectiles from a single shot and it says as much in their UFOPedia entries...

image
image
image

game/ui/components/agentassignment.cpp
AndreyCreator: In vanilla I can drag-&-drop units when clicking on their portraits. Here in OpenApoc I can't do that. When I start dragging - equipment screen pops up. This is why I changed buttons.

AndreyCreator:
@FilmBoy84

Now it just repeats firing process N times. Nothing more. If weapon has some initial scattering than you will see how projectiles fly in arc. If it is no scattering then projectiles will fly together and you will see all of them like one shot (but eyes fool you).

But you are right here. Plasma multi system and missile defenses must have different logic: they must fire at multiple targets at the same time. And it must work different way: seek for targets and attack each of them. I is good idea. I will implement it.

game/ui/components/agentassignment.cpp
redv: In this case need to change the drag&drop behaviour. I.e. check the MouseUp event instead the MouseClick on the icon.

AndreyCreator: Ok.

redv: I just tested vanilla X-Com behaviour. The equip screen is loaded after the MouseUp event.
I propose the algorithm:

- the MouseDown event for an agentControl or vehicleControl
```
if (mouseOverIcon) {
    remember the iconControl;
}
else {
    remember nullptr;
}
do a conventional drag&drop action for MouseDown; // if needed
```
- the MouseUp event for an agentControl or vehicleControl
```
if (mouseOverIcon && itsTheSameIcon) {
    cancel drag&drop;
    start an equip screen;
}
else {
    do a conventional drag&drop action for MouseUp; // if should be of course
}
```

AndreyCreator: This algorythm won't work because player can start dragging and release button while staying on the same icon. But don't worry. I've made things right already.

forms/multilistbox.cpp
```diff
@@ -215,7 +215,8 @@ void MultilistBox::eventOccured(Event *e)
 				{
 					selectedItem = child;
 					selectionAction = selectedSet.find(child) == selectedSet.end();
-					if (selectionAction && funcHandleSelection(e, child, true))
+					bool sel = funcHandleSelection(e, child, true);
+					if (selectionAction && sel)
 					{
```
redv: This code break the selection behavior.
Because of drag&drop operations the MouseDown and MouseUp events should handled separately. But the strategy function funcHandleSelection() should called only once per click (MouseDown + MouseUp). In your code the funcHandleSelection() called twice per click, that leads to the broken selection behaviour.

game/ui/components/agentassignment.cpp
```diff
@@ -694,6 +700,7 @@ void AgentAssignment::eventOccured(Event *e)
 			// if dragged - change location of dragged list
 			if (isDragged)
 			{
+				isMouseMoved = true;
```
redv: Modern mice have too sensitive sensors. Often when user do a click the cursor may shifts to several pixels. It's especially issue when need to check a difference between drag&drop and a conventional click especially if the MouseDown and MouseUp events handled separately. Thus added insensibility area (5x5 pixels in this case).
You set isMouseMoved=true just after MouseMove event. In this case if user want to load the equip screen his mouse must not be shifted. Or the user should click again.

May be better to relax a demands to users precision and allow to click in the insensibility area? Moreover, almost all code done below.

forms/multilistbox.cpp
AndreyCreator: No, it does not break. Tested already.

game/ui/components/agentassignment.cpp
AndreyCreator: It works without problems with my (not cheap) high precision Razor mouse.

redv: Just now tested. If mouse moved even one pixel the equip screen did not start.
The vanilla X-Com works well even if mouse shifted a bit during the click.

forms/multilistbox.cpp
redv: The funcHandleSelection() is the strategy function designed by coder. It should called only once per click because it may contain state. In this code the function run twice per click thus the state may switched back and forth. It may matter.

redv:
Split, please, the PR to several branches (each fix to a separate PR if it possible). Most fixes are OK and useful but some need to improve.

AndreyCreator:
Conflicts:
- vehiclesheet.cpp has "tr(...)" added only. It is easy to resolve;
- transactionscreen.cpp can be discarded. All that was made is open ufopaedia on right click and added "tr(...)" in one string. I can remake it with a new file.
But I can not resolve it now because current version of Openapoc is incompatible with other my changes, Can you just edit one file and remove 2 others?

JonnyH:
I've always used the git command line directly to rebase and handle conflicts. I don't really know how various UIs, like github desktop, deal with this, but I can give a summary of the commands I use:

For example, if I have the upstream repository set to the remote "upstream", "git rebase upstream/master" will try to rebase the currently checked out branch on top of the latest master.

This will try to apply each new commit in the currently checked out branch. If a patch fails (e.g. there's a conflict, where both the upstream/master branch and my local branch have ended up modifying the same chunk of code) it'll stop the rebase on that commit. It'll tell you which files need modifying with "git status" - open those files in an editor and look for the merge conflict lines (lots of ">>>>>" and "<<<<<" make it pretty easy to search)

You then edit the chunk that conflicted to be correct, in that is takes into account all master changes, and and changes you have made on your local branch /up to that commit/, and save the file. Note, if you changed that chunk again in a later commit, it's normally a good idea to not bring that change back, instead integrate it when the rebase gets to that later commit.

Once all the files with conflicts have been edited and fixed, you run "git add" the files, then continue the rebase with "git rebase --continue"

Once the rebase has succeeded, you can push that branch over the top of your fork's branch on github, and it'll automatically update the PR, and should allow it to be merged with no problems.

Note, if you make a mistake, or need to reset for some reason during the rebase - "git rebase --abort" brings the state back to as if you never run the command (if you have an outstanding rebase - if it's already completed you'll have to look in the reflog if you realise something's gone wrong).

During the rebase, don't use other git commands that modify the state of the repository (checkout other branches for example), and it's normally not a good idea to pull in "other" changes from commits "later" than the one it's currently looking at in this rebase, or "fix" things is you notice mistakes that are not /directly/ related to that specific commit already, as that will likely make later commits conflict in the rebase, making it more work overall. Instead, wait until the commit with that specific change is rebased, or add another commit after the whole rebase with any unrelated mistakes.

AndreyCreator:
I made exactly as you wrote but a problem appeared. When I tried to solve a problem for the first time I created 2 requests: "Merge branch 'master' into master" and "Revert". So, console git applies all changes including "revert" and returns me back to old OpenApoc version. Is there a way to remove these 2 requests from Github pull request?

PR #347, opened 2018-01-23
WIP: implementation of the Observer pattern [incomplete]

redv:
The Observer pattern is a possible replacement for the StateRefMap + StateRef.

I.e. ObservableLinkedMap contains agents:
```ObservableLinkedMap<Agent> agents;```

The Lab is Observer for the agents map.
```class Lab : ObserverLinkedMap<Agent>```

If agent died (during battle or even if the agents map destroyed during the game's quitting) the lab will receive the event "removed" and remove the agent from its own list.

The vehicles horizontal list could be observer for the vehicles map.
horizontallist_vehicles
If the player sold or buy a vehicle then the list updates automatically after receive an event. If a vehicle change its state (i.e. enter to building) then the list receives the "changed" event and changes picture of the vehicle accordingly.

The same for agents horizontal list (in battle mode too) and for the organisations list.

Main idea is forget about dereference hell of the StateRef and use the event-driven programming.
So, at this stage it is just an idea and the code is not ready for merging.

The observer.h contains the Observer and the Observable base classes.
The ObserverLinkedMap is a template of abstract class derived from the Observer.
The ObservableLinkedMap is a template class derived from the Observable.
The map is linked because the class contains hashmap and linked list to keep order of inserting.

### DIFF
--- a/game/state/CMakeLists.txt
+++ b/game/state/CMakeLists.txt
@@ -78,6 +78,7 @@ set (GAMESTATE_SOURCE_FILES
 	gamestate_serialize_generated.cpp
 	gametime.cpp
 	message.cpp
+	observablelinkedmap.cpp
 	savemanager.cpp
 	
 	tilemap/collision.cpp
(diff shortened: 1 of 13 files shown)


makus82:
Hey redv i send you invites to get to slack 
but now we in discord invite so here  if you want its easy 
https://discord.gg/0nS950g1olwDxzXB

PR #352, opened 2018-02-20
Clean README formatting

rogersachan:
This pull request does a number of cosmetic fixes to the README, including the following:
* Properly spaced headers
* No trailing whitespace
* Consistently formatted lists
* Removed underscores in headers
* Fix the grammar in the tweet button
* Remove stray html breaks.
* Further grammar fixes
* Upgraded a bunch of URLs to HTTPS and fixed a broken link

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,8 +114,8 @@ OpenApocalypse is built leveraging a number of libraries - to provide needed fun
 The following libraries are also used, but are shipped as submodules in the repository and directly included in the build, so you don't need to install these dependencies to build or use openapoc.
 
 * [GLM](https://glm.g-truc.net) - Math library.
-* [libsmacker](http://libsmacker.sourceforge.net/) - Decoder for .smk video files.
-* [lodepng](http://lodev.org/lodepng/) - Reading/writing PNG image files.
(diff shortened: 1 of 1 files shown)


rogersachan:
Could I suggest trimming the number of buttons on the README? It's really overkill to have 20 lines (code) of just buttons.

This PR doesn't do that, so if they need to stay this should not be an issue.

JonnyH:
I kind agree about the buttons - I kinda see github as developer focused - having a link to the latest build (build /status/ shouldn't be needed as such - as the builds should be verified before the PR is merged...), and a link to the website (from which you can get to the forums, IRC, other social media links etc.) should be sufficient... right?

Comments anyone else?

makus82:
we need all beacose you dont know when and how )

makus82:
..we can remove issues and trello buttons

makus82:
readme read usually who newcomers so why not to output all possible info?)

Thunderforge:
I was just thinking of creating a pull request for cleaning up the Readme. Is this something that can be merged? Or are there still outstanding issues (possibly ones that other people could do for this PR)?

Issue #353, opened 2018-03-07
No aequipement type matching ID "VEQUIPMENTTYPE_BOLTER_4000_LASER_GUN"


FilmBoy84:
After selling off my lasers and going to equip agent after an Alien alert, the game CTDs with the following messages and log

Please also note a number of typo-errors!
"aequipement" should be "equipment"

![image](https://user-images.githubusercontent.com/33913088/37101823-5d0fdd18-221e-11e8-91f3-6b3a724c444a.png)
![image](https://user-images.githubusercontent.com/33913088/37101834-618bb2b8-221e-11e8-8217-ccdec82db21b.png)
![image](https://user-images.githubusercontent.com/33913088/37101838-679ba74e-221e-11e8-9c92-4ecf3f6b359c.png)

Clicking Limp-Along responds with the following error
![image](https://user-images.githubusercontent.com/33913088/37102084-edd236c0-221e-11e8-84cd-5515ac1f20b6.png)
![image](https://user-images.githubusercontent.com/33913088/37102090-f27f0c20-221e-11e8-96c2-4ae935a303e9.png)

Clicking Limp-Along again responds with a further error
![image](https://user-images.githubusercontent.com/33913088/37102107-05c297f2-221f-11e8-9515-05784c19c11e.png)
![image](https://user-images.githubusercontent.com/33913088/37102114-0ae9dc4a-221f-11e8-8902-9f7007df2b63.png)

AndreyCreator:
Those "typo-errors" are not errors. Aequipment should stay as is because there is class "AEquipmentType" in the code and this message is related to it.

And it seems that the game tried to arm an agent with Bolter 4000 laser gun, but did not find proper data for this weapon.

FilmBoy84:
@AndreyCreator Many thanks for looking into this.

I believe there is still a typo there "aequipement" should be "aequipment".
I also believe that the term "aequipment" refers to "Agent Equipment"
The Bolter 4000 Laser gun is NOT agent equipment, so should not appear in the agent equip code.
https://www.ufopaedia.org/index.php/Bolter_4000_Laser_Gun_(Apocalypse)

Additionally, there is no Bolter 4000 at the base, the two provided on game start were sold off immediately to generate money; so it is unclear as to why the game is calling for this item when accessing the agent equipment screen for the following reasons:

1) Neither weapon exists in X-COM inventory
2) The weapon is not Agent Equipment
[...]

PR #374, opened 2018-05-19
Fix bug when selling vehicle equipment or ammo

Jarskih:
Fixed typo in buy and sell screen.

Related to:
No aequipement type matching ID "VEQUIPMENTTYPE_BOLTER_4000_LASER_GUN"
#353

Selling vehicle equipment/ammo resulted in selling non existing agent
equipment and ammo.

### DIFF
--- a/game/state/city/vequipment.cpp
+++ b/game/state/city/vequipment.cpp
@@ -192,7 +192,6 @@ bool VEquipment::reload(GameState &state, StateRef<Base> base)
 		}
 		else
 		{
-			LogWarning("Vehicle refueled/rearmed");
 			return ammoSpent > 0;
 		}
 	}
--- a/game/state/city/vequipment.cpp
+++ b/game/state/city/vequipment.cpp
@@ -5,8 +5,8 @@
 #include "game/state/city/base.h"
 #include "game/state/city/city.h"
 #include "game/state/city/vehicle.h"
-#include "game/state/gamestate.h"
 #include "game/state/gameevent.h"
+#include "game/state/gamestate.h"
 #include "game/state/rules/city/citycommonimagelist.h"
 #include "game/state/rules/city/vammotype.h"
 #include "game/state/rules/city/vequipmenttype.h"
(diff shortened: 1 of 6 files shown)


FilmBoy84:
Great stuff, many thanks @Jarskih <3

game/state/gameevent.cpp
```diff
@@ -124,9 +124,9 @@ UString GameVehicleEvent::message()
 				return tr("An illegal flyer has been detected.");
 			}
 		case GameEventType::NotEnoughAmmo:
-			return tr("Not enough ammo to rearm vehicle");
+			return format("%s %s", tr("Not enough ammo to rearm vehicle:"), vehicle->name);
```
JonnyH: This should probably have the "%s" placeholder thingie in the tr() string - just in case other languages don't make sense having it at the end

game/state/city/vequipment.cpp
```diff
@@ -138,8 +139,27 @@ void VEquipment::update(int ticks)
 	}
 }
 
+void VEquipment::noAmmoToReload(GameState &state, VEquipment *equipment) const {
```
JonnyH: I believe both "gamestate" and "equipment" arguments can be const here?

(I know it's not consistently applied through the codebase right now, but I prefer "const if possible")

JonnyH:
Actually I didn't notice the lint pass failed - we use clang-format (specifically version 4.0, the newer 5.0 is *probably* ok but it does some things slightly differently) to format the code according the standards we use.

Hopefully the ide/platform you use has some way of integrating that (many IDEs, like visual studio or qtcreator, have plugins that can automate it and use the config in the OpenApoc git, or you can install it and run it manually with "clang-format -i path/to/file.cpp". Or in the worse case, the output of the travis "lint" CI pass is a diff file that you could apply, or just manually look over for the fixes)

game/state/gameevent.cpp
Jarskih: I am not sure what you mean by this. Maybe you can change this one to fit other languages?

Jarskih:
Thanks Johnny for review. I think I fixed all the issues now. I am not sure how to fix the translation issue you mentioned.

PR #375, opened 2018-05-24
Messages for vehicle rearmed, refuelled and not enough ammo or fuel events

Jarskih:
- New notification if there is not enough ammo or fuel to refuel or
rearm the vehicle.
- Notifications when vehicle is succesfully rearmed and refuelled.


Jarskih:
This one can be closed since this is included in the earlier pull request.

PR #377, opened 2018-06-01
Null ptr checks

Jarskih:
For some reason vehicle tileobjects are null pointers. Added some checks to prevent game from crashing.


JonnyH:
A good part of me feels that handling null objects in TileObjects is a workaround for poor lifecycle management (not surprising as it was all ad-hoc with no real design) - but that's not going to change soon so fixed are always good :)

Issue #389, opened 2018-09-16
Crash after try to buy equipment for vehicles & ships


Maddead:
OpenApoc-v0.1-245-gb4416b1e

For the first time in 1st day buing equipment for vehicles and sell vehicles going normally, but fo the second time the game is crashes when I clicked icon "equipment for aircrafts" in buy menu.

"No aequipment type matching ID "VEHICLE_794" and I try to click limp along "No aequipment type matching ID "VEHICLE_794 found"
![bug 1](https://user-images.githubusercontent.com/43318210/45597834-e41f8180-b9fb-11e8-80c2-99708765fd2c.JPG)
![bug 2](https://user-images.githubusercontent.com/43318210/45597835-e550ae80-b9fb-11e8-9553-6d1fdf8b449d.JPG)

Problem signature:
  Problem Event Name:	APPCRASH
  Application Name:	OpenApoc.exe
  Application Version:	0.0.0.0
  Application Timestamp:	5b919c27
  Fault Module Name:	OpenApoc.exe
  Fault Module Version:	0.0.0.0
  Fault Module Timestamp:	5b919c27
  Exception Code:	c0000005
  Exception Offset:	002a522a
  OS Version:	6.1.7601.2.1.0.256.1
  Locale ID:	1049
  Additional Information 1:	0a9e
  Additional Information 2:	0a9e372d3b4ad19135b953a78882e789
  Additional Information 3:	0a9e
  Additional Information 4:	0a9e372d3b4ad19135b953a78882e789

FilmBoy84:
Thanks for the report, this is further confirmation that the bug reported in issue #260 is still present

I believe @redv and @JonnyH were looking into it, but the code is a mess and likely needs a complete refactoring

Workarounds are listed on issue #260 accordingly, but in effect it means not selling anything or hiring/firing staff

It's a monumentally frustrating bug that has largely stopped me testing the game until it's fixed

@makus82 do you have any skills that may assist in fixing it?

Jarskih:
@FilmBoy84 and @JonnyH this should be fixed.

FilmBoy84:
Testing on OpenApoc-x64-v0.1-251-g8cd7a0db confirms this as fixed
Great work Jarskih <3 and JonnyH for merging so quickly
Closing it now

PR #392, opened 2018-10-10
Fix to #389 & #260

Jarskih:
When selling vehicles the switch was not exited and caused bugs related
to agent equipment.


FilmBoy84:
Fantastic news! Many thanks @Jarskih for fixing this
I may be able to return to play-testing and bug-reporting again now
Also thanks JonnyH for merging this so promptly <3

PR #399, opened 2018-10-13
New feature: All units will be set to run and kneel at the beginning of combat (disabled by default)

Jarskih:
With this feature enabled you don't need to click through every unit and
set them to run and kneel.


FilmBoy84:
@Jarskih I've credited your work on the latest OpenApoc video at https://youtu.be/9PgQt_1qoIU
Looking forwards to seeing what features and fixes you submit next <3

PR #405, opened 2018-10-15
Idle vehicles turn towards enemies and shoot automatically

FilmBoy84:
Idle vehicles turn towards enemies and shoot automatically (turning was not implemented before).

Co-Authored-By: andreycreator <andreycreator@users.noreply.github.com>

### DIFF
--- a/forms/multilistbox.cpp
+++ b/forms/multilistbox.cpp
@@ -563,4 +563,4 @@ void MultilistBox::setFuncSelectionItemRender(std::function<void(sp<Control>)> f
 	funcSelectionItemRender = func;
 	setDirty();
 }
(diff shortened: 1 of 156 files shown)


Jarskih:
Did you use clang format on this? On windows you can run "clang_format_win.bat" to fix formatting issues.

JonnyH:
There's a lot of noise from unrelated format changes (clang-format 4.0 doesn't add the missing closing namespace comments, but doesn't reject them either - while later versions add them, so you running through a later version then 4.0 caused them all to appear)

But that's fine, I like them there anyway, and while it might cause a slightly dirty history... meh

Issue #404, opened 2018-10-15
Crash with collapsing building after raid DISABLED


FilmBoy84:
For some reason, even though collapsing buildings is disabled there appears to be the following example of a crash

Karl_4 save - this is in a mission, kill the last CoS, mission ends, as soon as game unpaused there is a CtD

CtD save - this is made just after the mission, if loaded, the building collapses - even though this feature is not enabled and the game does not CtD but continues as expected

[saves.zip](https://github.com/OpenApoc/OpenApoc/files/2478404/saves.zip)

Reported on DISCORD by @Dark-Lark

FilmBoy84:
I have not been able to replicate this issue on recent builds
Therefore closing
If the issue arises again, we can always make a new one with up to date info and reference this

PR #406, opened 2018-10-16
Fix to crash when finished raiding a building #404

Jarskih:
Fix to issue #404 
The current building id was not found for the after raid notification.
This fixes the issue but there might be another problem with raiding and
setting current mission.


FilmBoy84:
I have not been able to replicate issue 404 with recent builds and as this PR is out of date and needs changes i may as well close.

PR #407, opened 2018-10-16
No base error fixed #400

### LINKED ISSUE #400: "Failed to find base in XXX" Error
When searching a building for alien activity, the following crash occurs (if aliens not found? - Unsure, craft dropped them in this building moments earlier)

image
image
image

Jarskih:
Added check that base exists in the building owned by X-Com.


FilmBoy84:
Sadly, this still occurs in latest build, but as discussed on discord... That's probably because the building is owned by X-COM in the files... Even though it should be Evonet

image
image
image

JonnyH:
Looks good to me. I'm not sure what it means for the player to own a building without a base - but there's precedent (I think one of the cities starts with a recycling plant owned by x-com - probably a typo :)

PR #408, opened 2018-10-16
Pause game when x-com craft arrives to investigate for aliens

### LINKED ISSUE #354: No "Commence Investigation?" prompt on agent arrival at a "Live Alien Spotted" location
Although ordering agents to alien incidents in the city is now implemented when aliens are discovered, on the agent arrival at the location the game does not automatically prompt the user to begin the mission as happens in vanilla. This should be implemented as per the vanilla prompt below.

Please note ALL assigned agents and craft have to arrive before this prompt is given (meaning the game checks this)

image

Jarskih:
Added message box when craft is sent from alert screen as in original version. Clicking yes will start the mission and cancel will pause the game. 
Currently only works with one craft and it will not wait for each vehicle to arrive. Also doesn't work at all when sending agents by foot.
Fixes #354


game/state/city/building.h
```diff
@@ -66,6 +66,7 @@ class Building : public StateObject, public std::enable_shared_from_this<Buildin
 	unsigned ticksDetectionTimeOut = 0;
 	unsigned ticksDetectionAttemptAccumulated = 0;
 	bool detected = false;
+	bool investigate = false;
```
JonnyH: This probably needs to be added to the serialization code - just adding the member to gamestate_serialize.xml for the building object should be sufficient.

Otherwise saving/loading when an investigation is on it's way will lose the state.

game/state/city/vehiclemission.cpp
```diff
@@ -1402,6 +1402,23 @@ void VehicleMission::update(GameState &state, Vehicle &v, unsigned int ticks, bo
 		}
 		case MissionType::GotoLocation:
 		case MissionType::Land:
+			if (!finished || v.currentBuilding == nullptr)
```
JonnyH: What happens if you select multiple vehicles/agents on foot to investigate a sighting? IIRC the OG xcom waited for everyone to arrive

JonnyH: Just noticed you mentioned this exact case in the comment above :)

I'd prefer if we fix that up before merging though, unless it ends up so hilariously complicated I'd try not to leave known holes.

FranciscoDA: Wouldn't a semaphore-like approach work?
You can set `investigate` to `n` when sending `n` vehicles, then decrease the counter as they arrive to the building. When decreasing from 1 to 0 you can push the commence investigation event :)

PR #409, opened 2018-10-16
Fix to Agent tried to move walking without a standing/flying body state #401

### LINKED ISSUE #401: Agent tried to move walking without a standing/flying body state
Agent tried to move walking without a standing/flying body state
Whilst burning floor collapses and agents are given orders to move

image
image
image

Jarskih:
Added check which only makes current player's units to run and kneel.
Fixes #401


FilmBoy84:
This appears to be fixed in build OpenApoc-x64-v0.1-255-ge6655949
Excellent work Jari! <3

FilmBoy84:
As the patch that required this fix is already in Master; pushing to master now to stop the multiworm bug Issue #401 ASAP
Code checks out and formatting to standard required by OpenApoc definitions.
Great work again Jari

PR #411, opened 2018-10-18
fix compilation for boost >= 1.66

FranciscoDA:
boost/uuid/sha1.hpp has been deprecated and moved to boost/uuid/detail/sha1.hpp


FilmBoy84:
Looking great FranciscoDA; had a quick look over myself so will let this get reviewed by one of the more experienced devs and hopefully merged quickly

Appveyor Build is already up for x64 too :+1:

FilmBoy84:
@JonnyH All appears good here
Are you OK to merge? Or is there anything else you need done?

JonnyH:
It's fine to merge, I just was thinking we should move away from boost's sha1 implementation going forward, as they're clearly not really meant to be stable exposed APIs here

Issue #282, opened 2017-11-30
Randomise Starting Base


FilmBoy84:
Under Vanilla, the starting base is chosen at random for each map according to the location of "purchasable" bases.

6 Possible Locations (Novice, Easy)
7 Possible Locations (Medium, Hard)
8 Possible Locations (Super-Human)

This doesn't seem to happen in OpenApoc.
On Easy Map for example i get Warehouse 2 every time i start under OpenApoc. On Vanilla it changes.
Can the random start base please be included?

Details of Locations are here:
http://www.ufopaedia.org/index.php/Base_Locations_(Apocalypse)

FilmBoy84:
Thanks to FranciscoDA's efforts to add an RNG seed, this now randomises base location start
There's some issues with a bias towards certain buildings, but it does at least choose different ones occasionally
Closing issue as to all intents and purposes
It's fixed :)

PR #412, opened 2018-10-18
add option to seed rng on game start

FranciscoDA:


FilmBoy84:
Whilst not directly relevant to this issue, We probably also need to consider adding a random starting base as per Issue #282 and according to the documentation of vanilla behaviour at https://www.ufopaedia.org/index.php/Base_Locations_(Apocalypse)

Whilst vanilla had a preferential bias towards certain bases, there was intermittent instances where a player would find themselves starting at a COMPLETELY different location in the city.

For example, SuperHuman map had two different warehouses as the most likely of eight possible bases but if you restarted enough times, you would eventually get a game at any of the available options

Issue #421, opened 2018-10-22
[Linux] No any text in load/setting/buy/sell menu


DisDis:
after https://github.com/OpenApoc/OpenApoc/issues/414
No any text in load/setting/buy/sell menu
![image](https://user-images.githubusercontent.com/87201/47315357-06d13580-d64d-11e8-874c-10b903c5b549.png)
![image](https://user-images.githubusercontent.com/87201/47315500-6deeea00-d64d-11e8-8829-52af694129e8.png)

JonnyH:
This is likely due to the fonts not (Yet) containing the required characters for the translation - this problem will exist for all languages that have new characters that weren't supported by the original game (and in fact that was missing some variations, like capitals, of some characters too).

This isn't really a "bug", so much as these new translations are an incomplete feature of openapoc for now.

You should be able to try different languages by setting it in the config file - see here: https://www.ufopaedia.org/index.php/Installing_(OpenApoc)

The only one I /know/ that is complete is english, as it's the language I happen to use.

DisDis:
It is not work.
```
[Framework]
Language=en_GB.UTF-8
```
![image](https://user-images.githubusercontent.com/87201/47316377-a8f21d00-d64f-11e8-930a-8ffb317b14ed.png)
@JonnyH ^^^

JonnyH:
Huh, interesting. I'll have to look into it further.

JonnyH:
This bisected back to 73f3a28ff6103dd51d4a74ccc3a80f98ed030bb0

Assuming that's correct, that's a GLM update - maybe a transform we were using now behaves differently?

DisDis:
@JonnyH what will i can try?

JonnyH:
^^ You can try that build when it finishes

DisDis:
"master" works

PR #422, opened 2018-10-22
Enable GLM_FORCE_CTOR_INIT to init all glm times (e.g. vec2<>/vec3<>)…

JonnyH:
… to zeros

The recent glm update seems to have stopped doing this by default, but there's
at least one place (and likely many more) that expect this

This seems to cause issue #421

### DIFF
--- a/library/vec.h
+++ b/library/vec.h
@@ -2,6 +2,8 @@
 
 /* GLM_GTX_transform is an experimental extension now, apparently */
 #define GLM_ENABLE_EXPERIMENTAL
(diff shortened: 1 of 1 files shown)


JonnyH:
Not sure if this is a good fix going forward - or if we should fix any implicitly initialized vec<>s in our code.

SupSuper:
Well I fixed the uninitialized in 31213b74577fd0e610ad4086581804a921784dc6 just in case.
Seems the Travis check broke anyways, so might as well close this.

PR #424, opened 2018-10-24
add agent name translation

Skin36:


JonnyH:
Do we want to translate names like this? I don't think using the translation system here is a good fit, as there /must/ always be a 1:1 mapping between names of different langues.

Plus it makes sense for the agents to be named 'in-universe' - and not change based on what language I happen to be playing the game in.

Or would this just be used for alphabet conversion or similar for the 'same' name?

Issue #361, opened 2018-03-20
Measurements should be in Metres and NOT Tiles (Cityscape and Battlescape)


FilmBoy84:
(From @Cyriel 's report on Discord)

Measurements in Vanilla Apoc use metres for their calculations which allows for partial coverage of tiles under certain circumstances. Currently OpenApoc uses full tiles for measurement, and these are often very different to their vanilla counterparts.

As an example...
VANILLA:
![example2](https://user-images.githubusercontent.com/33913088/37648520-1f6c8f98-2c27-11e8-8b37-5cc6f58230fd.PNG)

OPENAPOC:
![example1](https://user-images.githubusercontent.com/33913088/37648549-2e6abefc-2c27-11e8-8095-f1a17edef57f.png)

The effect of this problem is that OpenApoc weapons can reach far further than their Vanilla equivalents in most cases. The battlescape (agent) weapons have similar problems too.

To fix all of this we need to find out what the "metres per tile" value is for the Cityscape and, if it is different, what the "metres per tile" value is for the Battlescape and allow for weapons to use the Metres values from Vanilla.

FilmBoy84:
@skin36 reports Cityscape Tile values as:

X & Y directions - 32metres
Z direction - 16M

In short, tiles are NOT cubes

FilmBoy84:
Regarding how tiles were split up in the X-COM games (From Terry Greer: http://www.terrygreer.com/xcomapocalyse.html ):

"_Tiles were complex data structures. Each one also had a reference to line of sight definition – this was a simple 4x4x4 grid [UFO/TFTD] – a sort 3d texture – where each cell was either solid or empty. There were a relatively small number of these solid line of sight definitions, but enough to approximate the 3D shape of any shape of tile created. This meant that the world was effectively broken into a 3d voxel grid where weapon fire could be accurately ray-traced. It was a ball ache to set up initially, as every tile had to have a line of sight definition assigned, but really made life easy later on as collision in game then became automatically generated from the map editor.

It was a genius approach, especially in the years before decent 3d raytracing – and one of the key identifying features of an xcom game. Players always loved being able to snipe enemy forces from right across the map through a couple of windows or blown open walls. It’s also a technique which I think useful to reinvent for use in conjunction with true 3D worlds and can think of lots of good mechanics you could use them for._"

[...]

Issue #243, opened 2017-11-13
Time speed accelerates wich time speeds 1-4 but vehicles not (vehicle lot slower then in vanila)


makus82:
Time speed accelerates wich time speeds 1-4 but vehicles not (vehicle lot slower then in vanila)

makus82:
i think all time indicators need to be doubled...

makus82:
https://youtu.be/GB54cYHpyRE

redv:
It's an engine issue (I mean the game's engine :)
The engine's rendering isn't optimized and much slower than vanilla what affect to vehicles. I tested optimized version #247 (strategy view mode only). In that version, vehicles behave like vanilla.

makus82:
Strange beacose i use last build before appveyor down
and there vehicles slower then in vanilla
in openapoc speed 4 its like speed between 2-3 of vanilla 

redv могу инвайт кинуть в слак нужен твой email

redv:
No, I mean the my test build of OpenApoc #247 . In this build the strategy view mode works much faster and speed of vehicles is close to vanilla. Thus I believe that this issue because the game's engine itself but not because vehicle's code improperly designed.

Мой емайл должен быть виден в аккаунте: https://github.com/redv

JonnyH:
The update() look *should* be doing a fixed time increment every rendered frame - so if the device it's running on can't keep up, it should slow down consistently. The same amount of time should pass per rendered frame, not wall clock time, so everything should slow down at the same rate - vehicle movement, in-game clock time, any other time related stuff.

The most interesting thing to measure would be the difference in openapoc between the different speeds - if the vehicles (and clock time) don't move 2x faster between speed 2 and 3, for example, that's a bug. If they are proportionally correct, then it's likely a scale issue with the values stored as the vehicle speed (We already scale the values slightly differently, as the original ran at much lower than 60fps :)

makus82:
ok so i will close this when it will be in last build)

JonnyH:
No, don't close it - unless you've proven the scale between the different speeds (1/2x, 1x, 2x, 4x etc) is correct relative to each other in-game - and then we likely need to tune the global scale so it's all approximately the same speed as the original (in real-time)

makus82:
I said so =P
if in last build will be correct speeds
but about global speed (speed 5) i dont know how to observe that they correct)

JonnyH:
[...]

PR #426, opened 2018-10-26
scale weapon range correctly in vehicle equipment sheet

FranciscoDA:
As discussed in the discord: the range values in apoc's data file seem to be in half-metres. This fix should scale them accordingly while still displaying the value in tiles, as intended (related issue #361)

### DIFF
--- a/game/state/rules/city/vequipmenttype.cpp
+++ b/game/state/rules/city/vequipmenttype.cpp
@@ -1,5 +1,6 @@
 #include "game/state/rules/city/vequipmenttype.h"
 #include "game/state/gamestate.h"
+#include "game/state/tilemap/tilemap.h"
 
 namespace OpenApoc
 {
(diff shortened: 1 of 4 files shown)


FilmBoy84:
Many thanks Francisco
I'll grab the build tonight and check out each UFOPedia entry to see if behaviour is uniform across all weapons
If all OK, ill merge straight away as it's not really a fix that's going to break anything :D

FilmBoy84:
OK, i've done some testing
It appears RANGE is still double what it should be
Additionally, Fire Rates are slower than OG so there's an additional problem there

Using the Justice Missile as an example
Here's the OG stats
image
Here's the same in OpenApoc
image

FilmBoy84:
It get's even wierder on the prophet stats

OG
image

OpenApoc
image

FranciscoDA:
I've fixed the range thing in the ufopaedia. I'm not really sure about the fire rate since the UI is using the same calculations as the game itself.

To match the values displayed in vanilla we need to double the amount of ticks per second or to halve the fire_delay. Of course it could turn out that we're doing the calculations correctly and it's only the ufopaedia of the original game showing incorrect stats. In any case I'd rather have the ufopaedia reflect the real in-game stats

FilmBoy84:
@FranciscoDA The ticks per second is an interesting issue
Curiously issue #243 suggests that OpenApoc is currently working at half the speed it should be

It would be good to get @JonnyH 's opinion on this. Do we double the ticks per second and potentially resolve both issues?

FilmBoy84:
Merging for @FranciscoDA as the code is fine and no bugs found.
We need a new issue and pull request for solving the issue with rate of fire found as part of this pull
I'll try to find a solution, but as JonnyH has pointed out, there's a fundamental issue with ticks-per-second that needs some substantial work to resolve.

PR #432, opened 2018-11-03
 Fix: Display Attack Range in metres when hovering over an item 

FilmBoy84:
When mouse cursor was hovering above an item on vehicle inventory, its attack range was displayed incorrectly as a tile value, not a Metre value.

Co-Authored-By: andreycreator <andreycreator@users.noreply.github.com>


FilmBoy84:
It's probably worth noting that this also relates to @FranciscoDA 's work as it alters the same line with a slight variation on the formula

FilmBoy84:
Closing as PR #426 has now taken over it's role
Thanks @FranciscoDA <3

PR #438, opened 2018-11-20
Cheat menu

FranciscoDA:
Adds a cheat menu accessible via de ingame options. Only the following are implemented:
 * Give all research
 * Modify funds
 * Org relationship (chaos, all hostile, all friendly, utopia)


FilmBoy84:
Excellent work Francisco

As mentioned on discord, can you prefix all none-functioning cheats with an (N) to keep continuity with OpenApoc's More Options Dialogue that marks incomplete or none functioning features in the same way?

Additionally, can the Add/Subtract money be changed to a slider or value box that allows the user to specify add/subtract -£10,000,000 through +£10,000,000?

PR #439, opened 2018-11-22
fix acquiring enemy for non player vehicles

FranciscoDA:
Vehicles controlled by non-xcom organisations would not attack targets if they were non-aggressive (even if they were hostile towards them)


FilmBoy84:
Apologies in time taken to get this reviewed
All checks out
Merged for you :)

PR #442, opened 2018-11-29
Make Wolfhound APC All-Terrain Capable via (M) Option

FilmBoy84:
The Wolfhound APC can now be made all-terrain, just like the Griffon Tank


FilmBoy84:
Thanks JonnyH, merging now <3

PR #444, opened 2018-12-01
Implement cheats

FranciscoDA:
Implements the following cheats:
 * Damage inflicted multiplier
 * Damage received multiplier
 * Stat growth multiplier
 * Infinite ammo
 * Hostiles multiplier

### DIFF
--- a/game/state/rules/battle/battlemap.cpp
+++ b/game/state/rules/battle/battlemap.cpp
@@ -1,4 +1,5 @@
 #include "game/state/rules/battle/battlemap.h"
+#include "framework/configfile.h"
 #include "game/state/battle/battle.h"
 #include "game/state/battle/battledoor.h"
 #include "game/state/battle/battleitem.h"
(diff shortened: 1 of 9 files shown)


game/ui/city/buildingscreen.cpp
```diff
@@ -31,17 +31,15 @@ std::shared_future<void> loadBattleBuilding(sp<GameState> state, sp<Building> bu
                                             StateRef<Vehicle> playerVehicle)
 {
 	auto loadTask = fw().threadPoolEnqueue(
-	    [hotseat, building, state, raid, playerAgents, playerVehicle]() -> void {
+	    [hotseat, building, state, raid, playerAgents, playerVehicle]() mutable -> void {
```
JonnyH: This should probably be pulled out into another commit

game/state/rules/battle/battlemap.cpp
```diff
@@ -134,10 +135,12 @@ sp<Battle> BattleMap::createBattle(GameState &state, StateRef<Organisation> oppo
 	}
 	auto alienOrg = state.getAliens();
 
+	float hostilesMultiplier = config().getFloat("OpenApoc.Cheat.HostilesMultiplier");
+
 	int countAliens = 0;
 	for (auto &a : *aliens)
 	{
-		for (int i = 0; i < a.second; i++)
+		for (int i = 0; i < (int)(a.second * hostilesMultiplier); i++)
```
JonnyH: Try to avoid c-style casts - preference is for static_cast<int>(whatever) (or an actual rounding function)

game/ui/city/buildingscreen.cpp
FranciscoDA: I've split this between commits 02bc978 and 439288c

game/state/rules/battle/battlemap.cpp
FranciscoDA: I changed these to use std::round, which is more appropriate anyways

FilmBoy84:
I believe all of @JonnyH 's requested changes have been resolved - anything else to do?
I can confirm that menu appears to work as intended in game
Are we good to merge?

Issue #372, opened 2018-04-10
Parked Road Vehicles run out of fuel and crash!


FilmBoy84:
For some reason, immobile (parked) vehicles such as Griffon Tanks will consume fuel when not moving and parked on the cityscape

When they run out of fuel they "crash" and must be recovered

Sometimes they are lost in the crash

This should not happen...

FranciscoDA:
Can you confirm wether this bug is still present after merging #445?

JohnSmith25:
I'm using 9d93056 and so far this bug does not appear, at least not for the APC which I did used this vehicle to park at some buildings, with some additional driving of the same APC I did not noticed a fuel usage though it didn't drive for very long either.  I haven't used the Griffon much but the brief period I did used it seems to work fine didn't tested the parking with that one though.

FilmBoy84:
Can confirm that @FranciscoDA has fixed this issue
Closing accordingly
Thanks for the fix!

PR #445, opened 2018-12-03
Vehicle fixes

FranciscoDA:
* Fixes vehicles becoming stuck when auto-facing towards enemies
 * Fixes ground vehicles consuming fuel when idle (#372)

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -3300,7 +3300,7 @@ float Vehicle::getAngularSpeed() const
 {
 	// FIXME: Proper turning speed
 	// This value was hand-made to look proper on annihilators
-	constexpr float TURNING_MULT = (float)M_PI / (float)TICK_SCALE / VELOCITY_SCALE_CITY.x / 1.5f;
+	constexpr float TURNING_MULT = (float)M_PI / (float)TICK_SCALE / VELOCITY_SCALE_CITY_X / 1.5f;
 	return getSpeed() * TURNING_MULT;
 }
 
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -42,6 +42,28 @@ namespace OpenApoc
 namespace
 {
 static const float M_2xPI = 2.0f * M_PI;
+float xyToFacing(const Vec2<float> &xy)
+{
+	float a1 = acosf(-xy.y);
+	float a2 = asinf(xy.x);
+	return a2 >= 0 ? a1 : M_2xPI - a1;
+}
+Vec2<float> facingToXY(float facing) { return {sinf(facing), -cosf(facing)}; }
+// return the difference f1-f2 in the range [-pi;pi]
+// a negative return value means f1 is CCW from f2.
+float facingDistance(float f1, float f2)
+{
+	float result = f1 - f2;
+	while (result > M_PI)
+	{
+		result -= M_2xPI;
+	}
+	while (result <= -M_PI)
+	{
+		result += M_2xPI;
+	}
+	return result;
+}
 }
 
 const UString &Vehicle::getPrefix()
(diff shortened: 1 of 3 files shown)


game/state/city/vehicle.cpp
```diff
@@ -42,6 +42,28 @@ namespace OpenApoc
 namespace
 {
 static const float M_2xPI = 2.0f * M_PI;
+float xyToFacing(const Vec2<float> &xy)
```
JonnyH: I'd love this to be constexpr, but I guess if we still want c++11 compat it'll be less clear to put it on a single like (this was relaxed for c++14 IIRC)

game/state/city/vehicle.cpp
```diff
@@ -2184,36 +2205,37 @@ void Vehicle::update(GameState &state, unsigned int ticks)
...
-										// direction: right or left.
-										if (facing2d.x * target2d.y - facing2d.y * target2d.x > 0)
+										this->goalFacing = xyToFacing(target2d);
+										float d = facingDistance(this->goalFacing, this->facing);
+										// TODO: Should this nudge CCW/CW for animation purposes?
+										if (d > 0.0f)
 										{
-											this->facing += 0.06f * (float)M_PI * (float)ticks;
+											// Rotate CW towards target
+											this->angularVelocity = this->getAngularSpeed();
```
JonnyH: Would this overshoot if it's less than getAngularSpeed() away? Should we clamp this to to min(d, getAngularSpeed())?

game/state/city/vehicle.cpp
```diff
@@ -2184,36 +2205,37 @@ void Vehicle::update(GameState &state, unsigned int ticks)
...
 								{
 									// Determine target position.
 									Vec3<float> enemypos = enemy->getPosition();
-									auto facing_direction = type->directionToVector(direction);
-									Vec2<float> facing2d = {facing_direction.x, facing_direction.y};
-									Vec2<float> target2d = {enemypos.x - position.x,
-									                        enemypos.y - position.y};
-									facing2d = glm::normalize(facing2d);
-									target2d = glm::normalize(target2d);
+									Vec2<float> facing2d = facingToXY(this->facing);
```
JonnyH: Can all these be const?

game/state/city/vehicle.cpp
FranciscoDA: I'm not entirely sure TBH. I think this will cause the ticksToTurn to be 0, so the VehicleMover will snap the vehicle rotation to the goalFacing value. I've implemented this anyways because I think it's better to make this case immediately apparent rather than to rely on that convoluted logic

game/state/city/vehicle.cpp
FranciscoDA: Yes ;)

game/state/city/vehicle.cpp
```diff
@@ -3268,6 +3296,13 @@ float Vehicle::getSpeed() const
 	auto et = getEquipmentTypes();
 	return this->type->getSpeed(et.begin(), et.end());
 }
+float Vehicle::getAngularSpeed() const
+{
+	// FIXME: Proper turning speed
+	// This value was hand-made to look proper on annihilators
+	float TURNING_MULT = (float)M_PI / (float)TICK_SCALE / VELOCITY_SCALE_CITY.x / 1.5f;
```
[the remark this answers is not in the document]
FranciscoDA: After digging deeper I found that glm does define the vec constructor as constexpr:
```
// dependencies/glm/glm/detail/type_vec3.hpp:79
~/OpenApoc/dependencies/glm/glm/./ext/../detail/type_vec3.hpp:79:31: note: declared here
                GLM_FUNC_DECL GLM_CONSTEXPR vec(T a, T b, T c);
```
However, this seems to require relaxed constexpr (C++14 and upwards):
```
// dependencies/glm/glm/detail/setup.hpp:277
#if (GLM_ARCH & GLM_ARCH_SIMD_BIT) // Compiler SIMD intrinsics don't support constexpr...
#	define GLM_HAS_CONSTEXPR 0
#elif (GLM_COMPILER & GLM_COMPILER_CLANG)
#	define GLM_HAS_CONSTEXPR __has_feature(cxx_relaxed_constexpr)
#elif (GLM_LANG & GLM_LANG_CXX14_FLAG)
#	define GLM_HAS_CONSTEXPR 1
#else
#	define GLM_HAS_CONSTEXPR ((GLM_LANG & GLM_LANG_CXX0X_FLAG) && GLM_HAS_INITIALIZER_LISTS && (\
		((GLM_COMPILER & GLM_COMPILER_INTEL) && (GLM_COMPILER >= GLM_COMPILER_INTEL17)) || \
		((GLM_COMPILER & GLM_COMPILER_VC) && (GLM_COMPILER >= GLM_COMPILER_VC15))))
#endif
```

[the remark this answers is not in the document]
JonnyH: All the more reason to push to c++14 as a minimum :)

PR #448, opened 2018-12-19
Pause game when x-com arrives for investigation

FranciscoDA:
Merges #408 and adds support for multiple vehicles and agents on foot.
Should also work if:
 * The vehicle/agent was destroyed while traveling
 * The vehicle/agent changed mission while traveling

### DIFF
--- a/game/state/city/building.h
+++ b/game/state/city/building.h
@@ -66,6 +66,7 @@ class Building : public StateObject, public std::enable_shared_from_this<Buildin
 	unsigned ticksDetectionTimeOut = 0;
 	unsigned ticksDetectionAttemptAccumulated = 0;
 	bool detected = false;
+	bool investigate = false;
 	// Unlocks when successful at raiding this
 	std::list<StateRef<ResearchTopic>> researchUnlock;
 	// Access to building
--- a/game/state/city/building.h
+++ b/game/state/city/building.h
@@ -66,7 +66,7 @@ class Building : public StateObject, public std::enable_shared_from_this<Buildin
 	unsigned ticksDetectionTimeOut = 0;
 	unsigned ticksDetectionAttemptAccumulated = 0;
 	bool detected = false;
-	bool investigate = false;
+	int investigate = 0;
 	// Unlocks when successful at raiding this
 	std::list<StateRef<ResearchTopic>> researchUnlock;
 	// Access to building
@@ -74,6 +74,8 @@ class Building : public StateObject, public std::enable_shared_from_this<Buildin
 	// Victory when successful at raiding this
 	bool victory = false;
 
+	// may fire a 'commence investigation' event
+	void decreaseInvestigateCount(GameState &state);
 	bool hasAliens() const;
 	void updateDetection(GameState &state, unsigned int ticks);
 	void updateCargo(GameState &state);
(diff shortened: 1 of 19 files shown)


FranciscoDA:
Relevant note:
The clang4 build was failing because the algorithm for generating the starting base tries random layouts by adding and removing facilities. Removing living quarters while there are agents in the building should fail, therefore causing an error in the base generation.
It's possible for a starting base to have agents inside it if they are spawned as hireable agents (this happens during `GameState::fillOrgStartingProperty`). Moreover, there's nothing preventing a hireable agent to spawn inside an XCom base!
This edge case is dependent on the seed used in the rng. Here's one that causes such behaviour:
 * Seed: `1545206368882`

I've worked around this issue by generating the starting base (by calling `GameState::fillPlayerStartingProperty`) before the hireable agents are created.
A more complete fix should prevent agents to spawn from XCom bases or purchaseable buildings.

game/state/city/building.h
```diff
@@ -66,13 +66,16 @@ class Building : public StateObject, public std::enable_shared_from_this<Buildin
 	unsigned ticksDetectionTimeOut = 0;
 	unsigned ticksDetectionAttemptAccumulated = 0;
 	bool detected = false;
+	int investigate = 0;
```
JonnyH: 'investigate' doesn't seem that descriptive to me.

Maybe something more like "enrotue_investigator_count"?

JonnyH: this method also forbids ever having multiple different investigations en route to a single building. While this matches vanilla behaviour - you can only ever investigate aliens, and you can realistically never get a second investigation spawned before the first arrives and is resolved (either investigated or aborted) - I can see this as something modders may want to change - for example having multiple 'enemies' or investigation 'types'.

So "Good for 1.0", as I can't suggest an obvious improvement :)

Acutally, that might be interesting to see if you can get that to happen in vanilla - perhaps by 'trapping' an investigator behind unrouteable tiles and letting another 'alien spotted' event trigger?

FilmBoy84: There is actually a rare occurrence in OG where that can occur
If an agent on route to an investigation, in a people tube that has it's ends severed, or on a road that has segments either side destroyed that agent will be trapped
This allows multiple alien incidents to appear if the aliens split their forces between multiple buildings (more likely with larger UFO drops) or the alien life-cycle results in more city based alien spawns.

FilmBoy84: Ah, same building in OG, no that doesn't happen (at least not that I've seen).

However it can happen in Beta2... If the first agent gets to the first mission and clears it, the building appears to count as cleared for purposes of alien infiltration on the parent organisation (which makes sense, all aliens are dead or have ran to other buildings during the mission)...

But as soon as the agents assigned to a second mission arrives, another tactical mission will start; it will however be void of any aliens and X-COM get's an instant "all hostile units have left the combat zone" prompt, it will then end the mission and automatically collect any equipment left in the building (such as elerium pods, guns, craft ammo, etc. from the map) and X-COM will get a positive score as a result...

As far as I understand it, the reason for the Beta2 behaviour is a bug created by the removal of the espionage functions.... At one time you would want that second unit to still check the cleared building as VIPs may have been identified within as a result of the agents from the first mission spotting them inside. Once the second mission starts the player would then have a chance to kill/stun/tag the VIP and leave, but because VIPs are no longer a thing in Beta2 the game doesn't acknowledge any hostiles present and ends the mission.

So i agree with your view that for 1.0 not having to allow the option of two missions in the same building is the way forwards. But we should also consider that modders may want to add this functionality in the future... Particularly as adding back in espionage is one frequently cited mod possibility.

FranciscoDA:
`Building::investigate` was renamed to `Building::pendingInvestigatorCount` in a505ea0
Eventually we should add a serializable state object to represent an ongoing investigation to allow multiple investigations in the same building and maybe vehicles

Issue #284, opened 2017-11-30
Item falls off the end of the world CTD


FilmBoy84:
Whilst running a CoS raid under 0.1-117

![image](https://user-images.githubusercontent.com/33913088/33438902-fef7b6c8-d5e3-11e7-815f-8d1bc68a78d1.png)

![image](https://user-images.githubusercontent.com/33913088/33438945-203338b2-d5e4-11e7-9832-e5f537202d47.png)

![image](https://user-images.githubusercontent.com/33913088/33438961-280c6d38-d5e4-11e7-8efe-0b08612ad425.png)

FilmBoy84:
Another variant of this bug.

![image](https://user-images.githubusercontent.com/33913088/34469644-aaf0ee18-ef1b-11e7-93a2-ae7be9720cd4.png)
![image](https://user-images.githubusercontent.com/33913088/34469645-b111bdd6-ef1b-11e7-8569-593bccc6542e.png)
![image](https://user-images.githubusercontent.com/33913088/34469647-b6423056-ef1b-11e7-9b3b-3d14aeb4fc40.png)

FilmBoy84:
Cause of bug found in issue #366 Please see this issue also for details

FilmBoy84:
All fixed as of today
Closing

PR #455, opened 2019-01-01
fix item fall through gravlift floor

### LINKED ISSUE #366: Item fell of the end of the world - Grav lift
I tried this multiple times. When You fire a brainsucker launcher, throw an item in a grav lift or simply put it down there is a error - Item fell of the end of the world

image

kaja47:
This should fix #366 and #284.
Bottom of gravlift couldn't provide support for items and they fell through the floor, possibly out of the world.

### DIFF
--- a/game/state/tilemap/tile.cpp
+++ b/game/state/tilemap/tile.cpp
@@ -376,8 +376,7 @@ void Tile::updateBattlescapeParameters()
 				continue;
 			}
 			height = std::max(height, (float)mp->type->height);
-			if ((mp->type->floor || o->getType() == TileObject::Type::Feature) &&
-			    !mp->type->gravlift)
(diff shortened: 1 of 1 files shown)


FranciscoDA:
Hey, can you run `clang-format -i -style=file game/state/tilemap/tile.cpp` this will re-format the file and make so it will pass the linting check. Then you have to add it and make a new commit
I also suggest you do a `git rebase -i HEAD~2` and mark the formatting commit as a fixup of the previous one (this way we keep the commit log clean of formatting commits)

FilmBoy84:
Just a heads up, some playing with the "Alternate Gravlift Sound" option, under more options may be required as there may be crash (can't be certain myself if it's this causing it)

I've had several crashes with the option ticked when running builds with this fix applied

FilmBoy84:
Now fixed in PR #480 
Closing this as it's no longer needed

Issue #460, opened 2019-01-03
Clicking on Cityscape Map causes CTD


FilmBoy84:
Please load attached game and click on the map
Game will CTD without writing to log!
[save_SuperHuman 1.zip](https://github.com/OpenApoc/OpenApoc/files/2723463/save_SuperHuman.1.zip)

FilmBoy84:
Interesting fix to this
Load the save
Press F1 (for debug)
Press F4 (to show all aliens on map)
Click the map icon
Game won't crash

If you do anything with the map before following the above, it CTDs

JonnyH:
Quickly ran this through a debugger:

It's crashing at citytileview.cpp:745

In the save, it seems that "VEHICLE_3864" has somehow ended up with an AttackVehicle mission with a null targetVehicle

FilmBoy84:
@JonnyH Ah, many thanks, then it's also related to issue #456 
Those vehicles are stuck in the building on both save games
I just realised that they became stuck when the UFO they were attacking left the city and back to alien dimension or were among the UFOs destroyed before they left the building

It seems that there is a bug in dealing with how vehicles clear their orders when the target entity either leaves the dimension that it's in, or is destroyed

Will update #456 accordingly :)

FranciscoDA:
I think this is a rendering issue. It seems it's possible for a vehicle to have a mission with a null target vehicle, but the mission is popped almost immediately.
In this case a bunch of vehicles are stuck in a takeoff mission (as in #456) so there's no way to pop the attack vehicle mission

FilmBoy84:
Issue appears to be fixed by FranciscoDA's work
Many thanks

PR #469, opened 2019-01-07
fix map rendering with null targetVehicle in AttackVehicle mission

FranciscoDA:
Prevents crash to desktop as described in #460

### DIFF
--- a/game/ui/tileview/citytileview.cpp
+++ b/game/ui/tileview/citytileview.cpp
@@ -740,6 +740,8 @@ void CityTileView::render()
 					{
 						case VehicleMission::MissionType::AttackVehicle:
 						{
+							if (!m->targetVehicle)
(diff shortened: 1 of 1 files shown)


game/ui/tileview/citytileview.cpp
```diff
@@ -740,15 +740,21 @@ void CityTileView::render()
 					{
 						case VehicleMission::MissionType::AttackVehicle:
 						{
-							vehiclesUnderAttack.insert(m->targetVehicle);
```
JonnyH: Would this be cleaner as:
if (!m->targetVehicle)
    break;

rendering_code();

FranciscoDA: Fair enough, that will also reduce the indentation level

Issue #456, opened 2019-01-01
Vehicles Get Stuck and cannot accept orders or Manual Control!


BeornTB:
After the first UFO invasion, my Valkyrie refuses to fly.

Attached log.txt and linked .save
[openapoc_log.txt](https://github.com/OpenApoc/OpenApoc/files/2719503/openapoc_log.txt)
https://www.dropbox.com/s/pgzxz15zso3o32k/save_Valkyrie%20Stuck.save?dl=0

Thanks!

BeornTB:
2nd example: valkyrie became stuck during the UFO invasion. This time you can see it in the city.

https://www.dropbox.com/s/b219toyrwkrmi1v/save_Valkyrie%20Stuck%20Again.save?dl=0
[openapoc_log.txt](https://github.com/OpenApoc/OpenApoc/files/2719514/openapoc_log.txt)

FilmBoy84:
Hmmmm, i can get the valkyrie to recover the UFO under the road
![image](https://user-images.githubusercontent.com/33913088/50632885-2f28ba80-0f41-11e9-9339-e45e4e46c828.png)
![image](https://user-images.githubusercontent.com/33913088/50632891-32bc4180-0f41-11e9-8b67-52a19c036c18.png)

And the people tube
![image](https://user-images.githubusercontent.com/33913088/50632911-45cf1180-0f41-11e9-831e-805033de52fb.png)

You are correct with the other saves showing a stuck craft though
Not even manual control (n) will fix it :(

FilmBoy84:
Hmmmm, an update on the issue regarding stuck vehicles
I have a whole building full of vehicles stuck on this save!!!
[save_SuperHuman 1.zip](https://github.com/OpenApoc/OpenApoc/files/2723450/save_SuperHuman.1.zip)

FilmBoy84:
In relation to this issue

It appears vehicles became stuck when the UFO they were attacking left the city and back to alien dimension or were among the UFOs destroyed before they left the building

It seems that there is a bug in dealing with how vehicles clear their orders when the target entity either leaves the dimension that it's in, or is destroyed

See my superhuman save above ;) as well as @BeornTB 's save too ;)

FilmBoy84:
Issue appears to be fixed by FranciscoDA's work
Many thanks

PR #470, opened 2019-01-08
Fix vehicles becoming stuck in buildings when clearing missions

FranciscoDA:
When calling `Vehicle::setMission`, it's possible for some missions to reach the front in the vehicle's mission list without being properly started. Here's an example:

Before clearMissions:
```
front          back
  v              v
Snooze, TakeOff, X
```

After clearMissions:
```
TakeOff
```

After addMission:
```
TakeOff, X
```

Here, the TakeOff mission was never started although it is at the front of the list. X can be another mission like AttackVehicle, GotoLocation, GotoBuilding, etc.
This may be the cause of #456, although it won't fix the already broken saves.

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -3062,7 +3062,8 @@ void Vehicle::setManualFirePosition(const Vec3<float> &pos)
 	manualFire = true;
 }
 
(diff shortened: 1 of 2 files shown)


FilmBoy84:
All seems good Francisco, i'll await Jonny's approval and then merge
Testing reveals it to work as intended

PR #473, opened 2019-01-09
fix EconomyInfo::update scaling

FranciscoDA:


FilmBoy84:
Just with regards to pricing changes in vanilla
They appear to change according to supply and demand by a few percent a week
**THIS IS CUMULATIVE**

Additionally, organisation relations appear to have an impact that is also cumulative on the pricing

So you have both supply/demand and relationship affecting price
You can, to a degree, compensate for an increase in demand raising the cost of an item by having a good relationship with the supplier (that reduces cost)

Can this be fixed in the code too?

game/state/city/economyinfo.cpp
```diff
@@ -35,7 +35,7 @@ bool EconomyInfo::update(GameState &state, bool xcom)
 		// Price update
 		if (soldThisWeek > 2 * maxStock)
 		{
-			currentPrice = currentPrice * randBoundsInclusive(state.rng, 85, 5);
+			currentPrice = currentPrice * randBoundsInclusive(state.rng, 85, 95);
```
JonnyH: These values really shouldn't be added to the code - instead maybe make it somewhere in the GameState so it can be changed with a mod?

game/state/city/economyinfo.cpp
```diff
@@ -45,6 +45,7 @@ bool EconomyInfo::update(GameState &state, bool xcom)
 		{
 			currentPrice = currentPrice * randBoundsInclusive(state.rng, 95, 97);
 		}
+		currentPrice = std::lround(static_cast<double>(currentPrice) / 100.0);
```
JonnyH: If this 'currentPrice' is actually 100* the price (to avoid floating point weirdness I guess?), please name it something different or to the /100 immediately - as otherwise I can see someone trying to modify this in the future getting confused.

FranciscoDA: The `currentPrice` is a member of `EconomyInfo` i.e the actual price of the item. The /100 part was missing so I added it to compensate since `randBoundsInclusive` only works with integer values.
I just changed that in 01bb029 to use uniform_real_distribution, otherwise `randBoundsInclusive` would only allow increases/decreases in price by whole percents like 1%, 2%, 3%, etc

FranciscoDA:
Closing as #491 supersedes this

PR #479, opened 2019-01-12
Set player funds for each difficulty patch

FranciscoDA:


FilmBoy84:
Looking great Francisco
I'm happy to merge if JonnyH happy?

PR #480, opened 2019-01-15
Fix for item falling through grav-lift floor

FilmBoy84:
An updated version of PR #455 , thanks to @kaja47 for identifying the original line at fault


JonnyH:
The APC stuff feels a little out of place in this PR - is that fallout of a bad merge?

The gravlift fix itself LGTM

FilmBoy84:
My mistake, it's because i updated my branch from master
If you merge this PR it won't actually change any more files that the gravlift tile.cpp as the other commits are already dealt with

FilmBoy84:
Rebased and merged in PR #481 
Thanks JonnyH

Issue #485, opened 2019-01-18
Hard difficulty: Base layouts and availability doesn't match the vanilla game


BeornTB:
As compared to https://www.ufopaedia.org/index.php/Base_Locations_(Apocalypse)#Hard_Level

WH 5 (furthest south-east) is correctly government-owned and named but unavailable for purchase
WH8 has the layout of WH5
Campesino has a layout normally found on novice only (the crab map of both Heavenly towers and WH7)

FilmBoy84:
As pointed out on Discord by Beorn
On EASY Warehouse 9 has the wrong template
It should look like this

![image](https://user-images.githubusercontent.com/33913088/51400620-c8122500-1b40-11e9-8f13-0ed85a5f0a19.png)


For resolving these issues, we need to check all bases against the information at
https://www.ufopaedia.org/index.php/Base_Locations_(Apocalypse)

Jarskih:
Ok I will fix that as well.

JonnyH:
Probably best to check against the game - I think there's been a few times the wiki isn't as accurate as you'd hope.

FilmBoy84:
Agreed
Ill aim to confirm later

FilmBoy84:
Hard difficulty fixed in pr #487 , closing issue as #488 and #489 cover other problems found

PR #487, opened 2019-01-18
Fix wrong warehouse layouts on hard difficulty

Jarskih:
#485 Added missing warehouse and changed the layouts to correct one.


FilmBoy84:
Many thanks @Jarskih I'll merge once the build is complete
We also have WH9 on easy to fix, but best to do that as a separate commit TBH as it's a slightly different issue

PR #496, opened 2019-01-21
Fix chrysalis autopsy showing megaspawn picture

### LINKED ISSUE #402: Chrysalis Autopsy Research Project shows Megaspawn Alien
The chrysalis autopsy project uses the wrong artwork
image

FranciscoDA:
Fixes #402


FranciscoDA:
I would like to move the hardcoded attributes in `/tools/extractors/extract_agent_types.cpp` to the common_path directory. However it seems the extractor fills in some attributes for shadow_offset and dropped_offset, hence the alien containment items would still need to be declared in the extractor, unless we can omit those attributes?

Issue #446, opened 2018-12-03
Base Invasion (Alien or Org) at an X-COM Base with no agents causes CTD


FilmBoy84:
I'm merging issue #320 and issue #418 into this bug report as it's still current

If an organisation, or the aliens, invades an X-COM base that has no agents present, even if that base has security stations (and therefore can protect itself), the game will first state that no X-COM agents are alive or conscious, then crash as it attempts to end the mission as a lose for X-COM

To facilitate OG functionality the base should be "lost" without starting a mission. It's facilities destroyed and X-COM losing the building (though they can re-purchase later)

A better proposal for OpenApoc would be that if security stations are present, start a tactical mission. This allows the player the opportunity to see the invaders mown down by security station fire and maybe save the base.

If the player aborts the mission having not killed the invaders then the base is lost

If there are NO security stations, then the base is lost as normal for the OG

Original Report by @Darkestaxe1 Issue 320
_"to reproduce, make a new game on medium,
buy the $4800 slums base in the upper-right of the map, (I started building stores and living quarters)
pass time until aliens invade it on day 2"_

Original Report by @DisDis Issue 418
_"At the 0 move. That is, immediately after the start. The message "Opponent killed or unconscious" was displayed.
After that game is crash.
Log and save:
https://github.com/DisDis/OpenApocBugfiles/tree/master/1"_

FilmBoy84:
Another iteration of this bug, a base attack has started with NO HOSTILES generated
The game reports that all hostile units are dead, unconscious or left and crashes
Save attached
[save_SuperHuman 1.zip](https://github.com/OpenApoc/OpenApoc/files/2723609/save_SuperHuman.1.zip)

FilmBoy84:
From@BeornTB

`Sorry, no save or log.

RT base defense, 2 androids 2 defense towers vs no opponents. Instant victory, CTD.`

BeornTB:
Working base defense save:
https://www.dropbox.com/s/qexunozphx3wtfz/save_336%20TB%20Base%20D.save?dl=0

Jarskih:
I think this crash is caused by base defense turrets. They are trying to go back to base after battle. I am puzzled though why base defense starts without enemies?

FilmBoy84:
This issue is no longer replicatable
It appears @FranciscoDA fixed it
Great job sir!

PR #499, opened 2019-01-23
Base defense fixes

### LINKED ISSUE #463: CTD when unpausing cityscape, coming back from a mission
https://www.dropbox.com/s/0gks2xf898c15ar/save_318%20CTD%20mission%20end.zip?dl=0

Could be related to Bug #441 

Build 318, save and log in .zip

Mission success screen incorrectly states that there are no live alien captured, then the game crashes as soon as I unpause the cityscape.

I've played this mission 4-5 times, it crashes every time.
The first time, I lost a soldier to a popper and the CTD said something about being unable to find soldier 9, then didn't manage to "limp along". IIRC that crash occured before having a victory screen, unlike the current save.

2 things are fishy: the post-mission errors (no live aliens, dead soldier can't be found when there was one) and the CTD when unpausing (could be the valk or any other vehicle error).

Btw, I still have bugs with UFO recovery on Valkyries. However, saving and reloading fixes it. There's still something fishy here as well but it's workable for playtesting.

FranciscoDA:
Fixes alien movement roll triggering base defense missions without aliens actually moving
Hopefully fixes base defense crashing #446 ?

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -2649,8 +2649,9 @@ void Battle::finishBattle(GameState &state)
 			}
 			loot.push_back(e);
 		}
-		// Send home if not on vehicle
-		if (!u.second->agent->currentVehicle && !state.current_battle->skirmish)
+		// Send home if not on vehicle. Note that some units may not have a home building
+		if (!u.second->agent->currentVehicle && !state.current_battle->skirmish &&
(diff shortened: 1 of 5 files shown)


FilmBoy84:
This seems to fix the problems in testing; even retroactively with my broken Superhuman save
Merging
Thanks FranciscoDA <3

PR #507, opened 2019-02-04
List control tweaks

FranciscoDA:
* Allow mouse-move interactions with captured scrollbar even outside of its' bounds (easier dragging, especially near the left end of the scrollbar)
 * Adjust the scroll limits according to the parent listbox's contents (easier scrolling in a listbox with many items)

### DIFF
--- a/forms/listbox.cpp
+++ b/forms/listbox.cpp
@@ -205,6 +205,25 @@ void ListBox::update()
 	}
 	if (scroller)
 	{
+		size_t mainAxisSize = 0;
+		switch (ListOrientation)
+		{
+			case Orientation::Vertical:
+				mainAxisSize = Size.y;
+				break;
+			case Orientation::Horizontal:
+				mainAxisSize = Size.x;
+				break;
+			default:
+				LogWarning("Unknown ListBox::ListOrientation value: %d",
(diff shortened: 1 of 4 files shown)


forms/listbox.cpp
```diff
@@ -205,6 +205,24 @@ void ListBox::update()
...
+		switch (ListOrientation)
+		{
+			case Orientation::Vertical:
+				mainAxisSize = Size.y;
+				break;
+			case Orientation::Horizontal:
+				mainAxisSize = Size.x;
+				break;
+			default:
+				// shouldn't happen
```
JonnyH: If this shows a programming error, maybe a LogWarning (or a LogError if it *can never happen* without magic weird int->enum conversions or memory corruption/un-inited values as there's only 2 values in the enum?)

forms/listbox.cpp
```diff
@@ -205,6 +205,24 @@ void ListBox::update()
...
+				return;
+		}
+		size_t mainAxisContentSize =
+		    ItemSize * Controls.size() + ItemSpacing * (Controls.size() - 1);
```
JonnyH: Could this go negative and blow up if Controls.size() is zero? As size_t is unsigned it'll be MAX_SIZE_T?

PR #508, opened 2019-02-04
Tinted label

FranciscoDA:
Adds support for tinted labels:
 * Recruit screen
recruit screen
 * Lab screen
lab screen
 * Buy and sell screen (hostile manufacturer)
buy and sell

### DIFF
--- a/library/colour.cpp
+++ b/library/colour.cpp
@@ -1,0 +1,74 @@
+#include "colour.h"
+
+#include <algorithm>
+#include <array>
+#include <map>
+
+#include "framework/logger.h"
+
+namespace OpenApoc
+{
+
+namespace
+{
+
+static const std::map<UString, Colour> html4Colours{
+    {"white", {255, 255, 255}}, {"silver", {192, 192, 192}}, {"gray", {128, 128, 128}},
+    {"black", {0, 0, 0}},       {"red", {255, 0, 0}},        {"maroon", {128, 0, 0}},
+    {"yellow", {255, 255, 0}},  {"olive", {128, 128, 0}},    {"lime", {0, 255, 0}},
+    {"green", {0, 255, 0}},     {"aqua", {0, 255, 255}},     {"teal", {0, 128, 128}},
+    {"blue", {0, 0, 255}},      {"navy", {0, 0, 128}},       {"fuchsia", {255, 0, 255}},
+    {"purple", {128, 0, 128}},
+};
+}
+
+Colour Colour::FromHtmlName(const UString &name)
+{
+	auto it = html4Colours.find(name.toLower());
+	if (it != html4Colours.end())
+		return it->second;
+	LogWarning("Unknown colour name: %s", name);
+	return {0, 0, 0, 0};
+}
+
+Colour Colour::FromHex(const UString &hexcode)
+{
+	UString hexcode_lower = hexcode.toLower();
+	if (hexcode_lower.empty())
+	{
+		LogWarning("Empty colour hex code");
+		return {0, 0, 0};
+	}
+	// invalid initial character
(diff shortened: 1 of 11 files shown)


library/colour.cpp
```diff
@@ -0,0 +1,63 @@
+#include <algorithm>
+#include "colour.h"
+
+namespace OpenApoc {
+
+Colour Colour::FromHtmlName(const UString& name)
+{
+	UString name_lower = name.toLower();
```
JonnyH: Would this be neater with a static map<UString, Color> Lookup table?

library/colour.cpp
```diff
@@ -0,0 +1,63 @@
...
+	}
+	uint8_t digits[hexcode_lower.length()-1];
+	std::transform(++hexcode_lower.begin(), hexcode_lower.end(), digits, [](int cp) { return cp-'0'; });
+	if (hexcode_lower.length() == 4)
+		return {
+			static_cast<uint8_t>(digits[0] * 0x11U),
+			static_cast<uint8_t>(digits[1] * 0x11U),
+			static_cast<uint8_t>(digits[2] * 0x11U)
+		};
+	else
```
JonnyH: I might prefer an "else if (length == 7)" or a comment on that the other case is - at first glance I didn't see the early-out for the non-4 non-7 lengths

library/colour.cpp
```diff
@@ -0,0 +1,63 @@
...
+	else if (*hexcode_lower.begin() != '#')
+	{
+		return {0, 0, 0};
+	}
+	// invalid characters
+	else if (std::any_of(++hexcode_lower.begin(), hexcode_lower.end(), isxdigit))
+	{
+		return {0, 0, 0};
+	}
+	uint8_t digits[hexcode_lower.length()-1];
```
JonnyH: Aren't VLAs non-standard c++? And then considered evil? (possibly opening up stack-smashing issues, and support issues going forward)

Probably a vector<uint8_t> would be a better solution here.

Or even better - as it's a known max size - just make a std::array<6> and use that?

library/colour.cpp
```diff
@@ -0,0 +1,63 @@
...
+	else if (name_lower == "lime")    return {0,   255, 0};
+	else if (name_lower == "green")   return {0,   255, 0};
+	else if (name_lower == "aqua")    return {0,   255, 255};
+	else if (name_lower == "teal")    return {0,   128, 128};
+	else if (name_lower == "blue")    return {0,   0,   255};
+	else if (name_lower == "navy")    return {0,   0,   128};
+	else if (name_lower == "fuchsia") return {255, 0,   255};
+	else if (name_lower == "purple")  return {128, 0,   128};
+	// unknown color name
+	return {0, 0, 0};
```
JonnyH: Maybe logWarning()? Might be useful for modders etc. to be explicitly told they're doing something unexpected

library/colour.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+Colour Colour::FromHex(const UString& hexcode)
+{
+	UString hexcode_lower = hexcode.toLower();
+	// invalid length
+	if (hexcode_lower.length() != 4 && hexcode_lower.length() != 7)
+	{
+		return {0, 0, 0};
```
JonnyH: Maybe LogWarning() on all the invalid color string paths?

FilmBoy84:
Just doing a quick review of the code before submitting; noticed that the skill tint is a very light grey; whilst it appears OK to me in the game, it might require tweaking in the future to a slightly darker grey for more contrast. Maybe 150,150,150? We'll see what player feedback says nonetheless

Merging as is and as requested as testing checks out :)

PR #510, opened 2019-02-06
Add tooltips

FranciscoDA:
Adds configurable tooltips (font, background color, border color)
Note that only the tooltip for the 'Bases' button in the cityscape was implemented. The rest of the tooltips should be added in the appropiate form files

### DIFF
--- a/library/strings.cpp
+++ b/library/strings.cpp
@@ -351,6 +351,39 @@ bool UString::endsWith(const UString &suffix) const
 	return boost::ends_with(str(), suffix.str());
 }
 
+UString UString::trimLeft() const
+{
+	auto first = begin();
+	auto last = end();
+	while (first != last && Strings::isWhiteSpace(*first))
+		++first;
+	return {first, last};
+}
+UString UString::trimRight() const
+{
+	auto first = begin();
+	auto last = end();
+	if (first == last)
+		return {};
+	--last;
+	while (first != last && Strings::isWhiteSpace(*last))
+		--last;
+	return {first, ++last};
+}
+UString UString::trim() const
+{
+	auto first = begin();
+	auto last = end();
+	while (first != last && Strings::isWhiteSpace(*first))
+		++first;
+	if (first == last)
+		return {};
+	--last;
+	while (first != last && Strings::isWhiteSpace(*last))
+		--last;
+	return {first, ++last};
+}
+
 UString::ConstIterator UString::begin() const { return UString::ConstIterator(*this, 0); }
 
 UString::ConstIterator UString::end() const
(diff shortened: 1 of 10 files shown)


library/colour.cpp
```diff
@@ -46,14 +46,14 @@ Colour Colour::FromHex(const UString &hexcode)
 		return {0, 0, 0};
 	}
 	// invalid characters
-	else if (std::any_of(++hexcode_lower.begin(), hexcode_lower.end(), isxdigit))
+	else if (std::none_of(++hexcode_lower.begin(), hexcode_lower.end(), isxdigit))
```
JonnyH: Was this a bug that made it through the last PR? If this is intended probably best to pull it out to a new PR, as it's not really "tooltip"-specific

library/colour.cpp
```diff
@@ -46,14 +46,14 @@ Colour Colour::FromHex(const UString &hexcode)
...
 	{
 		LogWarning("Invalid characters found in hex triplet: %s", hexcode);
 		return {0, 0, 0};
 	}
 	std::array<int, 6> digits;
 	std::transform(++hexcode_lower.begin(), hexcode_lower.end(), digits.begin(),
-	               [](int cp) { return cp - '0'; });
+	               [](int cp) { return '0' <= cp && cp <= '9' ? cp - '0' : cp - 'a' + 10; });
```
JonnyH: Another non-tooltip fix?

library/strings.cpp
```diff
@@ -369,6 +402,19 @@ UString::ConstIterator UString::ConstIterator::operator++()
 	return *this;
 }
 
+UString::ConstIterator UString::ConstIterator::operator--()
+{
+	const char *ptr = s.cStr();
+	ptr += --this->offset;
+	while (static_cast<unsigned char>(*ptr) >= 0x10000000 &&
```
JonnyH: This looks broken - an unsigned char can never be >= 0x1000000?

One cool thing about utf8 is you can find the start byte of all character as they're the only ones with the highest bit unset.

So this can be simplified to:
while (*ptr & 0x80) { go back 1 byte }
?

FranciscoDA: You're right, it should have been [0**b**10000000 ; 0**b**11000000), which is the range for continuation bytes. Unfortunately the first byte in multi-byte characters have their highest bit set (unlike single-byte chars), but they are always outside the aforemetioned range

JonnyH: Yeah, brain fart - I mean bit 2 (0b01000000) is only set on the first of a multibyte sequence. I somehow neglected single-byte chars, and got the wrong bit on top :)

So the first byte either has (byte & 0b10000000) for a single byte sequence, or (byte & 0b01000000) for the first of a multibyte. but I guess that's equivalent to the numerical comparisons)

JonnyH:
LGTM, not actively tested myself yet though :)

PR #511, opened 2019-02-07
Fix hexcode triplet decoding in colour.cpp

FranciscoDA:
Split from #510


FilmBoy84:
Francisco split fropm #510 as JonnyH requested, have checked the build and all seems good
Ready to merge pending approval :)

PR #512, opened 2019-02-07
serialize VehicleMission::MissionType::ArriveFromDimensionGate

FranciscoDA:
Should fix the error regarding unknown vehicle mission type when loading a savegame at the start of a UFO incursion (reported by @BeornTB on discord)

### DIFF
--- a/game/state/gamestate_serialize.xml
+++ b/game/state/gamestate_serialize.xml
@@ -1453,6 +1453,7 @@
 		<value>Teleport</value>
 		<value>SelfDestruct</value>
 		<value>DepartToSpace</value>
+		<value>ArriveFromDimensionGate</value>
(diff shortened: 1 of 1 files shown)


JonnyH:
Maybe we should generate the enums from the xml, to guarantee that they'll never be missed?

I actually looked into hooking into clang to pull stuff like this from the c++ directly, so we wouldn't need the serialisation xml, but it was a *lot* of work and seemed extremely reliant on specific custom builds of clang - which feels like a lot of work to avoid a "small" amount of effort....

FranciscoDA:
There are quite a lot of "smart enum" libraries that could do this but they all rely on preprocessor directives and that sort of uglyness, so I'm afraid the cure may be worse than the disease

PR #526, opened 2019-03-07
refactor CityView to use subforms

FranciscoDA:
* Link subforms in XML instead of hardcoding the names of the forms for the tabs
 * Sets correct order for event handling (active tab is processed after the base form). This should make tooltips work in the base form

### DIFF
--- a/game/ui/tileview/cityview.cpp
+++ b/game/ui/tileview/cityview.cpp
@@ -103,11 +103,6 @@ std::shared_future<void> loadBattleBuilding(sp<GameState> state, sp<Building> bu
 	return loadTask;
 }
 
-static const std::vector<UString> TAB_FORM_NAMES = {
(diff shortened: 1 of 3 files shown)


game/ui/tileview/cityview.cpp
```diff
@@ -901,62 +905,23 @@ CityView::CityView(sp<GameState> state)
 	    ->addCallback(FormEventType::ButtonClick,
 	                  [this](Event *) { setSelectionState(CitySelectionState::Normal); });
 	baseForm->findControlTyped<RadioButton>("BUTTON_SPEED1")->setChecked(true);
-	for (auto &formName : TAB_FORM_NAMES)
+	for (int i = 0; i < 8; ++i)
```
SupSuper: This should probably use a constant. `i < uiTabs.size()` ?

PR #527, opened 2019-03-08
Use CMake for Windows builds

SupSuper:
I think it's time to get rid of one of the bigger pain points for new coders: maintaining both CMake and VC projects. Normally I favor VC projects on Windows, but given OpenApoc is split into a monstrosity of custom targets, tests, commands and more, I don't think it's worth it. :P And now that VS2017's CMake support is well and truly mature (and VS2019 is coming...), it's fairly painless to use on Windows.

So this PR takes care of:
- Getting rid of all the old manually made VCprojs, Batch files, etc.
- Updates Appveyor to use CMake instead. The results seem consistent, you can see them here: https://ci.appveyor.com/project/SupSuper/openapoc
- Fetches dependencies from Vcpkg instead of Nuget, so they are properly cached and up-to-date.
- Fixes any inconsistencies left in CMakefiles in Windows.

Is there any other stuff I should add? Optimizations? Enable/disable tests? Etc.
Might also be time to re-examine C++17 again: https://github.com/OpenApoc/OpenApoc/pull/391


CMakeLists.txt
```diff
@@ -70,26 +75,19 @@ if (LTO)
...
-		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
-elseif (COMPILER_SUPPORTS_CXX14)
-		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
-elseif (COMPILER_SUPPORTS_CXX11)
-		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-else()
-		# Other compilers may use different option flags, or not support switching standard versions, so try anyway
-		message(WARNING "Couldn't find a c++ version this compiler supports, trying with default options")
-endif()
+set(CMAKE_CXX_STANDARD 17)
```
FranciscoDA: Maybe we should lower this to C++14. I think clang 4.0 doesn't support C++17 and it might be either using a C++03 fallback or maybe C++1x/z/y.
Either way it doesn't get picked up by glm as supporting C++11 and causes the Travis build to fail.

Or maybe we update clang to 5.0+?

JonnyH: My reading of cmake's cods, CMAKE_CXX_STANDARD should automatically set the "latest supported version, up to 17" when specifying "17" here - so it should support c++14/c++11 (or ever just c++98) compilers. CMAKE_CXX_STANDARD_REQUIRED is the option to set to say "this exact version is /needed/".

Naturally, this 'decay' is dangerous if you actually start using c++17 features without the correct version check macros (which I /really/ want to avoid)

The problem I have with that is there's not a huge amount of point enabling c++17 if we can't actually use any of the c++17 features. Some (like the <variant> and <optional> library features) are pretty easy to wrap with a different implementation, but core language stuff would be near impossible.

I personally would lean towards cutting out older compilers, we're still in "only useful for hardcore devs" mode, so this is the perfect time to actually change core stuff like this.

JonnyH: ^^ Additionally, I can't remember /why/ I did that if(c++17) else if (c++14)... chain in the first place. I remember trying the CMAKE_CXX_STANDARD property and deciding it didn't quite work for what I wanted it to do, but can't for the life of me remember /what/ that was - or if it was just a bug in the version I happened to be using or something dumb.

Issue #541, opened 2019-03-21
Cancel button and whole pop up window is click through


makus82:
When you on Vehicle tab and select vehicle and use mouse to give command to atack other vehicle, for example, and then click to cancel button its pop up message to atack building

Seems its click through cancel button and click at atack building button instead
But i can cancel this by hotkey ESC

makus82:
![image](https://user-images.githubusercontent.com/11581207/54777593-4bbcc080-4c1b-11e9-81bb-2eba2f7ff110.png)

makus82:
whole pop up window wich description is click through so i can press at any button under

makus82:
https://ci.appveyor.com/project/OpenApoc/openapoc/builds/23273083
in this build cancel button work and not longer click through
but other part of pop up window is can be click through and press at buttons under

makus82:
this done

PR #544, opened 2019-03-22
Prevent clickthrough on overlapping controls

### LINKED ISSUE #188: Buttons are funky (click comes through them to next one)
Basically if you try to click attack building in the latest build you'll encounter this most likely. You click button, and nothing happens. What actually happens is a "buttonclick" event happens both for the attack button and the "close" button (the cross). You can check by clicking the bottommost part of the attack button - it will work. 

Shouldn't ButtonClick event require mouse to go down on the button and then go up? Shouldn't it NOT work the way it is now?

FranciscoDA:
Fixes #188 and #541

### DIFF
--- a/data/forms/city/overlay.form
+++ b/data/forms/city/overlay.form
@@ -5,7 +5,7 @@
     <style minwidth="480" minheight="60">
       <position x="centre" y="bottom"/>
       <size width="480" height="60"/>
-      <graphic>
+      <graphic isclickable="true">
         <size width="480" height="60"/>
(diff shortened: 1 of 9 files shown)


makus82:
in https://github.com/Openapoc/OpenApoc/commit/85c57aadad7e2b88e4faa72dd3db26e2d5ef3dcd
nothing change in regard #541

FranciscoDA:
Whoops, sorry. I just force-pushed into my branch with a fix. Thanks for report

PR #561, opened 2019-04-27
Check an Event is a form event before trying to read the form union

JonnyH:
This could cause heap overflows and crashes.

### DIFF
--- a/forms/control.cpp
+++ b/forms/control.cpp
@@ -331,13 +331,13 @@ void Control::eventOccured(Event *e)
 				                     Vec2<int>{surface->size.x / 2, surface->size.y});
 			}
 		}
-	}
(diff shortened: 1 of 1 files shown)


JonnyH:
This was causing bad memory accesses every (at least) KeyboardEvent - as the size of the keyboard event is smaller than the offset of the form event's EventFlag.

Possibly causing a number of crashes throughout the game?

JonnyH:
@FranciscoDA I'm guessing this was just a typo - and this else-if block should have been inside the 	if (e->type() == EVENT_FORM_INTERACTION) block?

FranciscoDA:
@JonnyH yes, it must have been an oversight from my part

PR #577, opened 2019-05-01
dont inherit null tooltip fonts from listboxes

### LINKED ISSUE #565: Tooltip related crash in options menu
On macOS 10.14.4, click Options from the main menu. Mouse cursor movement becomes very choppy. Aftermoving around and scrolling randomly a bit, the game eventually crashes with:

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==42772==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000103c1e45a bp 0x7ffeec476d70 sp 0x7ffeec476640 T0)
==42772==The signal is caused by a READ memory access.
==42772==Hint: address points to the zero page.

    #0 0x103c1e459 in OpenApoc::Control::eventOccured(OpenApoc::Event*) control.cpp:307
    #1 0x103c1b61c in OpenApoc::Control::eventOccured(OpenApoc::Event*) control.cpp:105
    #2 0x103cae711 in OpenApoc::ListBox::eventOccured(OpenApoc::Event*) listbox.cpp:151
    #3 0x103c1b61c in OpenApoc::Control::eventOccured(OpenApoc::Event*) control.cpp:105
    #4 0x105891050 in OpenApoc::OptionsMenu::eventOccurred(OpenApoc::Event*) optionsmenu.cpp:70
    #5 0x10391c684 in OpenApoc::Framework::processEvents() framework.cpp:717
    #6 0x10391776f in OpenApoc::Framework::run(std::__1::shared_ptr<OpenApoc::Stage>) framework.cpp:573
    #7 0x10378b6af in main main.cpp:26
    #8 0x7fff5a0f63d4 in start (libdyld.dylib:x86_64+0x163d4)

==42772==Register values:
rax = 0x0000100000000000  rbx = 0x00007ffeec476ba0  rcx = 0xb62c70b084ab021f  rdx = 0x0000100000000000  
rdi = 0x0000000000000000  rsi = 0x000000010d3fd860  rbp = 0x00007ffeec476d70  rsp = 0x00007ffeec476640  
 r8 = 0x0000000000000000   r9 = 0x0000000000000002  r10 = 0x00007fff5a22cbe6  r11 = 0x0000000000000202  
r12 = 0x000000010d3fd880  r13 = 0x00007ffeec476890  r14 = 0x0000000000000000  r15 = 0x000061400005a750  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV control.cpp:307 in OpenApoc::Control::eventOccured(OpenApoc::Event*)
==42772==ABORTING
Abort trap: 6
```

Looks like the root cause is ToolTipFont being nullptr on control.cpp:307. I added a log statement and found that the tooltip text of the offending control is "Quit after this many frames - 0 = unlimited":
```
W 40431460431 virtual void OpenApoc::Control::eventOccured(OpenApoc::Event *): About to make bad access, name: Control, pos: {268,6}, ttt: Quit after this many frames - 0 = unlimited
```

Maybe the tool tip font isn't getting initialized for the options form that's auto generated from flags?

FranciscoDA:
Fixes #565

### DIFF
--- a/forms/listbox.cpp
+++ b/forms/listbox.cpp
@@ -285,7 +285,6 @@ void ListBox::clear()
 void ListBox::addItem(sp<Control> Item)
 {
 	Item->setParent(shared_from_this());
-	Item->ToolTipFont = this->ToolTipFont;
 	resolveLocation();
 	if (selected == nullptr)
 	{
@@ -306,7 +305,6 @@ void ListBox::replaceItem(sp<Control> Item)
 		{
 			Controls.erase(Controls.begin() + i);
 			Item->setParent(shared_from_this(), i);
-			Item->ToolTipFont = this->ToolTipFont;
 			resolveLocation();
 			if (oldItem == this->selected)
 			{
--- a/data/forms/options.form
+++ b/data/forms/options.form
@@ -23,6 +23,7 @@
         <position x="28" y="74"/>
         <size width="510" height="375"/>
         <item size="21" spacing="4"/>
+		<tooltip font="smallset" />
       </listbox>
       <graphicbutton id="LISTBOX_SCROLL_UP" scrollprev="LISTBOX_SCROLL">
         <tooltip text="Scroll Up" font="smallset"/>
--- a/game/ui/general/optionsmenu.cpp
+++ b/game/ui/general/optionsmenu.cpp
@@ -24,7 +24,7 @@ OptionsMenu::OptionsMenu() : Stage(), menuform(ui().getForm("options"))
 			listbox->addItem(mksp<TextButton>(section.first, ui().getFont("smalfont")));
 			for (auto &opt : section.second)
 			{
-				listbox->addItem(createOptionRow(opt));
+				this->createOptionRow(opt, listbox);
 			}
 		}
 	}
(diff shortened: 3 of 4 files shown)


forms/listbox.cpp
```diff
@@ -285,7 +285,10 @@ void ListBox::clear()
 void ListBox::addItem(sp<Control> Item)
 {
 	Item->setParent(shared_from_this());
-	Item->ToolTipFont = this->ToolTipFont;
+	if (!Item->ToolTipFont && this->ToolTipFont)
```
JonnyH: You could just simplify this to if (!Item->ToolTipFont), as if this->ToolTipFont is null then you're just setting null to null so no change

JonnyH: But long term - we should probably cope with null fonts - probably implying some "default"?

Or make the Control constructor set a default, so it'll never be null (unless someone explicitly sets it)

FranciscoDA: Actually I changed my mind: now I think it will be better if we avoid implicit inheritance for control attributes altogether. Sometimes listbox items are more complex than a single widget and inheriting attributes only to the top-level widget may be confusing and inheriting to the whole subtree may be too complex/expensive. I think it will be better to manually inherit each attribute in place where required.

I thought about the "default" font, but I don't recall that we have a fallback font defined in the framework? We don't want to tie the control library to a specific set of required resources either

JonnyH:
I think this isn't needed since PR #595 was merged

Issue #568, opened 2019-04-28
macOS: "Abandon and Restart game" --> "Start campaign game" sometimes crashes


ashenomo:
Here is one case that reliably reproduces, but I think there are others:

1. Run OpenApoc
1. Start new game in superhuman
1. Options --> "Abandon and Restart game" (goes back to main menu)
1. Start new game in superhuman
--> Crash

```
W 9615745276 void OpenApoc::City::initialSceneryLinkUp(): Begun scenery link up!
W 9699892947 void OpenApoc::City::initialSceneryLinkUp(): Begun scenery link up cycle!
W 9726545156 void OpenApoc::City::initialSceneryLinkUp(): Attempting link up of unlinked parts
W 9772007758 void OpenApoc::City::initialSceneryLinkUp(): Link up finished!
W 9773366767 void OpenApoc::GameState::validate(): Validating GameState
W 9783263046 void OpenApoc::GameState::validate(): Validated GameState
/Users/ice/PycharmProjects/openapoc/OpenApoc/forms/scrollbar.cpp:215:38: runtime error: division by zero
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/ice/PycharmProjects/openapoc/OpenApoc/forms/scrollbar.cpp:215:38 in 

// Step 4: Click "Superhuman"

AddressSanitizer:DEADLYSIGNAL
=================================================================
==45956==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7fff36d106c4 bp 0x70000f7766f0 sp 0x70000f7766f0 T4)
==45956==The signal is caused by a READ memory access.
==45956==Hint: address points to the zero page.
    #0 0x7fff36d106c3 in glDeleteTextures (libGL.dylib:x86_64+0x16c3)
    #1 0x10b81ed15 in (anonymous namespace)::GLPaletteImage::~GLPaletteImage() gl_2_0.hpp:1996
    #2 0x10b81eddf in (anonymous namespace)::GLPaletteImage::~GLPaletteImage() ogl_2_0_renderer.cpp:651
    #3 0x10b81f3f9 in std::__1::__shared_ptr_pointer<(anonymous namespace)::GLPaletteImage*, std::__1::default_delete<(anonymous namespace)::GLPaletteImage>, std::__1::allocator<(anonymous namespace)::GLPaletteImage> >::__on_zero_shared() memory:2335
[...]

Issue #542, opened 2019-03-21
[CRASH] when start alien invasion mission in building after center camera on it from Message Log (segmentation fault)


Atrosha:
When I select "alien spotted" to focus camera view on Megapol building and Valkyria with agents flied into it. After click on icon to start alien invasion mission the game crashes. Sometimes on loading screen.

Video with bug presentation: https://youtube.com/video/OivfcPrhfVE/

Guided by @FranciscoDA I tried to start with args
--Framework.Renderers=GL_2_0 
--Framework.Renderers=GLES_3_0
--Framework.Data.ImageSetCacheSize=10000

Compiled on Linux Mint 19.1 with g++ and clang++
Bug persist with different versions on linux nVIDIA drivers and opensource nouveau driver.

[save_Atrosha Strategy.save.gz](https://github.com/OpenApoc/OpenApoc/files/2993004/save_Atrosha.Strategy.save.gz)

[gdb_output.txt](https://github.com/OpenApoc/OpenApoc/files/2993007/gdb_output.txt)

Atrosha:
In result of our prospects with @FranciscoDA we found the origin that causes a crash:
![chat](https://user-images.githubusercontent.com/37053399/54881390-63d15180-4e60-11e9-8616-635860857d9d.png)

I attach full log with Gl debug info:
[debug_stderr_glinfo_output.txt](https://github.com/OpenApoc/OpenApoc/files/3000514/debug_stderr_glinfo_output.txt)

Atrosha:
For now to avoid the crash use
`--Game.ASyncLoading=0 --Framework.Data.ImageCacheSize=10000 --Framework.Data.ImageSetCacheSize=10000`
as game parameters.

It causes cache handler do not delete textures.

JonnyH:
The root cause of this issue is when a thread other than the "main" renderer thread tries to load an image, it may cause a cached image to be deleted - and if that was used in the renderer it will also destroy the renderer's private data in a thread where the GL context isn't bound, which is undefined (and indeed some drivers crash or abort). This also can cause crashes at exit for a similar reason - where the renderer (and GL context) are destroyed before the last Images are.

The fix for this would be have a threaded renderer, so all the render ops are called from the same thread as the context is bound - I have some prototypes of this hanging around, but never got around to merging - as it happens none of the drivers I happen to use crash - they just spit out a warning and never destroy the texture (effectively leaking)

Atrosha:
We can try together.

PR #593, opened 2019-06-09
Wip/lazy renderer image destruction

JonnyH:
This defers destruction of GL objects where the destructor is called from the non-renderer thread.

This should hopefully solve crashes with some drivers - e.g. issue #568 or #542 (and maybe a number of others)


Atrosha:
A lot of respect! Thanks.

makus82:
\o/ nice , ty Jonny =]

JonnyH:
Though this can still cause a crash-on-exit (due to the Images outliving the Renderer) - that issue was there previously anyway, so I'll merge this as a "we're getting better, if not perfect" fix as a crash on exit is a lot better than a possible crash anytime :)

PR #595, opened 2019-06-10
Set a default tooltip font

JonnyH:
If unset, use "smallset" (settable with Forms.TooltipFont option).

Fixes crashes due to null unset tooltip fonts when trying to show the tooltip


JonnyH:
Thought having a default if unset is less fragile than the proposal in #577 - but the 'code smell' of inheriting random things still exists

PR #602, opened 2019-06-13
Destroy the Data implemention before the renderer

JonnyH:
This should ensure that any possibly cached images are destroyed first,
as their destructors can call back into the Renderer to destroy any
private data.

Doing this should avoid a crash at exit due to use-after-free of the
renderer.


makus82:
oooh nice

Issue #569, opened 2019-04-28
macOS: Mouse click target off actual pointer in some screens


ashenomo:
**Agent / vehicle assignment screen**
[Screen Recording 2019-04-28 at 22.46.42.mov.zip](https://github.com/OpenApoc/OpenApoc/files/3125046/Screen.Recording.2019-04-28.at.22.46.42.mov.zip)

* The white outline should be following the mouse cursor, but it seems to be off by some scaling factor
* Clicking selects the white outlined agent *and* the agent at which the cursor is actually pointing.

**Base view**
[Screen Recording 2019-04-28 at 22.56.06.mov.zip](https://github.com/OpenApoc/OpenApoc/files/3125052/Screen.Recording.2019-04-28.at.22.56.06.mov.zip)

* Base facility highlight is off by a constant factor?
* The Yes/No buttons of the "Destroy facility?" confirmation seem to be highlighted when the cursor is at the off location, but clicking here does not activate the button.
* Basically impossible to play 😞

ashenomo:
Battle mode seems to work fine.

ashenomo:
Original report was for windowed 1280x720. I also tried in Fullscreen mode and windowed 640x480 mode. Issues persisted in both cases.

Issue #566, opened 2019-04-28
On macOS, moving the mouse stops intro movie


ashenomo:
I assume the expected behavior is to stop the intro movie when hitting Esc or maybe clicking the mouse, but not by just moving the mouse.

Hitting Esc also stops the movie.

JonnyH:
This seems to be due to MacOS sending finger down/up events from touchpads - and due to mobile we see "finger down" as a reason to skip the intro movie.

I don't really know how to fix this - I might just disable all touchscreen support, as it's unlikely that chasing mobile support would give a good experience without a /large/ UI and controls overhaul anyway...

PR #614, opened 2019-06-21
Add an option to ignore touch events

JonnyH:
On non-android platforms, touch events are now default ignored. Set the
Framework.EnableTouchEvents option to override.


JonnyH:
This should fix some weirdness on MacOS laptop devices - as SDL sends both a touch and mouse event the touchpad.
We interpreted any touch event as skipping the movies, so ignoring them should solve issue #566 
And we also interpreted any touch events as absolute onscreen coords, rather than relative to the mouse cursor, which could cause some weirdness as 'clicks' may not happen in the expected place (IE beneath the drawn cursor). May be related to issue #569 ?

PR #624, opened 2019-06-22
Fix build on c++17 due to removed std::random_shuffle

JonnyH:
Use std::shuffle with a URBG instead

### DIFF
--- a/framework/framework.cpp
+++ b/framework/framework.cpp
@@ -12,6 +12,7 @@
 #include "framework/stagestack.h"
 #include "framework/trace.h"
 #include "library/sp.h"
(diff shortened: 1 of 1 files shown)


framework/framework.cpp
```diff
@@ -324,14 +325,15 @@ class JukeBoxImpl : public JukeBox
 	std::vector<sp<MusicTrack>> trackList;
 	PlayMode mode;
 	PlayList list;
+	Xorshift128Plus<uint64_t> rng;
```
JonnyH: With no dynamic seed this will always play the same music order - is this ok?

PR #645, opened 2019-07-21
Fix UI agent stats bar boundaries in Agent Equipment Sheet

Atrosha:
Minor interface fix of non-critical visual problem.

Current interface have only only 88 pixels for stats bar width, but its set to 100 in code.
So stats bars with > 88 of width are out of boundaries.

So I simple changed ImageSize.x in AgentSheet::StatsBar to 88 from 100.


Atrosha:
ImageSize.x not maxValue

Atrosha:
Sorry for that. Was drunk. Thanks @JonnyH for help. Merged in #645

Issue #643, opened 2019-07-20
Assertion error when replacing item in hand on a specific agent


Quickmind01:
With the unit Alice Taylor, if I replace the toxi-gun with another item and click OK, the game crashes and gives the attached error. It only seems to be her so far, though I didn't test any non-hybrids as I thought it was a mind bender issue. I will update later with a clip from my stream, but the test ends at about 46 minutes into the stream.

![Assertion Error](https://user-images.githubusercontent.com/51022165/61578836-5a7c8c80-aac2-11e9-9097-0ffe0c8322be.jpg)
[openapoc_log.txt](https://github.com/OpenApoc/OpenApoc/files/3413602/openapoc_log.txt)

Quickmind01:
[save_Stream Test - New Battle.zip](https://github.com/OpenApoc/OpenApoc/files/3413626/save_Stream.Test.-.New.Battle.zip)
 Forgot to upload save.

Quickmind01:
I later tested in a different mission by changing Alice Taylor's equipment in the mission setup screen. The battle loaded fine, but when I selected her group (Team 2), the game crashed with the same error. So something seems to be up with this particular agent and NOT having a toxi-gun in her right hand.

makus82:
great work Q , great stream https://www.twitch.tv/quickmind

Quickmind01:
Here is a link to the video showing some testing I did. Essentially, this one agent is the only one with this issue. https://www.twitch.tv/videos/456100905

JonnyH:
This seems to be due to the accuracy bar drawing - it's 50 pixels wide, but doesn't clamp. The selected agent and gun happens to end up with a real accuracy of 52, so tries to write to beyond the bounds of the image when drawing that weapons' accuracy indicator.

PR #650, opened 2019-07-22
Clamp the accuracy indicator to the image width

JonnyH:
Without this, large accuracy values may go over the end of the accuracy bar
image, hitting a LogAssert


JonnyH:
This is a fix for issue #643 - it seems to fix it for me

I haven't verified that the high resulting accuracy (>50) is correct, but I guess the near-max stats of the agent and high accuracy of the mind bender device produce a pretty rare scenario, and likely not useful for the user to see any accuracy difference.

Issue #646, opened 2019-07-21
Battleship won't selfdestruct/May be preventing save


Quickmind01:
This may be related to #629 . After completing the mission, the UFO on the Cityscape didn't self destruct. I think this may be preventing me from saving on the next day as the game errors out before the save is complete.

[save_Stream Test.zip](https://github.com/OpenApoc/OpenApoc/files/3414613/save_Stream.Test.zip)
[openapoc_log.txt](https://github.com/OpenApoc/OpenApoc/files/3414614/openapoc_log.txt)

Quickmind01:
Quick update: I was able to get the error only hours after the mission. You don't have to wait until the next day. However, the game did let me save immediately after loading the save the first time.

Quickmind01:
Another update: I believe firmly now that the UFO not self destructing is preventing the game from completing a save. I was able to save the game originally because time had been paused and the code to self destruct had not been ran so the UFO was legitimately still there. However, since the UFO DIDN'T self destruct, something is going wrong in the code when the game tries to save.

W 12605447300 void __cdecl OpenApoc::GameState::validate(void): Validating GameState
W 12614002600 void __cdecl OpenApoc::GameState::validate(void): Validated GameState
W 105548142800 void __cdecl OpenApoc::GameState::validate(void): Validating GameState
W 105555531100 void __cdecl OpenApoc::GameState::validate(void): Validated GameState
E 154428341400 void __cdecl OpenApoc::serializeOut<enum OpenApoc::VehicleMission::MissionType>(class OpenApoc::SerializationNode *,const enum OpenApoc::VehicleMission::MissionType &,const enum OpenApoc::VehicleMission::MissionType &,const class std::map<enum OpenApoc::VehicleMission::MissionType,class OpenApoc::UString,struct std::less<enum OpenApoc::VehicleMission::MissionType>,class std::allocator<struct std::pair<enum OpenApoc::VehicleMission::MissionType const ,class OpenApoc::UString> > > &): Invalid enum value for enum OpenApoc::VehicleMission::MissionType: 6
  0x00007FF70EAB998A PHYSFS_writeSLE64+0x1cb26a
  0x00007FF70EB15DFD PHYSFS_writeSLE64+0x2276dd
  0x00007FF70EB0FD5D PHYSFS_writeSLE64+0x22163d
  0x00007FF70EAA5F6C PHYSFS_writeSLE64+0x1b784c
[...]

PR #651, opened 2019-07-22
Add VehicleMission::MissionType::RestartNextMission enum serialization

JonnyH:
This was missing from the serialization generation source, so any saves that
happened to contain missions of this type would fail


JonnyH:
This should fix the save issue in #646

PR #652, opened 2019-07-23
Fixup some unused parameter warnings

JonnyH:


JonnyH:
Going forward, I want to remove any "std::ignore = var;" patterns (as it's not guaranteed to work as it's a bit of a hack using an impementation detail of std::tuple) and unnamed parameters (as often the name is good documentation), and instead use [[maybe_unused]].

This puts a requirement on c++17 support, so relies on #653

JonnyH:
Superseded by #659

Issue #639, opened 2019-07-14
Inventory Space/Template Bug


Quickmind01:
Related to Issue #637 with this bug, the storage space goes to 999%+. However, when attempting to sell these toxi-gun A clips, they do not appear in the list of things to sell. This then leaves storage too full to buy, sell, or transfer anything, including aliens. Also, it appears craft can't drop aliens (and captured equipment) off at other bases before returning to their home base. This also prevents the transfer of things.

[save_Inventory Space-Template Bug.zip](https://github.com/OpenApoc/OpenApoc/files/3390089/save_Inventory.Space-Template.Bug.zip)

Quickmind01:
Update: In my stream today, I learned that if I duplicate more clips, it pushes the inventory number to even more ridiculous numbers. However, this somehow allows selling and transferring to work again, though the toxi-gun A, B, and C clips are then not able to be sold from that base.

JonnyH:
So I have found an issue when a template stores a weapon with more ammo than is available when applying it - it doesn't check or clamp the current value, causing an underflow in the inventory (which as it's unsigned causes it to go a /really/ high value, some 4 billion on a 32bit integer).

JonnyH:
Think this was fixed by #654

PR #654, opened 2019-07-23
Cleanup agent equipment screen code

JonnyH:
Previously, they were re-processed every time there was a key event on each of
the numbers, so each of {up, down, press}, causing it to be processed
multiple times needlessly


JonnyH:
Note that this causes some asserts to get hit, so it seems that an underflow of ammo is possible - likely the cause of #639 and similar

Issue #609, opened 2019-06-15
Toxi-Gun A and B (and potentially more items) aren't going to inventory


Quickmind01:
I have manufactured dozens of Toxi-gun A and B clips, but only a few show up in inventory, even though the engineering screen shows them being produced normally. This is happening at two different bases with both sized engineering bays. I haven't tested, but this could potentially be happening with other items. I attempted two orders of 1 B-clip. Once both orders finished, my quantity of B-clips did not go up.
[save_Stream Test.zip](https://github.com/OpenApoc/OpenApoc/files/3293384/save_Stream.Test.zip)

JonnyH:
Noticed this while debugging #639 - it seems as though the agent equipment screen expects the 'inventory count' to be individual rounds - not clips - but the manufacturing system is only incrementing the inventory count by 1 each unit. So it's effectively producing a single round at a time, instead of a single clip, so you'd need to manufacture a number of units equal to the clip capacity before the agent equipscreen number will increment.

This is clearly wrong :)

Quickmind01:
Holy crap! That makes so much sense!

PR #655, opened 2019-07-24
Manufacture ammunition a whole clip at a time

JonnyH:
Previously only a single unit was created, which is intepreted as a single
round, so would take multiple manufactured units to be considered a complete
clip


JonnyH:
Should fix #609

PR #658, opened 2019-07-26
Use c++17 filesystem

JonnyH:
Now we require c++17, we no longer need to rely on boost::filesystem or
the experimental pre-standardization version


JonnyH:
It looks like we have to keep a boost::Filesystem fallback for now - as some toolchains (libcxx based?) seem to 'support' c++17 but lack the <Filesystem> implementation on some systems.

At least that's the case for MacOS' xcode 10.3

Issue #586, opened 2019-05-18
game crashes for no reason


Erydian:
Hi
as in title, game crashes for no reason at 2:55.
Add save and log
[Desktop.zip](https://github.com/OpenApoc/OpenApoc/files/3193699/Desktop.zip)

makus82:
ty for report @Erydian but "No reason" does not describe what you did before the crash )

Atrosha:
Only one strange event are occurs in this savegame next after loading: civic flying transport vehicle crash, as you see at my video:

https://youtu.be/DgPUgPbPpfs

But all resolves normally.

BeornTB:
I'd like to jump in the bandwagon of the "No reason" that describes nothing.

All I can offer is the log:
[openapoc_log.txt](https://github.com/OpenApoc/OpenApoc/files/3237741/openapoc_log.txt)

BeornTB:
Another crash, another log:
[openapoc_log.txt](https://github.com/OpenApoc/OpenApoc/files/3237780/openapoc_log.txt)

JonnyH:
I have, as yet, not managed to reproduce this crash.

Perhaps it's related to the GL texture destruction crash on macos (While we're /technically/ doing something wrong here by calling things from the wrong thread, on every non-macos driver I can test it just results in a small leak instead of a crash).

Does setting the "Framework.Data.ImageSetCacheSize" option to a high value (e.g. 9999999999) so images are never destroyed avoid this crash?

Atrosha:
Sorry for my wrong report. At second iteration of this issue testing I've got a crash at 2:55 on latest build.
I attach the gdb backtrace.
[Uploading 586_gdb_output.txt…]()

JonnyH:
Somehow this save has a civilian biochemist named "Jennifer Williams" with a homeBuilding "BUILDING_TEMPLE_OF_THE_VISITORS" with a modified_stats health of "0" and a current_stats health of "36", so Agent::updateHourly is trying to heal them.

As they're in their "home building", it tries to query how many medical bays there are in the base.. but naturally there is no base, as it's a civilian in a civilian building :)

PR #663, opened 2019-07-26
Fix crash when healing agents not in a building with base

JonnyH:
Non-xcom agents may have a 'home building' without a base, so check
the existance of a base before deref-ing to see how many medical bays
there are.


JonnyH:
Fixes #586

Issue #617, opened 2019-06-22
Bugged savegames Overwrite


Atrosha:
When I try to save game while selecting already present slot, the keyboard cursor is started blink from the begin of string with savegame's name and enteredg text overlay existing. After press ENTER - nothing happens. And OVERWRITE "yes/no" message dialog doesn't shows.

FilmBoy84:
What build are you running?
I cannot recreate this on OpenApoc-X64-v0.1-475-g16a9c33c

FilmBoy84:
Ah no, ignore that, it appears to apply to old save games only
Hmmmm will look into this
Adding the "help wanted" tag as i do not fully understand what is going on

FilmBoy84:
Not sure what happened in this, but broken in a recent update
So assigning @JonnyH as he knows more than me what's changed :D

JonnyH:
Just reproduced this - it seems like some Control activation issue - clicking on the name of the saved game shows the problem above, but clicking elsewhere on the button works fine.

JonnyH:
So the issue here is that the saved game name control is a Label, with an invisible TextEdit over the top that is only made visible when that row's button is pressed (with SaveMenu::beginEditing()) and filled with the expected save name contents to be modified.

But in this case that 'invisible' TextEdit is still handling events, so when it gets clicked it handles the event instead of passing it down to the parent button, so it starts editing it's (empty) contents instead of calling SaveMenu::beginEditing().

PR #664, opened 2019-07-26
Always hide SaveMenu TextEdit control at screen open

JonnyH:
Even if we're saving, we hide the TextEdit until a save slot has been selected,
as until then the names are displayed by a Label and the TextEdit not intended
to be active.


JonnyH:
Fixes #617

Issue #667, opened 2019-07-27
UFO is outside of the map\world border


Quickmind01:
When loading a different save file, one of the four UFOs on the map is way out of bounds. I don't believe this occurred during the original play though but only occurred after loading the save. The game will immediately crash when unpaused unless the debug command to crash all craft is used. However, I believe the game will still crash later due to the UFO still being out of bounds.

**Here are the last few lines when the game crashes without crashing the UFOs:**

W 28314459700 void __cdecl OpenApoc::TileObject::setPosition(struct glm::vec<3,float,0>): Clamped object to {0,0,0}
W 28319742700 void __cdecl OpenApoc::TileObject::setPosition(struct glm::vec<3,float,0>): Trying to place object at {-9001,-9000.27,0} in map of size {140,140,13}
W 28326447400 void __cdecl OpenApoc::TileObject::setPosition(struct glm::vec<3,float,0>): Clamped object to {0,0,0}
W 28350414200 void __cdecl OpenApoc::TileObject::setPosition(struct glm::vec<3,float,0>): Trying to place object at {-9001,-9000.19,-9001} in map of size {140,140,13}
W 28358854300 void __cdecl OpenApoc::TileObject::setPosition(struct glm::vec<3,float,0>): Clamped object to {0,0,0}
W 28368126200 void __cdecl OpenApoc::TileObject::setPosition(struct glm::vec<3,float,0>): Trying to place object at {-9001,-9000.19,0} in map of size {140,140,13}
W 28381914700 void __cdecl OpenApoc::TileObject::setPosition(struct glm::vec<3,float,0>): Clamped object to {0,0,0}
E 28399855200 void __cdecl OpenApoc::Vehicle::leaveDimensionGate(class OpenApoc::GameState &): Trying to launch already-launched vehicle
  0x00007FF60C7D72C8 PHYSFS_writeSLE64+0x100e48
  0x00007FF60C7CB405 PHYSFS_writeSLE64+0xf4f85
  0x00007FF60C7D7815 PHYSFS_writeSLE64+0x101395
  0x00007FF60C7DC75A PHYSFS_writeSLE64+0x1062da
  0x00007FF60C6F173A PHYSFS_writeSLE64+0x1b2ba
  0x00007FF60C5EBB40 PHYSFS_writeSLE64+0xfffffffffff156c0
  0x00007FF60C515857 PHYSFS_writeSLE64+0xffffffffffe3f3d7
[...]

PR #669, opened 2019-07-28
Don't create tileObject of vehicles between dimensions at game load

JonnyH:
Vehicles currently in a dimension gate are currently assigned a tileObject in
GameState::initGame(), despite them not having valid position coords, so appear
on the map at a nonsense location (0,0,0) and hit a LogError() in
Vehicle::leaveDimensionGate() as it doesn't expect an existing tileObject.

So instead store a flag for if a Vehicle is between dimensions - and don't
create TileObjects for Vehicles with that flag set at game load time.

Note: This doesn't fix existing saves where there are vehicles currently in the
gates, as the flag won't be set correctly. New games, or games loaded from a
state where there aren't any Vehicles that happen to be in the gates at that
moment, should be fine going forward.

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -1271,6 +1271,7 @@ void Vehicle::leaveDimensionGate(GameState &state)
 	auto initialFacing = 0.0f;
 
 	LogInfo("Leaving dimension gate %s", this->name);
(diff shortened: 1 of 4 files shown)


JonnyH:
Should fix #667

Issue #649, opened 2019-07-21
Request: More UFOs in Debug Mode Crash Feature


Quickmind01:
Would it be possible to add more UFOs to the debug feature that drops three crashed UFOs? Specifically, could the Assault UFO be added? Perhaps a list could be chosen from randomly? Right now, it looks like only three specific UFOs are dropped.

PR #670, opened 2019-07-28
Randomize which UFOs are generated when generating crashed UFOs

JonnyH:
Should allow easier progress is debugging issues with things gated behind
specific UFOs


JonnyH:
As mentioned in #649

Issue #459, opened 2019-01-03
Alien Assault Ship cannot be targetted (or even hit)


FilmBoy84:
For some reason, the assault UFO cannot be targetted or hit by vehicles in the game
Shots go right through without damaging
The UFO cannot be selected or targeted
![image](https://user-images.githubusercontent.com/33913088/50634178-abbd9800-0f45-11e9-8786-f4a0a5aed690.png)

FilmBoy84:
[save_SuperHuman 2.zip](https://github.com/OpenApoc/OpenApoc/files/2760353/save_SuperHuman.2.zip)
Savegame attached, the alien assault ship cannot be targetted by any vehicle
(It was artificially crashed with debug command, but could not be targetted before this either)

FilmBoy84:
Another instance of this problem, mid battle...
Interestingly the Griffon AFV does try to hit the assault ship, but cannot do so
[save_SuperHuman 1B.zip](https://github.com/OpenApoc/OpenApoc/files/2795893/save_SuperHuman.1B.zip)

Quickmind01:
I found this today as well. My craft also didn't seem to hit it when shooting at it (after being left stationary).
[save_Cant Target Assault UFO.zip](https://github.com/OpenApoc/OpenApoc/files/3269890/save_Cant.Target.Assault.UFO.zip)

Quickmind01:
To add more: I crashed two of these with the debug command, but they still cannot be targeted and therefore a mission into them still cannot be done. This prevents further research for Retaliator and Annihilator.

makus82:
this bug serious coz ufo appears in the game quite early and often. And this UFO is not even possible to destroy using debug Hotkeys !!!!

JonnyH:
So it looks like the VehicleType voxelMapsLOS and voxelMaps generated by the dataextractor are completely empty....

JonnyH:
Yeah, the extraction logic looks wrong:

In InitialGameStateExtractor::extractVehicles():

The v.loftemps_height of the assault ship is 16 - so as it's a 1x1x1 unit then the loftemps should completely fill the height of the unit (probably expected, as it appears visually "tall" while other UFOs tend to be thinner but wider).

That then means that the "freeSpace" variable is 0.

So "start" is set to (0+1)/2, truncated to 0.
and "end" is ((v.size_z * 16) - (freespace / 2 )) % 16
=((1 * 16) - (0 / 2)) % 16
[...]

PR #671, opened 2019-07-29
Fix LOFtemps for vehicles with exactly 16-high LOFTemp stacks

JonnyH:
This is what was causing the Alien Assault ship to be unhittable/unselectable

Note this data is stored in the saved game state, so new games will be required


JonnyH:
Should fix #459 

Note that this is a "minimal" targeted fix - I have some reservations about the LOFTemps extraction code being overly complex, and some things about the about may not be correct (like the clipping of the >1 height ships (alien mothership/battleship) doesn't seem right?

makus82:
great work! will test it

PR #706, opened 2019-08-20
Small serialization speedups

JonnyH:


JonnyH:
The biggest win here for me seems to be the vector<up<XMLSerializationNode>> replacement, giving ~25% speedup in test_serialize. The UString change improves things by ~5-10%, and the final class is near the noise (gcc/clang both seem to be able to figure that there's no other subclasses anyway)

Issue #709, opened 2019-08-30
Deadlock in BattleUnitMission causing infinite loop


JonnyH:
Seen in QuickMind's 2019/08/30 stream at 6:30:15

PR #710, opened 2019-08-30
Snooze for a tick when restarting a GoTo mission that's blocked

JonnyH:
I think this may be a possible cause of a deadlock - if a unit can't route any
closer to it's intended destination, it'll loop in Restart->Goto->Restart
forever.

Adding a 1 tick snooze should allow the game to progress, and possibly allow the
blockage to clear itself if it's temporary (e.g. another unit)


JonnyH:
I /think/ this may be what's happening in #709

Issue #711, opened 2019-08-30
Use-after-free in Teleport mission causing possible crash


JonnyH:
When starting a teleport mission, it clears all that unit's missions, causing them to be deleted - including the current TeleportMission, so any accesses into that mission become invalid.

Noticed using the debug 'middle click' teleport mode in a skirmish battle, but looks like it affects all TeleportMissions.

Output from Address Sanitizer:

W void OpenApoc::BattleView::orderTeleport(Vec3<int>, bool) Using teleporter cheat!                                                    W void OpenApoc::BattleUnitMission::start(OpenApoc::GameState &, OpenApoc::BattleUnit &) Unit BATTLEUNIT_10 mission "Teleport to  {37,5
2,1}" starting                                                                                                                         =================================================================
==37381==ERROR: AddressSanitizer: heap-use-after-free on address 0x61000022ef44 at pc 0x0001031f042e bp 0x7ffeecf59bf0 sp 0x7ffeecf59be
8                                                                                                                                      READ of size 1 at 0x61000022ef44 thread T0
    #0 0x1031f042d in OpenApoc::BattleUnit::addMission(OpenApoc::GameState&, OpenApoc::BattleUnitMission*, bool) battleunit.cpp:5835
    #1 0x1031f20f4 in OpenApoc::BattleUnit::setMission(OpenApoc::GameState&, OpenApoc::BattleUnitMission*) battleunit.cpp:5767             #2 0x10416699f in OpenApoc::BattleView::orderTeleport(glm::vec<3, int, (glm::qualifier)0>, bool) battleview.cpp:2708
    #3 0x104172256 in OpenApoc::BattleView::handleMouseDown(OpenApoc::Event*) battleview.cpp:3823                                          #4 0x104168ed4 in OpenApoc::BattleView::eventOccurred(OpenApoc::Event*) battleview.cpp:2887
    #5 0x102d80ad3 in OpenApoc::Framework::processEvents() framework.cpp:807                                                               #6 0x102d7e5bf in OpenApoc::Framework::run(std::__1::shared_ptr<OpenApoc::Stage>) framework.cpp:655
    #7 0x102ca70e9 in main main.cpp:26                                                                                                     #8 0x7fff6000a3d4 in start (libdyld.dylib:x86_64+0x163d4)
                                                                                                                                       0x61000022ef44 is located 4 bytes inside of 192-byte region [0x61000022ef40,0x61000022f000)
[...]

PR #712, opened 2019-08-30
Don't remove the current TeleportMission in TeleportMission's start()

JonnyH:
Deleting this current mission could cause crashes due to use-after-free errors
on any subsequent mission data access


JonnyH:
Should fix #711

Issue #690, opened 2019-08-13
[Renderer] Cityscape Projectile Sprites are not longer shown


FilmBoy84:
For some reason, missiles, beams and other projectile sprites in the cityscape no longer display
The sound is made, the damages applied, but no visual tracer exists user-side

![image](https://user-images.githubusercontent.com/33913088/62935551-5856d680-bdbf-11e9-8e75-89ccea34f8d5.png)

JonnyH:
Is this every weapon? Every game? Or is there some variance to it?

I can't seem to reproduce this locally. What platform are you on? What GPU and drivers?

Can you get an apitrace of a quick run where a shot should be visible? (The shorter the better, as the file sizes can get pretty mad :)

JonnyH:
Actually, I have an idea: Does rebuilding the data help? (IE remove the extractor .exes and rebuild everything)

I think there's a race in how cmake may parallelise the data extraction - if the vehicle equipment extractor runs before the bullet sprites are created, it'll fail to open them and write empty sprite references into the generated vehicle equipment files.

By force-rebuilding the data while there's an already-extracted set of data, we know that there will be valid bullet sprites available from the start (IE the last extraction run)

FilmBoy84:
Just aside to this issue

All the bullet sprite entries for cityscape weapons are empty is this intentional or not?

![image](https://user-images.githubusercontent.com/33913088/63211262-d7fce200-c0ec-11e9-8417-b325c811e2de.png)

JonnyH:
That's caused by a race, where if the generator for the sprites themselves isn't finished before the equipment extractor runs it tries to load the "missing" sprites, fails, so writes nothing - causing the empty sprite lists you see.

I might just remove the multi-process extraction - it doesn't really save that much time, and causes race issues like this...

PR #715, opened 2019-08-31
Don't parallelize data extraction in build

JonnyH:
Some extraction targets depend on the output of others, and currently the
parallelization doesn't correctly handle these dependencies, so races may occur
causing incomplete data generation.


JonnyH:
Fixes #690

Issue #575, opened 2019-05-01
Occasional use-after-free in updateFalling (both city and battle)


ashenomo:
I've repro'd this a couple times in city view by Ctrl+Alt+Shift+Rightclicking a whole bunch of buildings in quick succession, and once in battle when a lot of explosives were going off at the same time.

In both cases the failure is around a for loop that's iterating through `tileObject->getOwningTile()->ownedObjects)`. Maybe things are getting removed from the container during iteration?

City scape failure:

```
W 100903857833 void OpenApoc::Scenery::collapse(OpenApoc::GameState &): Scenery at {50,25,5} type CITYTILE_CITYMAP_204 now falling
W 100903894479 void OpenApoc::Scenery::collapse(OpenApoc::GameState &): Scenery at {50,25,6} type CITYTILE_CITYMAP_206 now falling
=================================================================
==57863==ERROR: AddressSanitizer: heap-use-after-free on address 0x60400353a298 at pc 0x00010be68ad4 bp 0x7ffee48ee1c0 sp 0x7ffee48ee1b8
READ of size 8 at 0x60400353a298 thread T0
    #0 0x10be68ad3 in std::__1::__tree_end_node<std::__1::__tree_node_base<void*>*>* std::__1::__tree_next_iter<std::__1::__tree_end_node<std::__1::__tree_node_base<void*>*>*, std::__1::__tree_node_base<void*>*>(std::__1::__tree_node_base<void*>*) __tree:176
    #1 0x10be25b2f in std::__1::__tree_const_iterator<std::__1::shared_ptr<OpenApoc::TileObject>, std::__1::__tree_node<std::__1::shared_ptr<OpenApoc::TileObject>, void*>*, long>::operator++() __tree:920
    #2 0x10be59e97 in OpenApoc::Scenery::updateFalling(OpenApoc::GameState&, unsigned int) scenery.cpp:1508
    #3 0x10bdb244b in OpenApoc::City::update(OpenApoc::GameState&, unsigned int) city.cpp:289
    #4 0x10c254883 in OpenApoc::GameState::update(unsigned int) gamestate.cpp:972
    #5 0x10d8d6628 in OpenApoc::CityView::update() cityview.cpp:1739
    #6 0x10b499ff1 in OpenApoc::Framework::run(std::__1::shared_ptr<OpenApoc::Stage>) framework.cpp:581
    #7 0x10b30dd1f in main main.cpp:26
[...]

PR #716, opened 2019-08-31
Handle invalidating Tile::ownedObject iterators in Scenery::updateFalling

JonnyH:
This should avoid a crash caused by Vehicle::applyDamage() possibly destroying
a vehicle, invaldating the current TileObject iterator on Tile::ownedObjects.


JonnyH:
I think this should fix #575

Issue #719, opened 2019-09-01
Transfer Crash


Quickmind01:
When attempting to transfer 6 personal shields, 5 quantum physicists, a dead psimorph, and a dead micronoid aggregate from Base 1 to Base 2, the game crashes after clicking the OK button. Attached is the log file and the save file. This is replicatable.

This is using the Weapon Mod.
[Transfer Crash.zip](https://github.com/OpenApoc/OpenApoc/files/3563780/Transfer.Crash.zip)

PR #720, opened 2019-09-02
Don't try to remove unassigned scientists from lab when transferring

JonnyH:
This was causing a crash as agent->lab_assigned will be nullptr on
agents not assigned to a lab.


JonnyH:
Should fix #719

PR #704, opened 2019-08-20
Add a launcher

JonnyH:
CURRENTLY A WORK IN PROGRESS - NOT COMPLETE

Will give an 'easy' place to set some config options before running the game

Done:
* Allows you to select the resolution (including 'custom' entry)
* Allows toggling fullscreen
* Allows to set data/cd paths in a nicer way (IE 'find file' dialogs)

Todo:
* Test on windows
* Finish Mods directory
* Tweak layout to make it less awful
* Add logo/image?

PR #721, opened 2019-09-03
Add a launcher to set some options

JonnyH:
Some options are hard to setup in-game (IE things that pretty much require a renderer restart, like screen size/fullscreen options), and gives a nicer UI to select the CD / Data paths.

Going forward I intend to make the 'mods' tab actually do something :)


JonnyH:
This is a cleaned up version of #704 - I hope to merge this Soon (tm)

PR #731, opened 2019-09-05
Initial WAV sample loading

JonnyH:
Only supports u8/s16_lsb mono WAV files


JonnyH:
Use in XML with the string:
"WAV:path/to/file.wav"

Hopefully pretty simple :)

Note that unsupported file (more than 1 channel, not u8/s16_lsb) will be rejected. And currently we create a 22050hz output device, so more than that is just more work right now.

Issue #718, opened 2019-09-01
UFO Mission Item fell off the end of the world


Quickmind01:
Attached is my save and log for a UFO mission (Assault Ship?) where I was shooting aliens coming out of the UFO when this crash occurred. This likely occurred at the UFO entrance.

I am using the Weapons Mod V5 and the latest build.

[Dropped of end of the world.zip](https://github.com/OpenApoc/OpenApoc/files/3563622/Dropped.of.end.of.the.world.zip)

FilmBoy84:
REFERENCING ISSUES #587 #581 and #718 IN THIS COMMENT AS THEY APPEAR TO ALL BE RELATED
PLEASE CLOSE ALL WHEN RESOLVED

FilmBoy84:
SEE ALSO #393 and #693 

Here's another trajectory, end of the world and Incorrect Tile Coordinates loop

![image](https://user-images.githubusercontent.com/33913088/64485829-25b4c800-d21d-11e9-8f73-04b2c19d07a0.png)

![image](https://user-images.githubusercontent.com/33913088/64485835-2f3e3000-d21d-11e9-931d-d58880577701.png)

![image](https://user-images.githubusercontent.com/33913088/64485843-36653e00-d21d-11e9-808c-ecee4fac0162.png)

![image](https://user-images.githubusercontent.com/33913088/64485847-3c5b1f00-d21d-11e9-925f-b3f0f6066582.png)

!!!AND A VIDEO TOO!!!
https://youtu.be/gjESNoUMAE8

FilmBoy84:
It seems that one of the triggers for this is units flying whilst being targetted

JonnyH:
@FilmBoy84 That nonsense values (e.g. -MAX_INT tile coords) looks like a different issue to the "fell out of world" issues.

makus82:
![screenshot007](https://user-images.githubusercontent.com/11581207/64495038-5c322780-d29d-11e9-86e7-00c45fc48bde.png)
load with mod and some agent was with cultist skin)

Issue #581, opened 2019-05-11
bug "unit fell off the end of world"


Erydian:
Hi
while firing weapons with exploding ammunition, the floor is falling and some creatures are falling and game crashes.
i am adding a log file, if it will help ? 
PC ver. OpenApoc-v0.1-451-gcf7fc374
[Uploading openapoc_log.txt…]()

![error](https://user-images.githubusercontent.com/50518240/57569189-a68bc080-73f1-11e9-8f73-d3230bf39413.png)

Quickmind01:
I just had the same issue with skeletoids dying way up in the air. This is repeatable with the save file. Just run the game.
[openapoc_log.txt](https://github.com/OpenApoc/OpenApoc/files/3413654/openapoc_log.txt)

[save_Stream Test - New Battle.zip](https://github.com/OpenApoc/OpenApoc/files/3413650/save_Stream.Test.-.New.Battle.zip)

FilmBoy84:
REFERENCING ISSUES #587 #581 and #718 IN THIS COMMENT AS THEY APPEAR TO ALL BE RELATED
PLEASE CLOSE ALL WHEN RESOLVED

FilmBoy84:
From QMs stream

**_Possible issue with jumping (forced by [J] or via UnitAI); causes a unit to fall off edge of world - this may be the cause of some of our other falling off errors if aliens, security and civvies are attempted a jump_**

JonnyH:
Do you have a a repro case on video? Can you link the time? Just jumping near the map edge doesn't seem to hit this for me, I can't jump "off" the map?

I suspect this is more likely due to somehow the lowest z=0 floor tile getting destroyed, despite it /should/ be set as immortal, so you literally fall out the bottom of the map.

JonnyH:
I have a theory for at least some of the 'item falling off the world' errors:

Z=0 ground tiles are considered immortal, but will have the "damaged" flag set as normal (IE if they have suffered enough damage and have a valid damaged_map_part link).

But damaged tiles are not considered valid for support in BattleMapPart::findSupport(), so anything attempting to use this ground object as support will actually fall through it, into the magical void underneath the map (and hitting the "if z < 0, print the "unit fell off the end of world" error" issue seen here).

And due to the way that z=0 immortal tiles are handled, when they would be damaged or destroyed they still call the functions to check support, despite the tile not being removed in die(), so anything on top of this tile (items, other battle map objects) would instantly start falling through the tile.

[...]

PR #747, opened 2019-09-08
Damaged floor tile should still provide support for scenery

JonnyH:


JonnyH:
A tentative fix for #718 and #581 (and maybe others)

My theory is that findSupport() isn't treating damaged floor tiles as valid to support scenery, but the z=0 "immortal" tiles still get damaged, so things can "fall through" them off the bottom of the map (which causes the "fell out of world" error messages).

I haven't yet been able to find a good repro case to really test this with....

PR #758, opened 2019-09-13
Add burst mode for weapons and fix Stormdog ammo with it

JonnyH:
Add burst mode weapons and fix the stormdog ammunition as part of that

### DIFF
--- a/game/state/gamestate.cpp
+++ b/game/state/gamestate.cpp
@@ -654,7 +654,7 @@ void GameState::fillPlayerStartingProperty()
 	    auto v = current_city->placeVehicle(*this, {this, type}, this->getPlayer(), {this, bld});
 	    v->homeBuilding = v->currentBuilding;
 	}*/
-	
+
 	for (auto &pair : this->initial_vehicles)
 	{
 		auto v = current_city->createVehicle(*this, pair.first, this->getPlayer(), {this, bld});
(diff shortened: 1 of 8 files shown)


JonnyH:
Untested rebase of #758

game/state/gamestate.cpp
```diff
@@ -654,31 +654,37 @@ void GameState::fillPlayerStartingProperty()
...
 	}*/
+
 	for (auto &pair : this->initial_vehicles)
 	{
 		auto v = current_city->createVehicle(*this, pair.first, this->getPlayer(), {this, bld});
 		v->homeBuilding = v->currentBuilding;
 		for (auto &eq : pair.second)
 		{
-			v->addEquipment(*this, eq);
+			auto device = v->addEquipment(*this, eq);
```
JonnyH: This is setting all the initial vehicle ammo to full? I've noticed that before, good to have it fixed - but it should probably be a different change?

game/state/gamestate.cpp
```diff
@@ -654,31 +654,37 @@ void GameState::fillPlayerStartingProperty()
...
+			device->ammo = eq->max_ammo;
 		}
 	}
+
 	// Give the player initial vehicle equipment
 	for (auto &pair : this->initial_vehicle_equipment)
 	{
 		base->inventoryVehicleEquipment[pair.first.id] = pair.second;
 	}
+
```
JonnyH: These extra lines aren't really needed, and if you really think they make it clearer it should be a separate change anyway

game/state/gamestate.cpp
FilmBoy84: Yes, it's been a bug in OA for a while
Wasn't sure if to separate or not as it relates to the weapon this PR is changing for burst mode
If you want it separate i can do that?
I have to change the burst from 10 to 3 anyway

game/state/gamestate.cpp
FilmBoy84: Sorry, which two lines? The two blank ones in the section above or all the changes to gamestate.cpp

If just the blank ones, then agreed, they are not needed im happy for them to be removed

JonnyH: The blank ones - separating out logical changes into different commits (or even PRs) often makes things easier going forward - when doing stuff like merges or rebases like this, for example :)

So each "single change", be it bug fix, feature, or formatting change, ideally should each be in different changes.

FilmBoy84:
Referencing this in PR #790 which carries an updated version of this PR
Once 790 is merged, this PR can also be closed without merge

Issue #767, opened 2019-09-16
Ammo and Clip Quantity Bug


Quickmind01:
If I recall correctly, if the ammo is sold in this save, you will end up with the millions of clips bug again. The Heavy Plasma Clip is a good example. I had sold it as a partial clip, and when I rebought it, it was still a partial clip. Had I had a second full clip, it would have likely refilled. Also, I believe this clip bug affected my storage space.

This is using OpenApoc-v0.2.0.2-16-g2f96387e and V7.1 EWM.

[save_Ammo and Clip Bug.zip](https://github.com/OpenApoc/OpenApoc/files/3614484/save_Ammo.and.Clip.Bug.zip)

JonnyH:
It seems like the save already has a nonsense number of ammo clips, no need to buy/sell anything.

But this gives a good clue - it looks like the buy/sell code isn't checking if an ammo clip is partially full - it just takes the (MAX_AMMO) of one clip out of the stores.

But that naturally fails if you try to sell a non-full clip, as it takes MAX_AMMO off a value less than MAX_AMMO, causing an underflow (as it's an unsigned value, instead of going negative it ends up with UINT32_MAX - however many it would be below zero, so over 4 billion). Which neatly explains the issue.

So this raises the question of how we should handle partially filled clips - us trying to be "clever" and add features onto the OG by tracking the number of bullets not clips outside battles has a number of knock on effects that I'm not sure it's "worth" the extra complexity for a feature most people won't even notice:

- Should buying/selling a partially filled clip still have the same cost as a full one?
- The buy/sell UI really has no way of informing the user that a clip is partially filled, either buying or selling
- There's already no way in-battle to merge partially filled clips, so if you're not careful you might end up in the situation where you fight multiple battles in a row and have multiple near-empty clips on each agent, which just seems frustrating

I might go through and slash and burn this feature and just auto-merge partially filled clips at the end of each battle. I'm not sure if this management really /adds/ anything to the game.

JonnyH:
So looking further into the code, it seems like the market doesn't actually support partial clips, after removing the number of rounds from the stores, the economy just tracks the number of clips.

That means you always buy full clips, even if you sold it partially filled - so a simple clamp here will magically refill a clip if you sell it and buy it back.

But as the selling will underflow to a CRAZY large number, so the buying it back will then overflow that back down to a sane number - back to where you started.

Paddywhacker:
[...]

PR #771, opened 2019-09-16
Don't underflow inventory selling partially-filled ammo clips

JonnyH:
Clamp against the current round inventory count, as otherwise selling your last
partially filled clip will underflow the 'unsigned' inventory cound and you get
a silly value.

Note that as the market only tracks whole clips, this cause selling and rebuying
a partially filled clip to refill it for free, but that seems less worse than
giving the player ~4 billion rounds.


JonnyH:
Should fix #767

PR #783, opened 2019-10-11
Add an "onmodload" script to be run at mod load, and use that to implement difficulty

JonnyH:
This exposes a new "appendGameState(GameState, path)" to lua, which can be used to load a datadir-relative archive into the supplied gameState.

It also adds a new <modloadscript> tag to the modinfo.xml, which contains a (dataDir-relative) path to a lua script to be run once that mod has finished loading (but before any subsequent mods in the load order). Then it uses this to implement the difficulty selection, rather than the current "hacky" way of always loading it after all other mods (which caused issues if mods wanted to alter values contained with - it would just override them as it was always at the very end of the mod load order). Now this script is run (and difficulty GameState loaded) immediately after the 'base' mod in the load order.

Note that the datadir paths above may be overridden by other mods - so it's a good idea to put them in a known unique path (e.g. using the mod ID, the two base examples being put in "data/DIR/org.openapoc.base/FILE" - that still gives other mods the ability to override them if necessary, but avoid accident collision.


FilmBoy84:
Just an observation with this and no mods (except base) it loads each difficulty as expected

If i used the EWM below, it uses only the base NOVICE settings for difficulty (confirmed by the 140000 starting money)

It does not appear to read the EWM difficulties, it only reads base difficulty at the novice level, regardless of which difficulty option the user selects

[OpenApoc Extended Weapons Mod - V9 PREVIEW.zip](https://github.com/OpenApoc/OpenApoc/files/3720557/OpenApoc.Extended.Weapons.Mod.-.V9.PREVIEW.zip)

FilmBoy84:
I'm just adding the LUA script to modinfo to see if this fixes it... But it is still of note that it locks all difficulties to base novice level if there is a problem...

FilmBoy84:
Hmmmm added the modinfo and other required directories, now getting cities.xml errors on loading any difficulty

Updated Mod attached

image
image
[OpenApoc Extended Weapons Mod - V9 PREVIEW 2.zip](https://github.com/OpenApoc/OpenApoc/files/3720597/OpenApoc.Extended.Weapons.Mod.-.V9.PREVIEW.2.zip)

FilmBoy84:
NP, it's because i hadn't removed cities.xml from the mod
It was therefore calling for what cities.xml was but finding that there was no subdirectory in the mod with that name

Perhaps the loader should use default at that difficulty if not found?

FilmBoy84:
Latest EWM for testing loading of difficulties

[OpenApoc Extended Weapons Mod - V9 PREVIEW 3.zip](https://github.com/OpenApoc/OpenApoc/files/3720609/OpenApoc.Extended.Weapons.Mod.-.V9.PREVIEW.3.zip)

PR #790, opened 2019-11-06
Add Burst Mode for Weapons

FilmBoy84:
Another attempt at adding burst mode for weapons both in the cityscape and battlescape.


game/state/city/vequipment.cpp
```diff
@@ -67,12 +69,22 @@ bool VEquipment::fire(GameState &state, Vec3<float> targetPosition, Vec3<float>
 	}
 	this->reloadTime = type->fire_delay;
 	this->weaponState = WeaponState::Reloading;
+
+	number_of_shots = this->type->burst;
+	if (number_of_shots == 0)
```
JonnyH: Would this be easier to set the 'default' burst to 1?

game/state/city/vequipment.cpp
```diff
@@ -89,22 +101,27 @@ bool VEquipment::fire(GameState &state, Vec3<float> targetPosition, Vec3<float>
...
 	auto fromScaled = vehicleMuzzle * VELOCITY_SCALE_CITY;
 	auto toScaled = targetPosition * VELOCITY_SCALE_CITY;
-	City::accuracyAlgorithmCity(state, fromScaled, toScaled, type->accuracy + owner->getAccuracy(),
-	                            targetVehicle && targetVehicle->isCloaked());
 
-	Vec3<float> velocity = toScaled - fromScaled;
-	velocity = glm::normalize(velocity);
-	// I believe this is the correct formula
-	velocity *= type->speed * PROJECTILE_VELOCITY_MULTIPLIER;
+	for (number_of_shots; number_of_shots > 0; number_of_shots--)
```
JonnyH: There's no need for the first statement in this for loop, it could just be:
for (;number_of_shots > 0; number_of_shots--)
{....}
But then you could just do:
while (number_of_shots--)
{....}

Or it might be more explicit to do:

for (int shot_number = 0; shot_number < number_of_shots; shot_number++)
{....}

or similar - adding new variables isn't bad, especially if those variables are well named after what they represent (free documentation :)

game/state/city/vequipment.cpp
```diff
@@ -89,22 +101,27 @@ bool VEquipment::fire(GameState &state, Vec3<float> targetPosition, Vec3<float>
...
-	auto projectile = mksp<Projectile>(
-	    type->guided ? Projectile::Type::Missile : Projectile::Type::Beam, owner, targetVehicle,
-	    homingPosition, muzzle, velocity, type->turn_rate, type->ttl, type->damage, /*delay*/ 0,
-	    /*depletion rate*/ 0, type->tail_size, type->projectile_sprites, type->impact_sfx,
-	    type->explosion_graphic, state.city_common_image_list->projectileVoxelMap, type->stunTicks,
-	    type->splitIntoTypes, manual);
-	owner->tileObject->map.addObjectToMap(projectile);
-	owner->city->projectiles.insert(projectile);
+		auto projectile = mksp<Projectile>(
+		    type->guided ? Projectile::Type::Missile : Projectile::Type::Beam, owner, targetVehicle,
```
JonnyH: I know it was like that previously, but this is getting *very* busy. Possibly pulling out the ternary expression would help?

eg. add a like like:
` const auto projectile_type = type->isGuided ? Projectile::Type::Missile : Projectile::Type::Beam;`

then use the "projectile_type" variable as the first arg would help.

Everything else appears to be trivial expressions, but I could see that being missed from a quick scan.

game/state/shared/aequipment.cpp
```diff
@@ -702,12 +702,19 @@ void AEquipment::fire(GameState &state, Vec3<float> targetPosition, StateRef<Bat
...
 	Vec3<float> originalTarget = targetPosition;
+	int number_of_shots;
 
 	if (payload->fire_sfx)
 	{
 		fw().soundBackend->playSample(payload->fire_sfx, unit->position);
 	}
 
+	number_of_shots = this->type->burst;
+	if (number_of_shots == 0)
```
JonnyH: Similarly, would setting the default to "1" mean less code here?

game/state/shared/aequipment.cpp
```diff
@@ -730,42 +737,51 @@ void AEquipment::fire(GameState &state, Vec3<float> targetPosition, StateRef<Bat
 			return;
 		}
 		// Throw item (accuracy applied inside)
-		item->throwItem(state, targetPosition, velocityXY, velocityZ, true);
+
+		for (number_of_shots; number_of_shots > 0; number_of_shots--)
```
JonnyH: Same unnecessary statement in for loop as above

game/state/shared/aequipment.cpp
```diff
@@ -730,42 +737,51 @@ void AEquipment::fire(GameState &state, Vec3<float> targetPosition, StateRef<Bat
...
+			// inside
+			// the soldier
+			unitPos += velocity * 3.5f / 8.0f;
+			// Scale velocity according to speed
+			velocity *= payload->speed * PROJECTILE_VELOCITY_MULTIPLIER;
+
+			if (state.current_battle->map->tileIsValid(unitPos))
+			{
+				auto p = mksp<Projectile>(
+				    payload->guided ? Projectile::Type::Missile : Projectile::Type::Beam, unit,
```
JonnyH: Another ternary expression that might be pulled out of a massive argument list?

game/state/city/vequipment.cpp
FilmBoy84: That is what i initially changed from Andrey's code. HOWEVER, because many weapons do not have a <burst> value set, it defaults to null/0 causing a CTD unless i explicitly state that a null/0 value is actually a single shot.

Additionally, this protects the game from bad modding etiquette where modders may, instead of just removing <burst> from the XML just simply state "0" as the value out of laziness or misunderstanding

It's a little over-engineered, but at least the above method is largely foolproof with regards to avoiding problems further down the line; most of which will be generated by any modding community.

game/state/shared/aequipment.cpp
FilmBoy84: See response to the same code in vequipment.cpp

This was the only way i could get around <burst> not being specified and returning a null value or modders simply giving "0" as a value rather than removing the line completely

game/state/city/vequipment.cpp
FilmBoy84: Taking your comment below into account also, would this be better for this entire section of code?

(Also consider my reluctance to change from a check on null/0 value for number_of_shots in <burst> due to the aforementioned issues )

`for (int shot_number = 0; shot_number < number_of_shots; shot_number++)
	{
		City::accuracyAlgorithmCity(state, fromScaled, toScaled,
		                            type->accuracy + owner->getAccuracy(),
		                            targetVehicle && targetVehicle->isCloaked());

		Vec3<float> velocity = toScaled - fromScaled;
		velocity = glm::normalize(velocity);
		// I believe this is the correct formula
		velocity *= type->speed * PROJECTILE_VELOCITY_MULTIPLIER;

		const auto projectile_type = type->isGuided ? Projectile::Type::Missile : Projectile::Type::Beam; owner, targetVehicle,
		    homingPosition, muzzle, velocity, type->turn_rate, type->ttl, type->damage, /*delay*/ 0,
		    /*depletion rate*/ 0, type->tail_size, type->projectile_sprites, type->impact_sfx,
		    type->explosion_graphic, state.city_common_image_list->projectileVoxelMap,
		    type->stunTicks, type->splitIntoTypes, manual);
		owner->tileObject->map.addObjectToMap(projectile);
		owner->city->projectiles.insert(projectile);
	}`

game/state/shared/aequipment.cpp
FilmBoy84: Change to?

```for (int shot_number = 0; shot_number < number_of_shots; shot_number++)```

File: game/state/rules/aequipmenttype.h [hunk 7]
```diff
@@ -118,6 +118,7 @@ class AEquipmentType : public StateObject
 
 	// Ammo, Weapons & Grenades with built-in ammo, General with charge
 	int max_ammo = 0;
+	int burst = 0;
```
JonnyH: WRT the previous "default" commit, changing this to "=1" should mean that all instances of the AEquipmentType that don't explicitly specify a burst value in the xml should have it set to "1", and as the "bust" member currently doesn't exist, then no xml should contain said value, so shouldn't be an issue?

Then we can remove the "0 is actually 1" workaround in the firing code.

Or was this what you tried previously, and there's something else that's overwriting it with a '0', or otherwise modifying the value? If so, then /that/ is probably a bug.

game/state/city/vequipment.cpp
JonnyH: Yes, I think that looks a bit clearer - that's probably the "most common" for loop construct style, so most experienced c++ coders will instantly recognise what the loop is doing - and optimising for "developer understanding" is a good goal.

game/state/rules/aequipmenttype.h
FilmBoy84: The problem only occurs when lazy modding (or use of editors) fails to remove the <burst> tag from a piece of equipment with it specified using a "0" instead of removing the line completely.

In these instances, if a 0 value is specified, instead of complete removal of the <burst> line. The game crashes. It needs to be told that if a 0 is explicitly given, this is in fact a 1.

For example, we have a weapon like the one below added, the modder decides they do not want burst, and instead of removing the line completely just specifies "0".

Without the code to recognise 0 as a 1, OpenApoc will CTD

```
<entry>
		<key>AEQUIPMENTTYPE_GAUSS_PISTOL_CLIP</key>
		<value>
			<type>Ammo</type>
			<id>AEQUIPMENTTYPE_GAUSS_PISTOL_CLIP</id>
			<name>Surplus Gauss Pistol Clip</name>
			<weight>1</weight>
			<held_image_pack>BATTLEUNITIMAGEPACK_item16</held_image_pack>
			<dropped_sprite>PCK:xcom3/tacdata/gameobj.pck:xcom3/tacdata/gameobj.tab:16</dropped_sprite>
			<dropped_offset>
				<x>23</x>
				<y>34</y>
			</dropped_offset>
			<dropped_shadow_sprite>PCKSHADOW:xcom3/tacdata/oshadow.pck:xcom3/tacdata/oshadow.tab:16</dropped_shadow_sprite>
			<shadow_offset>
				<x>23</x>
				<y>6</y>
			</shadow_offset>
			<equipscreen_sprite>ufodata/pequip/gausspistolclip.png</equipscreen_sprite>
			<equipscreen_size>
				<x>1</x>
				<y>1</y>
			</equipscreen_size>
			<manufacturer>ORG_SUPERDYNAMICS</manufacturer>
			<store_space>5</store_space>
			<armor>40</armor>
			<score>1</score>
			<research_dependency>
				<type>All</type>
			</research_dependency>
			<weapon_types>
				<entry>AEQUIPMENTTYPE_GAUSS_PISTOL</entry>
			</weapon_types>
			<max_ammo>20</max_ammo>
			<burst>0</burst>
			<speed>60</speed>
			<projectile_sprites>
				<entry>bulletsprites/battle/35.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/23.png</entry>
				<entry>bulletsprites/battle/24.png</entry>
				<entry>bulletsprites/battle/25.png</entry>
				<entry>bulletsprites/battle/26.png</entry>
				<entry>bulletsprites/battle/27.png</entry>
				<entry>bulletsprites/battle/28.png</entry>
				<entry>bulletsprites/battle/29.png</entry>
				<entry>bulletsprites/battle/30.png</entry>
				<entry>bulletsprites/battle/31.png</entry>
				<entry>bulletsprites/battle/32.png</entry>
				<entry>bulletsprites/battle/33.png</entry>
				<entry>bulletsprites/battle/34.png</entry>
				<entry>bulletsprites/battle/35.png</entry>
				<entry>bulletsprites/battle/36.png</entry>
				<entry>bulletsprites/battle/37.png</entry>
				<entry>bulletsprites/battle/38.png</entry>
				<entry>bulletsprites/battle/39.png</entry>
			</projectile_sprites>
			<damage>45</damage>
			<accuracy>40</accuracy>
			<fire_delay>240</fire_delay>
			<tail_size>36</tail_size>
			<range>900</range>
			<ttl>2700</ttl>
			<explosion_graphic>DOODAD_23_PLASMA</explosion_graphic>
			<fire_sfx>RAWSOUND:rawsound/tactical/weapons/gausspistol1.wav:11025</fire_sfx>
			<impact_sfx>RAWSOUND:rawsound/tactical/weapons/gausshit1.wav:11025</impact_sfx>
			<damage_type>DAMAGETYPE_ARMOR_PIERCING</damage_type>
			<trigger_type>Timed</trigger_type>
		</value>
	</entry>
```

FilmBoy84:
And i've gone and buggered that up.
Will leave so you can help me see what i've done wrong here

game/state/rules/aequipmenttype.h
JonnyH: but currently, no mods should have a <burst> tag, as it currently doesn't do anything (and is ignored and not serialized out in the save).

So someone has to /add/ an invalid burst tag, which feels like something we can't really 'defend' against anyway - you can always break a mod by changing the data of the mod :)

FilmBoy84: It was something i noticed when using the EWM with builds for this PR

A couple of modders are starting to take copies of the EWM and make their own changes and with that in mind, im preempting someone just changing values to 0 sooner or later

Particularly as one of the main reasons for this PR is to add OG functionality to the 52mm Cannon and allow me to make burst weapons in the battlescape (which OG does not have).

To replicate the easiest thing to do is change the Airguard 52mm Cannon Ammo from <burst>2</burst> as this PR makes it and give a <burst>0</burst> value. It also means you don't have to set up a battlescape situation to replicate, just send out the starting stormdog to attack something when you start a new game. OpenApoc will CTD when it is fired.

In short, i know to remove the line completely to avoid a CTD if we default to 1 as you wish

When things get out into the real world, i cannot trust others to know that

So as a fail-safe (and to reduce answering questions about why XXX doesn't work or causes a CTD) it's easier for the code to just effectively go "there's a 0, this is actually a 1" and not have to deal with the questions from modders later

If you wish, i can change this (after-all im still learning this stuff, so have no real grounding on the right thing to do), but i do fear where it will head without a correction of 0/null values and modders tending to just C&P stuff modifying afterwards.

JonnyH: I'd very much prefer not to handle "invalid" values in code with a "We (probably) know what you actually meant" - as it can add a lot of complexity going forward. Trying to defend against every possible mistake is how you get into a mire of guessing what the user /wanted/, instead of just getting them to fix it. And by allowing "bad" values today, that just means you'll have to allow them forever, otherwise people will whine about breaking compatibility then.

But a clear error message might be useful (e.g. if (burst == 0) { LogError("Weapon WEAPON_ID has an invalid burst value of 0");})

Then there's no excuse - best to make sure things are 'correct' in the fist place, and give the mod dev as clear an idea of what's wrong, than silently seat and modify bad values.

Generally, 'data validation' should be done at data read time, not 50 callstacks deep where you'll never see it again. Maybe we should have some 'validity checks' in the serialization code?

And trying to 'correct' bad input data is the route to madness, so starting down that path would be a mistake IMHO.

game/state/city/vequipment.cpp
```diff
@@ -89,22 +101,27 @@ bool VEquipment::fire(GameState &state, Vec3<float> targetPosition, Vec3<float>
...
+		City::accuracyAlgorithmCity(state, fromScaled, toScaled,
+		                            type->accuracy + owner->getAccuracy(),
+		                            targetVehicle && targetVehicle->isCloaked());
+
+		Vec3<float> velocity = toScaled - fromScaled;
+		velocity = glm::normalize(velocity);
+		// I believe this is the correct formula
+		velocity *= type->speed * PROJECTILE_VELOCITY_MULTIPLIER;
 
-	auto projectile = mksp<Projectile>(
```
JonnyH: I think this should be:
```
const auto projectile_type =
		    type->isGuided ? Projectile::Type::Missile : Projectile::Type::Beam;

auto projectile = mksp<Projectile>(projectile_type, owner, targetVehicle,
		    homingPosition, muzzle, velocity, type->turn_rate, type->ttl, type->damage, /*delay*/ 0,
		    /*depletion rate*/ 0, type->tail_size, type->projectile_sprites, type->impact_sfx,
		    type->explosion_graphic, state.city_common_image_list->projectileVoxelMap,
		    type->stunTicks, type->splitIntoTypes, manual);
```

FilmBoy84:
Have updated branch with Jonny's Appveyor fix until this PR is in a state to merge
That way builds should only fail if i've messed up
Not because of Appveyor :)

game/state/shared/aequipment.cpp
```diff
@@ -702,12 +702,19 @@ void AEquipment::fire(GameState &state, Vec3<float> targetPosition, StateRef<Bat
...
 	auto payload = getPayloadType();
 	Vec3<float> originalTarget = targetPosition;
+	int number_of_shots;
 
 	if (payload->fire_sfx)
 	{
 		fw().soundBackend->playSample(payload->fire_sfx, unit->position);
 	}
 
+	number_of_shots = this->type->burst;
```
JonnyH: Should this be clamped against the current ammo?

e.g. `std::min(this->type->burst, this->ammo)` would make sure it only fires at most as many shots are loaded

JonnyH: Also, this currently only takes a single ammo out of the clip no matter the size of the burst - is this intended? IE in burst weapons should the "ammo" count actually be the "number of bursts?" Or the number of individual shots?

FilmBoy84: It's by design, but thinking about what you've said, that needs to change

My initial thinking was "It's a burst weapon, just specify the number of bursts as ammo"

However, and the more i think about it, I probably want to treat it as "amount of rounds". This way if i ever allow for multiple fire modes with modded weapons, for example, Single Shot, Burst and Full Automatic, it counts by round and not just how many times you can pull the trigger to whatever result

Thanks for prompting me to think about this :) It makes sense to change it now to save issue in the future

I'll change line 712 to what you suggested (I suspect also i will have to look at vequipment.cpp also)

As to how to get it to deduct only the number of rounds used by the fire mode, how would i do this?

game/state/rules/aequipmenttype.h
FilmBoy84: Agreed, i'll remove the argument from both agent and vehicle equipment

```
if (number_of_shots == 0)
	{
		number_of_shots = 1;
	}
```

Removed from aequipment.cpp and vequipment.cpp :)

FilmBoy84:
Hmmmm, nope, the result of those four commits is now the Stormdog an other burst weapons make a fire sound but do not throw a projectile or use any ammo!

Gah!

FilmBoy84:
Stormdog working again with latest commit

Still not getting weapons in battlescape to use burst though (or consume the burst rate of ammo)

https://youtu.be/mklYDE4864k

FilmBoy84:
I have a suspicion that (for agent equipment)

```number_of_shots = std::min(this->type->burst, this->ammo);```

Should be

```number_of_shots = std::min(this->type->burst, this->shot_number);```
OR
```number_of_shots = std::min(this->type->burst, this->payload);```

Your thoughts?

FilmBoy84:
OK, List of Current Issues:

We need the burst mode to deduct the amount of ammo specified in the burst value from the weapon or it's clips... How would this be done?

For some reason, whilst burst mode works in the cityscape, it does NOT work in the battlescape.

JonnyH:
The GameState storeage and patching looks good - as otherwise the extracted gamestate wouldn't contain the <burst> attribute, but that seems to be correct in the generated base_gamestate/gamestate/vehicle_equipment.xml

JonnyH:
And in a quick test the debuggercannon seems to correctly burst 10 shots at a time - what about the battlescape doesn't work?

JonnyH:
And is it a good idea to reduce the ammo by "burst_count" rather than 1 every shot? As if it's not really possible to use a non-burst-multiple of ammo, it's just busywork making the user mentally divide by the burst count to see how many times they can shoot?

And what happens if it doesn't divide - does it only shoot the number left in the mag?

It may be better redefining "Ammo count (in bullets)" as "Ammo count (in number of bursts)" - they just happen to be the same for anything with a burst of "1"

FilmBoy84:
> ...what about the battlescape doesn't work?

Try the current Burst Ammo build with the EWM preview attached... The gauss weapons have a <burst> value specified (rifle is the first one to have a true burst at the mo with a value of 2, heavy a value of 3). Gauss Pistol has only a value of 1 specified, so it should be no different to other weapons for now.

Sadly, with this mod, the Rifle does not burst two rounds, nor does the heavy burst three... It's most confusing as the code (and debugger cannon) should say it works... But it does not.

This is why i was wondering if it was a mod loading issue

[OpenApoc Extended Weapons Mod - V9 PREVIEW 4.zip](https://github.com/OpenApoc/OpenApoc/files/3891180/OpenApoc.Extended.Weapons.Mod.-.V9.PREVIEW.4.zip)

Atrosha:
How to activate this "burst mode" for agents at Battlescape?

FilmBoy84:
> 
> 
> How to activate this "burst mode" for agents at Battlescape?

Currently, you need a weapon capable of burst mode specified in the XML

The EWM Gauss Rifle and Heavy Gauss have this

But for some reason they are NOT working in the battlescape, hence the PR being stuck

FilmBoy84:
Still can't work out what's missing

Currently, the last outstanding issue is that bursts are not done by battlescape weapons

Cityscape works as intended

But Battlescape should work and is not

FilmBoy84:
Hmmm, tried getting back on this today
I still cannot get it to work with weapons in the Battlescape
Cityscape is fine and vehicles with <burst> weapons fire as expected, but Battlescape just doesn't seem to fire bursts

Also, how ammo is dealt with needs looking at in terms of how I resolve partial clips or issues when there are not enough spare rounds to fire a full burst - does it just fire what's available? Or, as JonnyH has previously suggested, do we just make every burst take one ammo from a clip - regardless of number of projectiles shot? This would be easier, but may cause problems if further firing modes are added later?

JonnyH:
How are you testing this? I note that the "BattleView::debugShot()" function spawns it's own projectiles, and therefore doesn't go through the 'standard' AEquipment::fire() function (and thus doesn't handle burst).

FilmBoy84:
Yes, Debug Cannon is not used to test this effect (I do wonder if, once merged, however we should update the Debug Cannon to use <burst> )

I've been using the Gauss Rifle and Heavy Gauss in the EWM mod version in the comments above (v9 preview 4). They have a burst value of 2 and 3 respectively so should behave differently.

The Gauss Pistol does not burst (although it has a value of burst 1 to test that this retains a single shot)

FilmBoy84:
_ORIGINAL MERGE TEXT_

**Add <burst> mode for weapons mods allowing multiple projectiles to be fired per pull of the trigger**

Add the option for mods to specify a <burst> value to weapons - this fires the given value of projectiles every time the weapon is fired. 

Ammunition is deducted according to the value set by <burst> also to allow the same clip to be used in weapons with different <burst> values.

Partial Clips with not enough ammo to fire the complete <burst> value will only fire the projectiles available from the remaining ammunition count in the clip.

Discussion has been had about each clip being treated as the number of "whole" bursts but this is open to exploits when clips are shared between weapons - for that reason, 1 projectile is 1 count of ammunition.

A great many thanks to JonnyH and Jarskih for rebasing and bringing this PR to fruition

PR #793, opened 2019-11-10
Lock the framework update loop to a set FPS (default 60)

JonnyH:
This sleep()s if we're going faster than the expected update rate,
hopefully avoiding any possible issues due to different vsync limits
causing significantly different in-game speeds - as it's set to
progress 1 tick every frame.

Note that if you are 60fps vsync limited (or lower) there's a good
chance the update loop will 'fall behind' due to natural jitter and
precision differences, it'll LogWarning() if that happens, but it
shouldn't cause any real issues, so long as the difference is small.

If you somehow ended up with some weird setup with a significantly lower
than 60fps vsync, however, it will run slower.

### DIFF
--- a/framework/framework.cpp
+++ b/framework/framework.cpp
@@ -396,14 +396,39 @@ void Framework::run(sp<Stage> initialStage)
 	TRACE_FN;
 	LogInfo("Program loop started");
 
+	auto target_frame_duration =
+	    std::chrono::duration<int64_t, std::micro>(1000000 / Options::targetFPS.get());
+
 	p->ProgramStages.push(initialStage);
 
 	this->renderer->setPalette(this->data->loadPalette("xcom3/ufodata/pal_06.dat"));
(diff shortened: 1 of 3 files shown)


JonnyH:
Naturally, a 'better' solution would be to unlink the updat()/render() stuff, allow them to be thread-safe, and trigger each from a timed event, unlinking the two (so you could have multiple update()s per render, or visa-versa, though that's likely useless as multiple renders without an update() will draw exactly the same thing as there's no frame interpolation or similar going on).

FilmBoy84:
With regards to unlinking, even if the renderer draws the same, would it at least allow us to specify a fixed ticks per second of 72 and let the renderer work out the maths of only showing at 60fps?

As discussed on discord, having the TPS as a straight 72 would be a hell of a lot easier for purposes of converting OG values to OpenApoc rather than having to deal with decimals and none-whole-values.

PR #794, opened 2019-11-12
WIP MacOS App bundle creation

JonnyH:
Very much a WIP:
- Only the OpenApoc executable is packaged - we probably want /at least/
the launcher in there too
- The post-build copy cmake command for the data/ directory into the
resources feels like a hack - same with the fixed location .app bundle
name where it's copying them to? Probably some cmake property we can
query?
- It seems the working directory of app bundles is always "/"? Which
breaks all the loading "local" data, so we chdir() based on the SDL
"base directory", which seems to point to the Resources dir in the
bundle?
- Keying the above off the working directory being "/" is a hack - e.g.
run from the terminal it may be unlikely and maybe a bad idea, but
possible that "/" was the intended working directory - at which point I
don't really know that the SDL "base dir" would be?
- A whole stack of the app bundle metadata isn't set (authors,
"friendly" name, copyright etc.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,8 +103,17 @@ if (WIN32)
 	set(sources ${sources} "apocicon.rc")
 endif()
 
+
+if (APPLE)
+set(MACOSX_BUNDLE_BUNDLE_NAME OpenApoc)
+
+ADD_EXECUTABLE(${CMAKE_PROJECT_NAME} MACOSX_BUNDLE game/main.cpp ${sources}
+		${FRAMEWORK_SOURCES})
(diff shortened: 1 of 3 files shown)


JonnyH:
Still a 'work in progress', but makes things less awful on macos ;)

JonnyH:
Merging because while it's not very clean or complete, it's still better than what's there today (and due to macos 10.15 now defaulting to hidpi mode unless package in an .app with it's corresponding metadata, and that not working and drawing everything at 1/2 size makes it a PITA to work on other stuff on macos....

Issue #796, opened 2019-11-13
VEquipScreen draws equipment in top-left corner


dl471:
There seems to be an issue with the vequipscreen where for the first few frames it draws the inventory items in the top-left hand corner of the screen. Usually it's not that noticeable since it goes by so fast and the users attention is focused on the centre of the screen, but now that I know to look for it I can't unsee it. Seems to happen in the Windows and Linux versions.

This is what it looks like:

![open_apoc_vequip](https://user-images.githubusercontent.com/32369415/68722600-d9d02780-05ad-11ea-927d-547089042a22.png)

makus82:
https://github.com/OpenApoc/OpenApoc/issues/572

Issue #572, opened 2019-04-29
Vehicles list not showed when sceen open, only after mouse over on vehicles list appears


makus82:
Vehicles list not showed when sceen open, only after mouse over on vehicles list appears

![image](https://user-images.githubusercontent.com/11581207/56876924-40975480-6a53-11e9-85c1-f7741c718521.png)
![image](https://user-images.githubusercontent.com/11581207/56876926-4725cc00-6a53-11e9-8fa3-6d35937e8227.png)

makus82:
Done! Thx emptyVoid?

PR #797, opened 2019-11-13
Avoid drawing vehicle equipment at (0, 0)

dl471:
Proposed fix for issue #796. Since the source of the problem appears to be the location of the inventory on screen being off for the first few frames avoiding drawing the inventory until the correct coordinates come along should fix it.

### DIFF
--- a/forms/control.cpp
+++ b/forms/control.cpp
@@ -1013,6 +1013,7 @@ void Control::copyControlData(sp<Control> CopyOf)
 
 	CopyOf->Name = this->Name;
 	CopyOf->Location = this->Location;
+	CopyOf->resolvedLocation = this->resolvedLocation;
 	CopyOf->Size = this->Size;
 	CopyOf->SelectionSize = this->SelectionSize;
 	CopyOf->BackgroundColour = this->BackgroundColour;
--- a/forms/listbox.cpp
+++ b/forms/listbox.cpp
@@ -292,6 +292,7 @@ void ListBox::addItem(sp<Control> Item)
 		selected = Item;
 	}
 	this->setDirty();
+	this->update();
 }
 
 void ListBox::replaceItem(sp<Control> Item)
--- a/forms/multilistbox.cpp
+++ b/forms/multilistbox.cpp
@@ -284,6 +284,7 @@ void MultilistBox::addItem(sp<Control> Item)
 {
 	this->setDirty();
 	Item->setParent(shared_from_this());
+	this->update();
 }
 
 void MultilistBox::replaceItem(sp<Control> Item)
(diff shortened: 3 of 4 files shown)


JonnyH:
It feels like skipping any drawing to {0,0} is a hacky workaround - I think the best solution would be to find why the positions aren't yet set, and ensure that they're correctly positioned before the first draw call.

I suspect the Forms don't setup their initial location correctly, perhaps calling the recalculate-position functions whenever a form is modified (e.g. a new child added) would fix that?

I can see this biting us in the ass later, as it may be possible for a modded UI to have {0,0} as the expected position, then we get bug reports on why it's not drawing correctly, or the 'wrong' value stops being {0,0}, and we get the same incorrect position issues.

dl471:
That's a good point. In retrospect I was making quite a lot of assumptions about presumed correct behaviour. 

I'll dig around the Forms and see if I can get a better idea of what's going on with it.

dl471:
Sorry for disappearing for a while.

As best as I can tell the issue arises when forms are copied. The initial location is correct but when the form is copied the resolvedLocation is not copied with it. This leaves it at zero until the next time update() is called. This can be resolved by copying the resolvedLocation when making a copy or having setParent call update() when it is called but these seem to reintroduce issue #572. That in turn can be resolved by having ListBox and MultiListBox update themselves when new items are added. I've pushed commits that show what this would look like.

Alternatively, both of those issues can be solved by calling the update function for the vequipscreen and vehicleselection at the end of the vequipscreen's begin() function. Not as general a solution as above but it has the advantage of localizing any changes made to that function alone. Since the first approach messes with various files and causes weird behaviour that could crop up elsewhere trying to limit the scope of the change like that might be preferable.

Would you have any comments on these approaches?

emptyVoid:
Looks neat to me.

Though for the sake of optimization I would prefer the second approach -- update lists after populating them with all the items.

IMO better yet would be to use the first approach (updating on item insertion) with an ability to suspend updates, e.g.:
```c++
class Control : public std::enable_shared_from_this<Control>
{
  private:
    int updateSuspensions = 0;
    bool pendingUpdate = false;

  protected:
    bool isUpdateSuspended() const { return updateSuspensions > 0; }

  public:
    void Control::suspendUpdate() { ++updateSuspensions; }
    void Control::resumeUpdate()
    {
      --updateSuspensions;
      if (updateSuspensions == 0 && pendingUpdate)
        update();
    }

    virtual void update()
    {
      if (isUpdateSuspended())
      {
        pendingUpdate = true;
        return;
      }

      pendingUpdate = false;

      ...
    }
}
```
So that forms with lists (such as `VEquipScreen`)  could suspend updates while populating.

Although this would require adjusting logic of all the controls.

PR #800, opened 2019-12-14
Adding Fedora or other RedHat distro building instructions to README

Atrosha:
Maybe somebody suggest more light solution for this:
`yum groupinstall "Development Tools" "Development Libraries"`

This command works like `apt install build-essential` in Ubuntu and installs boost libs also.


JonnyH:
Yeah, looks like you somehow marked everything executable - maybe a bad chmod -R, or you put it on a filesystem that doesn't support unix permissions (FAT or NTFS?)

The readme change itself LGTM - but at some point it may be worth either splitting out different platform build instructions to other files, or some other way, as there are more distros than I can count :P

Atrosha:
NTFS, yes. I just move my sources to other disk while changing Linux distro.

PR #811, opened 2020-01-15
fix of readme windows building instructions

Atrosha:
Tnx @JonnyH

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,9 +109,20 @@ git submodule update --init --recursive
 
 * All the other dependencies (Boost, SDL2) need to be supplied separately. Install [Vcpkg](https://github.com/Microsoft/vcpkg) and run the following command:
 
+* For x64 builds:
+
 ```cmd
-vcpkg install sdl2 boost-locale boost-program-options boost-uuid boost-crc qt5-base
+vcpkg --triplet x64-windows install sdl2 boost-locale boost-program-options boost-uuid boost-crc qt5-base
 ```
+
+* For x86 builds:
+
+```cmd
+vcpkg --triplet x86-windows install sdl2 boost-locale boost-program-options boost-uuid boost-crc qt5-base
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,10 +109,21 @@ git submodule update --init --recursive
...
+```cmd
+vcpkg --triplet x86-windows install sdl2 boost-locale boost-program-options boost-uuid boost-crc qt5-base
+```
+
+* For list of all supported by Vcpkg architectures: `vcpkg help triplet`
+
+
 * Copy the original XCom:Apocalypse .iso file into the "data/" directory. This could also be a directory containing all the extracted files from the CD, and it should be named the same (IE the directory should be data/cd.iso/). This is used during the build to extract some data tables.
 * Open the Git directory in Visual Studio (if you don't have an Open Folder option, generate a project from CMake).
 * Visual Studio should automatically detect and configure CMake appropriately. To add your Vcpkg dependencies, edit your CMake Settings file and add:
```
JonnyH: Maybe we should add a comment here that visual studio 2019 has box to select the toolchain file in the GUI, so no custom json needed?

Atrosha: Maybe better to move building instructions from README to special Wiki section. And just leave a link to Wiki in README?

FilmBoy84: Wikis go down, best to keep text instructions with the source
That way, even if things get outdated, they can always be updated later

Also, apart from JonnyH's request LGTM too

JonnyH: I think it may be worth splitting out the build instructions - e.g. having a BUILD_windows.md/BUILD_ubuntu.md/BUILD_whatever.md

For the most part, "other" platform build instructions aren't useful, so just pad the readme and make the "useful" data less visible

Atrosha: Some clearance... I was suggested to support Wiki not for make README shorter (I saw a bigger mds) but for editing easiness. In Wiki you don't need to open PR for edit two lines in mds.

Issue #799, opened 2019-11-15
Losing a base breaks base building (new bases overwrite old ones)


dl471:
When a player loses a base (through losing a base defence mission, destroying it with debug commands, etc.) they will become effectively unable to build new bases depending on which base they lose. Any save with multiple bases should be able to reproduce this.

The source of this problem seems to be how bases are tracked internally. 

https://github.com/OpenApoc/OpenApoc/blob/e87c9a92e224becd3e56406a8d1aeef84eb6cb95/game/ui/city/basebuyscreen.cpp#L144-L146

The bases are stored in a map, and the keys for that map are generated by combining a constant string and the number of bases that the player has. This works fine when the number of bases is always increasing. However if the number of bases falls then there is a chance that a new base will end up having the same identifier as an old base.

More specifically, if a player has Base 1 and Base 2 then loses a base then the next base they build will be Base 2. If they lost Base 2 this isn't a problem, but if they lost Base 1 then there will be two Base 2s fighting over the BASE_2 key. The new base will overwrite the old base.

This could be fixed by changing how the identifiers and/or names of the bases are generated to avoid collisions. Though how best to do this I'm not sure.

emptyVoid:
No longer reproducible after #814.

FilmBoy84:
No further reports, should be fixed
Closing
If the problem returns, please create a new issue and reference this as appropriate

PR #813, opened 2020-01-31
Fix to #799

Jarskih:
Refreshes base icons after losing a base.
Keeps track of the base number uses it to continually name new bases.

Merged commits for this one.


FilmBoy84:
Thanks for this Jarskih <3
I'll try and get a build together tomorrow for testing as Appveyor is failing
But reading through the commit LGTM :)

JonnyH:
So somehow the gamestate_serialize.xml file ended up with inconsistent line endings - which messed up loading on windows, causing the appveyor failure above.

I put a fixed version in #814 FYI

Issue #692, opened 2019-08-14
Destroying all the tiles of the player's base building crashes


JonnyH:
Seems repro-able just using the ctrl-alt-shift click to destroy the whole player's building.

Valgrind backtrace:

==10587== Invalid read of size 8                                                                [237/1915]
==10587==    at 0x4CED540: local_Rb_tree_increment (tree.cc:62)                                           
==10587==    by 0x4CED540: std::_Rb_tree_increment(std::_Rb_tree_node_base*) (tree.cc:85)                 
==10587==    by 0x51B397: operator++ (stl_tree.h:366)                                                     
==10587==    by 0x51B397: OpenApoc::Base::die(OpenApoc::GameState&, bool) (base.cpp:93)                   
==10587==    by 0x524176: OpenApoc::Building::buildingPartChange(OpenApoc::GameState&, glm::vec<3, int, ($
lm::qualifier)0>, bool) (building.cpp:989)                                                                
==10587==    by 0x54427E: OpenApoc::Scenery::collapse(OpenApoc::GameState&) (scenery.cpp:1451)            
==10587==    by 0x546E0D: OpenApoc::Scenery::update(OpenApoc::GameState&, unsigned int) (scenery.cpp:1472$
==10587==    by 0x52F47E: OpenApoc::City::update(OpenApoc::GameState&, unsigned int) (city.cpp:289)       
==10587==    by 0x5F5E45: OpenApoc::GameState::update(unsigned int) (gamestate.cpp:974)                   
==10587==    by 0x315E87: OpenApoc::CityView::update() (cityview.cpp:1749)                                
==10587==    by 0x1F3C90: OpenApoc::Framework::run(std::shared_ptr<OpenApoc::Stage>) (framework.cpp:663)  
==10587==    by 0x1C72E3: main (main.cpp:26)                                                              
==10587==  Address 0x2bc50038 is 24 bytes inside a block of size 88 free'd                                
==10587==    at 0x4836ECB: operator delete(void*) (vg_replace_malloc.c:586)                               
==10587==    by 0x5664A7: deallocate (new_allocator.h:128)                                                
==10587==    by 0x5664A7: deallocate (alloc_traits.h:470)                                                 
==10587==    by 0x5664A7: _M_put_node (stl_tree.h:584)                                                    
==10587==    by 0x5664A7: _M_drop_node (stl_tree.h:651)                                                   
[...]

PR #814, opened 2020-02-01
Refresh the base icons after losing a base

JonnyH:
Should fix issue #799

### DIFF
--- a/game/state/city/base.cpp
+++ b/game/state/city/base.cpp
@@ -88,11 +88,14 @@ void Base::die(GameState &state, bool collapse)
 	}
 	for (auto a : building->currentAgents)
 	{
-		a->die(state, true);
+		auto agent = a;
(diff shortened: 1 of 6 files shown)


JonnyH:
> Not sure if it's necessary. But without copying the vehicle I got random errors while testing. I think its related to how the GameState::cleanUpDeathNote() removes vehicles from the game. Probably a separate issue though.'
> 
> I think this was the issue I encontered: #692

I've seen some issues where the list sp<> was the last reference pinning the object, so holding just a reference to that (which may be how a foreach loop may get the variable) won't copy it, so won't increase the refcount, so if removing it from the list while that reference is still live may end up with a dangling reference to a now-destroyed object.

Long-term, I want to reduce our reliance on sp<> refcounting, which should 'solve' (or at least make more obvious) issues like that.

Issue #826, opened 2020-04-21
Intro cinematic gets corrupted 


emptyVoid:
Intro cinematic gets corrupted when UFO appears.

![image](https://user-images.githubusercontent.com/13639229/79841249-d2fcda80-83bf-11ea-8c1f-4691657a1f71.png)

These console messages appear during playback:
```
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 2 entries from 140 to 141
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 3 entries from 127 to 128
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 4 entries from 124 to 125
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 2 entries from 138 to 139
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 2 entries from 138 to 139
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 3 entries from 148 to 150
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 2 entries from 127 to 128
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 2 entries from 126 to 127
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 3 entries from 124 to 126
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 3 entries from 137 to 138
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 3 entries from 138 to 139
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 3 entries from 127 to 128
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 4 entries from 124 to 125
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 2 entries from 128 to 129
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 6 entries from 126 to 129
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 9 entries from 126 to 127
[...]

PR #828, opened 2020-04-21
Update libsmacker

JonnyH:
This contains a palette aliasing fix - maybe helps with issue #826


Atrosha:
Regression test failed.

Atrosha:
The same with newest changes.

Also... Is `WIP/allow-overlapping-palette-copies` branch needs to be tested?

JonnyH:
Make sure you've updated the dependencies - 'git submodule update --init' should be enough after a checkout.

git by default doesn't change submodules on parent checkout, though some git wrappers do (the inconsistency makes it more annoying than anything)

After the checkout and submodule update,the dependencies/libsmacker version should be 21332e23e4937cd13b2faf39678ff473866b327b ("Don't skip palette copies with overlapping src/dst regions") - which is on the WIP/allow-overlapping-palette-copies branch of the libsmacker repo - though you should rely on the 'git submodule' command to ensure you have the correct version (and 'git status' should not say that the dependencies/libsmacker directory has changed)

Atrosha:
It's my fault, sorry. After `git submodule update --init --recursive` all fine.

Screenshot from 2020-04-21 22-54-28

And finally regression test **passed**.

Issue #784, opened 2019-10-12
Craft Weapons Doubling up when using mods


FilmBoy84:
When using mods that add a new version of the vehicle weapons XMLs, the game appears to specify both sets of weapons (even if identical) on the craft in cityscape

This leads to a doubling up of all weapons in combat with craft shooting twice the number of shots they would normally

It is also clear with a vehicle selected that there are two weapons instead of one in the screen below

The extra weapons do not show in the equip screen however

![image](https://user-images.githubusercontent.com/33913088/66704517-8059ac00-ed14-11e9-952a-5bcb7434c01a.png)

FilmBoy84:
Hmmmm, on playing the EWM preview posted to PR #783 it also appears many things modified and therefore appearing in both the base mod xml and the ewm xml get duplicated

For example, a 3x UFO Type 2 and 2x UFO Type 1 invasion on week 1 becomes a 6x UFO Type 2 and 4x UFO Type 1 invasion

Not sure why the mod loading is happening in such a way as to duplicate values rather than replace them

makus82:
when open skirmish at equipment screen its crash )

emptyVoid:
> Not sure why the mod loading is happening in such a way as to duplicate values rather than replace them

It's due to the way the deserialization of `std::list` (and `std::vector`) is implemented -- any property represented as a list of entries will get the entries from mod added to the base entries.

https://github.com/OpenApoc/OpenApoc/blob/181a4fc46c92413f026af922a6ad291ac1585eb7/game/state/gamestate_serialize.h#L279-L312

One way to deal with this while not breaking existing functionality is to add an optional flag to the modded properties overriding default appending behavior, e.g. something like this:

```xml
<initial_equipment_list>
    <override>true</override>
    <entry>
        <first>
            <x>7</x>
            <y>1</y>
        </first>
        <second>VEQUIPMENTTYPE_RENDOR_PLASMA_GUN</second>
    </entry>
    <entry>
        <first>
            <x>4</x>
            <y>7</y>
        </first>
        <second>VEQUIPMENTTYPE_PROPHET_MISSILE_ARRAY</second>
    </entry>
    <entry>
        <first>
            <x>11</x>
            <y>7</y>
        </first>
        <second>VEQUIPMENTTYPE_JANITOR_MISSILE_ARRAY</second>
    </entry>
    <entry>
        <first>
            <x>6</x>
            <y>12</y>
        </first>
[...]

PR #829, opened 2020-04-21
WIP: add op="delete" attribute to serialization xml

JonnyH:
This adds a way of removing items from lists/maps/vectors/sets in the GameState.

Copied from #784 (which this should help solve):

Adding 'commands' to lists (e.g. 'append'/'delete') in the xml format has been on the todo list for a while - and is necessary for issues like this.

I'm generally leaning towards an attribute instead of a subnode, but I think the bikeshed should be painted mauve.

e.g. a 'op' attribute, only valid on lists/vectors/maps and map keys (though the last can be overridden, it's not currently possible to /remove/ them)

- For lists and vectors (as they're treated the same) the 'op' attribute can gave one of two values, 'append' and 'delete' - 'append' being the default. To replace a list, use 'delete' and then follow it with the new entries as normal.

e.g.
```
<my_list op="delete">
</my_list>
```

would remove all entries from my_list and:

```
<my_list op="delete">
<entry>1</entry>
</my_list>
```

would result in a list with a single entry - "1" - no matter how many entries were previously in the list.

No 'op' attribute would be treated as 'append' - which is the current behavior (any following entries are appended as new objects to the existing list)

- Maps would be 'override' and 'delete' - 'override' being the default, which would be the current behavior (e.g. add new or override if already matching key). 'delete' would replace the entire map with an empty map (removing all entries) before processing any following entries as normal.

e.g.
```
<my_map op="delete">
</my_map>
```

would result in an empty my_map

```
<my_map op="delete">
<entry>
<key>KEY1</key>
<value>VALUE1</value>
</entry>
</my_map>
```

would result in a map with a single entry with the key "KEY1" and "VALUE1"

- Map entries 'op' attribute can have one of the same values - 'override' and 'delete' - override being the default. Entries with a <key> with the op="delete" attribute but no value would be removed from the map, otherwise a new default <value> object would be created and populated as normal.

e.g.

```
<my_map>
<entry op="delete">
<key>KEY1</key>
</entry>
</my_map>
```

would remove a matching key of "KEY1" from the map - TODO: What happens if the key isn't found?

```
<my_map>
<entry op="delete">
<key>KEY1</key>
<value></value>
</entry>
</my_map>
```

would replace (or add) a matching key of "KEY1" from the map with a default constructed object - IE ALL it's members would return to default

```
<my_map>
<entry op="delete">
<key>KEY1</key>
<value>
  <child_value>3</child_value>
</value>
</entry>
</my_map>
```

would replace (or add) a matching key of "KEY1" from the map with a default constructed object then set "child_value" in it - IE ALL OTHER members would return to default

I think that fits all use cases?

### DIFF
--- a/framework/serialization/serialize.cpp
+++ b/framework/serialization/serialize.cpp
@@ -78,6 +78,8 @@ class XMLSerializationNode final : public SerializationNode
 
 	UString getName() override;
 	void setName(const UString &str) override;
+	UString getAttribute(const UString &attribute) override;
+	void setAttribute(const UString &attribute, const UString &value) override;
 	UString getValue() override;
 	void setValue(const UString &str) override;
 
@@ -314,6 +316,16 @@ void XMLSerializationNode::setName(const UString &str) { node.set_name(str.cStr(
 UString XMLSerializationNode::getValue() { return node.text().get(); }
 
 void XMLSerializationNode::setValue(const UString &str) { node.text().set(str.cStr()); }
+
+UString XMLSerializationNode::getAttribute(const UString &attribute)
+{
+	return node.attribute(attribute.cStr()).value();
+}
+
+void XMLSerializationNode::setAttribute(const UString &attribute, const UString &value)
+{
+	node.attribute(attribute.cStr()).set_value(value.cStr());
+}
 
 unsigned int XMLSerializationNode::getValueUInt() { return node.text().as_uint(); }
 
--- a/framework/serialization/serialize.h
+++ b/framework/serialization/serialize.h
@@ -28,6 +28,8 @@ class SerializationNode
 
 	virtual UString getName() = 0;
 	virtual void setName(const UString &str) = 0;
+	virtual UString getAttribute(const UString &attribute) = 0;
+	virtual void setAttribute(const UString &attribute, const UString &value) = 0;
 	virtual UString getValue() = 0;
 	virtual void setValue(const UString &str) = 0;
 
--- a/game/state/gamestate_serialize.h
+++ b/game/state/gamestate_serialize.h
@@ -71,9 +71,7 @@
 namespace OpenApoc
 {
 
-static const UString OVERRIDE_OP_ATTRIBUTE = "override";
 static const UString DELETE_OP_ATTRIBUTE = "delete";
-static const UString APPEND_OP_ATTRIBUTE = "append";
 
 template <typename T> bool operator==(const StateRefMap<T> &a, const StateRefMap<T> &b)
 {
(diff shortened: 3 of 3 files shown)


FilmBoy84:
Testing by adding a single weapon to agent_equipment.xml via a mod with the following code
Resulted in ALL agent equipment in the game being absent until week 2 when the stun mine appears

This seems to confirm that op=delete works as the XML did not contain the OG weapons, so forced them to be ignored on loading gamestate :)

```
<agent_equipment op=delete>
	<entry>
		<key>AEQUIPMENTTYPE_MARSEC_PROXIMITY_STUN_MINE</key>
		<value>
			<type>Grenade</type>
			<id>AEQUIPMENTTYPE_MARSEC_PROXIMITY_STUN_MINE</id>
			<name>Marsec Proximity Mine STUN GAS</name>
			<weight>3</weight>
			<held_image_pack>BATTLEUNITIMAGEPACK_item3</held_image_pack>
			<dropped_sprite>PCK:xcom3/tacdata/gameobj.pck:xcom3/tacdata/gameobj.tab:3</dropped_sprite>
			<dropped_offset>
				<x>23</x>
				<y>34</y>
			</dropped_offset>
			<dropped_shadow_sprite>PCKSHADOW:xcom3/tacdata/oshadow.pck:xcom3/tacdata/oshadow.tab:3</dropped_shadow_sprite>
			<shadow_offset>
				<x>23</x>
				<y>6</y>
			</shadow_offset>
			<equipscreen_sprite>ufodata/pequip/proximitystunmine.png</equipscreen_sprite>
			<equipscreen_size>
				<x>2</x>
				<y>1</y>
			</equipscreen_size>
			<manufacturer>ORG_MARSEC</manufacturer>
			<store_space>10</store_space>
			<armor>60</armor>
			<score>2</score>
			<research_dependency>
				<type>All</type>
			</research_dependency>
			<max_ammo>1</max_ammo>
			<speed>4</speed>
			<projectile_sprites>
				<entry>bulletsprites/battle/15.png</entry>
				<entry>bulletsprites/battle/00.png</entry>
				<entry>bulletsprites/battle/01.png</entry>
				<entry>bulletsprites/battle/02.png</entry>
				<entry>bulletsprites/battle/03.png</entry>
				<entry>bulletsprites/battle/04.png</entry>
				<entry>bulletsprites/battle/05.png</entry>
				<entry>bulletsprites/battle/06.png</entry>
				<entry>bulletsprites/battle/07.png</entry>
				<entry>bulletsprites/battle/08.png</entry>
				<entry>bulletsprites/battle/09.png</entry>
				<entry/>
				<entry>bulletsprites/battle/09.png</entry>
				<entry/>
				<entry/>
				<entry>bulletsprites/battle/09.png</entry>
			</projectile_sprites>
			<damage>100</damage>
			<accuracy>100</accuracy>
			<fire_delay>240</fire_delay>
			<tail_size>16</tail_size>
			<range>400</range>
			<ttl>1000</ttl>
			<damage_type>DAMAGETYPE_STUN_GAS</damage_type>
			<trigger_type>Proximity</trigger_type>
			<explosion_depletion_rate>15</explosion_depletion_rate>
		</value>
	</entry>
</agent_equipment>
```

JonnyH:
I'll merge this, as even if it's not perfectly tested it shouldn't *regress* anything, and keeping mod format changes out-of-tree for any length of time just makes it more difficult for everyone involved.

PR #833, opened 2020-04-23
Fix 'canEnterDimensionGate' property for Dimension Probe.

### LINKED ISSUE #363: Dimension Probe crashes when attempting to enter Dimension Gate
The dimension probe does not appear to be able to go through dimension gates; instead it just crashes to the ground.

image
image

emptyVoid:
Resolves #363

### DIFF
--- a/tools/extractors/extract_vehicles.cpp
+++ b/tools/extractors/extract_vehicles.cpp
@@ -293,7 +293,10 @@ void InitialGameStateExtractor::extractVehicles(GameState &state) const
 				};
 
 				vehicle->mapIconType = VehicleType::MapIconType::Arrow;
-				vehicle->canEnterDimensionGate = vehicle->manufacturer.id == "ORG_X-COM";
+				vehicle->canEnterDimensionGate =
+				    vehicle->manufacturer.id == "ORG_X-COM" ||
+				    // for some reason Dimension Probe has Marsec as manufacturer
+				    id == "VEHICLETYPE_DIMENSION_PROBE";
 
 				int image_offset = 0;
 				for (auto &bank : bankings)


JonnyH:
Thanks again for your work :)

It looks like the travis linter (that checks the source code formatting) rejected this - do you run clang-format locally? (we ship a config for that and should be considered the 'canonical coding style definition').

If you are running it, what version of clang-format are you using? We've hit issues in the past where different versions produce slightly different output despite using the same config (which would then get flagged if it didn't match the linter's expectation).

emptyVoid:
No, I personally don't use clang-format, and since [CODE_STYLE.md](https://github.com/OpenApoc/OpenApoc/blob/master/CODE_STYLE.md) says only *highly recommended to run* it I didn't use it before committing the code (it's somewhat annoying turning it on, formatting, and turning it back off in Visual Studio, as it provides no per-project setting for clang-format afaik).

Anyways, I've reformatted the code.

JonnyH:
We probably should change from *highly recommended* as it'll actually cause a CI build failure, and doing everything manually to actually match is a bit of a PITA :)

What IDE/Editors do you use? Many now actually include clang-format support directly, so might be worth setting up so it makes it match on-the-fly?

JonnyH:
I think this should probably be fixed in an xml patch changing the manufacturer of the dimension probe to x-com, as that feels like a mistake in the OG, and more sensible than having an in-code specific workaround

JonnyH:
FYI the "unknown" value here: https://github.com/OpenApoc/OpenApoc/blob/master/tools/extractors/common/vehicle.h#L17 seems to have zero values for all non-dimension capable craft, and non-zero for all the dimension-capable xcom and UFOs - maybe that's related?

`VEHICLETYPE_ALIEN_PROBE: 8
VEHICLETYPE_ALIEN_SCOUT: 2
VEHICLETYPE_ALIEN_TRANSPORTER: 2
VEHICLETYPE_ALIEN_FAST_ATTACK_SHIP: 5
VEHICLETYPE_ALIEN_DESTROYER: 3
VEHICLETYPE_ALIEN_ASSAULT_SHIP: 2
VEHICLETYPE_ALIEN_BOMBER: 3
VEHICLETYPE_ALIEN_ESCORT: 4
VEHICLETYPE_ALIEN_BATTLESHIP: 1
VEHICLETYPE_ALIEN_MOTHERSHIP: 1
VEHICLETYPE_POLICE_HOVERCAR: 0
VEHICLETYPE_AIRTAXI: 0
VEHICLETYPE_RESCUE_TRANSPORT: 0
VEHICLETYPE_CONSTRUCTION_VEHICLE: 0
VEHICLETYPE_AIRTRANS: 0
VEHICLETYPE_SPACE_LINER: 0
VEHICLETYPE_PHOENIX_HOVERCAR: 0
VEHICLETYPE_HOVERBIKE: 0
VEHICLETYPE_VALKYRIE_INTERCEPTOR: 0
VEHICLETYPE_HAWK_AIR_WARRIOR: 0
VEHICLETYPE_DIMENSION_PROBE: 6
VEHICLETYPE_BIOTRANS: 3
VEHICLETYPE_EXPLORER: 4
VEHICLETYPE_RETALIATOR: 5
VEHICLETYPE_ANNIHILATOR: 7
VEHICLETYPE_AUTOTAXI: 0
VEHICLETYPE_AUTOTRANS: 0
VEHICLETYPE_POLICE_CAR: 0
VEHICLETYPE_CIVILIAN_CAR: 0
VEHICLETYPE_STORMDOG: 0
VEHICLETYPE_WOLFHOUND_APC: 0
VEHICLETYPE_BLAZER_TURBO_BIKE: 0
VEHICLETYPE_GRIFFON_AFV: 0
`

JonnyH:
I don't actually know what the values aside from "nonzero" mean here, maybe somehow related to speed? (1=slow, 8=fastest)?

JonnyH:
Thanks for your fix here, but I tried to make a more general version in PR #834 which should solve the same issues (and fixes the 'incorrect' dimension probe manufacturer while I'm at it :)

emptyVoid:
Yeah, I agree, patching sounds more appropriate (I just didn't know the 'base' data is getting patched too).

Though I think the values you've extracted look awfully like an acceleration (when comparing to craft stats from [UFOpaedia.org](https://www.ufopaedia.org/)), human crafts have it at zero since their acceleration is provided by the equipped engine.

emptyVoid:
> What IDE/Editors do you use? Many now actually include clang-format support directly, so might be worth setting up so it makes it match on-the-fly?

Visual Studio 2019
It supports clang-format, though as I mentioned above it has no support for using clang-format on per project basis, and can switch to clang-format globally only.

PR #835, opened 2020-04-24
Make ammo autoreload prefer the same type as previously loaded

JonnyH:
This stops annoying stuff like autocannons switching to HE ammo and
shooting your feet, despite still having a spare AP reload available

### DIFF
--- a/framework/options.cpp
+++ b/framework/options.cpp
@@ -123,6 +123,7 @@ void dumpOptionsToLog()
 	dumpOption(optionSkipTurbo);
 	dumpOption(optionRunAndKneel);
 	dumpOption(optionSeedRng);
+	dumpOption(optionAutoReload);
 
 	dumpOption(optionStunHostileAction);
 	dumpOption(optionRaidHostileAction);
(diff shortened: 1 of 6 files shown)


JonnyH:
I just did a playthrough and noticed a few annoying issues, this being one where one of my agents decided to set everyone on fire as they autoreloaded to IN ammo, then another wounded himself when shooting a brainsucker at his feet with HE.

So instead store off the last loaded ammo type, and prefer to use the same type is available.

Note that going forward we probably should define a preference for autoloaded ammo - as currently it's std::set<StateRef<AEquipmentType>> order, effectively ordered by the operator< of StateRef, so ID string ordering. Which seems a bit arbitrary.

emptyVoid:
I would probably prefer no autoreload to different ammo type at all, just an out-of-ammo notification.

JonnyH:
The OG had some level of autoreload, and for the most part it saves a lot of micromanaging for fast firing/ammo using guns in the realtime mode.

But certainly a 'disable' option would be useful - maybe per-equipped agent or just a global toggle?

emptyVoid:
> The OG had some level of autoreload, and for the most part it saves a lot of micromanaging for fast firing/ammo using guns in the realtime mode.

I agree, autoreload to same ammo is necessary in realtime.

> But certainly a 'disable' option would be useful - maybe per-equipped agent or just a global toggle?

IMO a global toggle would be sufficient.

JonnyH:
Added an option to disable this - though it's default on.

We currently don't have a system to have different settings between turn-based or realtime, so it might be a bit of a pain if you keep switching and want different behavior, but it's still better than nothing

PR #839, opened 2020-04-26
[interface] Fix selecting already-in-progress list item after change selection

### LINKED ISSUE #786: [interface] "Project too large", etc... message dialogues rises twice
When I try to start construction of large item in small workshop the error message dialogue shows twice.

It's non critical interface bug.

Atrosha:
Fixes `researchselection` when already-in-progress list item unable to select after changing selection.

Found this bug and fix it while trying to fix #786

### DIFF
--- a/game/ui/base/researchselect.cpp
+++ b/game/ui/base/researchselect.cpp
@@ -74,36 +74,39 @@ void ResearchSelect::begin()
 		if (topic->current_lab)
 		{
 			LogInfo("Topic already in progress");
-			return;
 		}
-		if (topic->isComplete())
+		else
 		{
-			LogInfo("Topic already complete");
-			auto message_box =
-			    mksp<MessageBox>(tr("PROJECT COMPLETE"), tr("This project is already complete."),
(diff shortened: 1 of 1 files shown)


JonnyH:
Huh, it looks like you've somehow replaced every single file in git and added executable permissions.

What OS are you running? If on linux/macos make sure you didn't "chmod" things.

It might be easier if you just reset the whole repository back to the current OPenApoc 'master', then cherry-pick the specific change you've made (sha 8ec132e by the look of it) on top of that?

Atrosha:
It seems I ruined my fork by commiting to it from NTFS disk earlier.
For now I fetch my fork to Linux home with ext4 partition but problem persist.

JonnyH:
Yeah, ntfs doesn't support unix permissions well, so using ntfs on linux causes issues like this.

If you were running on windows, git then /knows/ unix permissions aren't supported, so ignores it :P

Atrosha:
`git-clang-format` doesn't work properly.

PR #842, opened 2020-04-27
Fix debug unit stunning in presence of retreated units.

emptyVoid:
Fixes a battlescape debug CTD when stunning a unit with [S] is preceded by removing some other unit(s) with [K].


FilmBoy84:
Great work as always @emptyVoid and much appreciated
LGTM, Built, tested and works so merging :)

PR #845, opened 2020-04-29
Fix a CTD while iterating over a changing set.

### LINKED ISSUE #453: Crash on TB mode during Alien/Civilian Movement
Seems to be caused by changing body state
Sound is also rumbling loudly on freeze

image
image

emptyVoid:
Resolves #453.


FilmBoy84:
Another great fix, thanks emptyvoid
Built, tested, working so merging as LGTM

Issue #339, opened 2018-01-10
Destroyed buildings often trap vehicles in their basement causing performance issues and eventual CTD.


FilmBoy84:
In the attached screens, vehicles have been spawned at two destroyed CoS Temples, their attempts to run a mission that cannot be completed (because the path out of the garage is destroyed) slows the game down and fills the log with repeated mission attempts... This may be relevant to issue #247 as well.

Also note, Construction Vehicle 34 is trying to recover a vehicle stuck in a small hangar... If you destroy that vehicle to clear mission, another is spawned (Vehicle 35) and does the same, plus the vehicle sent to destroy Vehicle 34 will not close its mission on returning to base either (Hoverbike 3)...

@redv There appears to be a problem with clearing vehicle missions

![image](https://user-images.githubusercontent.com/33913088/34785441-15e43b94-f629-11e7-8412-3fa32cc05214.png)
![image](https://user-images.githubusercontent.com/33913088/34785670-b643d64e-f629-11e7-9714-ce3536b91deb.png)

![image](https://user-images.githubusercontent.com/33913088/34782625-3b736266-f621-11e7-8689-08c20f01d0da.png)
![image](https://user-images.githubusercontent.com/33913088/34782652-510fdeec-f621-11e7-9cbd-62753148385a.png)
![image](https://user-images.githubusercontent.com/33913088/34782691-63302f96-f621-11e7-8a42-10e5413f59a9.png)

redv:
Could you attach the save?

FilmBoy84:
@redv Here you go :) The Construction vehicle is just to the north of Base 2

[save_Medium Test 1.zip](https://github.com/OpenApoc/OpenApoc/files/1620875/save_Medium.Test.1.zip)

redv:
It's a difficult case in the Megapols "Flayer Two" factory. Inside the factory the CoS's hoverbike 3 tries to get out on the road. But a police car vehicle crashed exactly in front of doors. The construction vehicle 34 tries to rescue the police car but cannot approach.

The hoverbike 3 tries to get out every 30 sec (in-game time). The time out is the same as for ground and for hover vehicles. I.e. for a single lift tube, battle vehicles can lift off every 30 sec to intercept UFOs. So, the time out is OK.

Possible to check: is the vehicle crashed, and if yes, then will be a good idea to increase time out to get out.

The police car should be self-destroyed after 12 hours. Thus the issue will be solved.

[...]

Issue #843, opened 2020-04-28
[Cityscape][Economy] Supplies purchased from orgs/buildings that have been destroyed give the wrong message


FilmBoy84:
Tested under EWM v9 preview 4 by Proselyte Ethan
HOWEVER
You can replicate in normal OpenApoc by doing the following

Destroy all Arms Factories/Flyers for Marsec and Megapol
Ensure you are still allied/friendly to those factions afterwards
Attempt to purchase an item that would normally come from those buildings

Even though the organisations are friendly, the game will prompt that it cannot make the delivery as the organisations are hostile - the wrong message

We need a new prompt to tell the user that the delivery cannot be made as the supplier is rebuilding the facility (we should be able to get the exact text string from OG)

Here's an example save under EWM v9 preview 4
[save_Psionic_Rebalance_Playthrough.zip](https://github.com/OpenApoc/OpenApoc/files/4545306/save_Psionic_Rebalance_Playthrough.zip)

Additional error message
![image](https://user-images.githubusercontent.com/33913088/80481277-0bfdf780-894a-11ea-8c1f-591db655c0de.png)

FilmBoy84:
One of the interesting things about that save is it shows a feature i was not aware of
Because, after a few hours, the order made cannot be delivered it expires and is refunded
We obviously do not want to lose this functionality when fixing the message
![image](https://user-images.githubusercontent.com/33913088/80483786-90527980-894e-11ea-861c-8ac59d47b870.png)

FilmBoy84:
It is unclear why the buildings in that save are collapsed

Ethan reports that he was using the raids collapse buildings function, but had not raided these particular establishments

There are also stuck vehicles at one of the arms factories

So there are a number of issues with this save

That said i have been able to replicate the order issue on Vanilla builds of OpenApoc so it is not exclusive to the EWM

emptyVoid:
> Even though the organisations are friendly, the game will prompt that it cannot make the delivery as the organisations are hostile - the wrong message
> 
> We need a new prompt to tell the user that the delivery cannot be made as the supplier is rebuilding the facility (we should be able to get the exact text string from OG)

I don't see any problem with collapsed building in this save game (aside from trapped ground vehicles). Goods from Marsec get delivered just fine. From Megapol you just have to deliver them with your own craft, since Transtellar is hostile towards Megapol and refuses transportation services for purchases made from Megapol.

PR #847, opened 2020-04-30
Notify road segment of a road tile repair to restore its passability.

emptyVoid:
Related to #339.

### DIFF
--- a/game/state/city/city.cpp
+++ b/game/state/city/city.cpp
@@ -412,6 +412,10 @@ void City::repairScenery(GameState &state)
 			owner->balance -= initialType->value;
 			s->damaged = false;
 			s->type = initialType;
+			if (s->type->tile_type == SceneryTileType::TileType::Road)
+			{
+				notifyRoadChange(s->initialPosition, true);
+			}
 		}
 	}
 	// Step 02: Repair destroyed scenery


FilmBoy84:
Oh, interesting find and thanks for the fix

So that's why, even though the road appears to be there visually, vehicles won't pass on it? - for traversing purposes it's not recognised as passable

Great work emptyvoid

Will build and test shortly and merge as appropriate

emptyVoid:
Yeah, the road segments for collapsed CoS Temple garages kept reporting their road tiles as not intact even after the Temples got reconstructed at the end of day.

FilmBoy84:
Out of interest is this also the potential cause for buildings that are collapsed and are rebuilt still acting as though collapsed in some instances?

I suspect if a bug exists with road, there is likely one similar in building tiles; it would explain some of the odd things in the save in #843 despite that issue dealing with another problem

emptyVoid:
> Out of interest is this also the potential cause for buildings that are collapsed and are rebuilt still acting as though collapsed in some instances?

It looks like the building "parts" are correctly restored during scenery repair at the end of day:
https://github.com/OpenApoc/OpenApoc/blob/5cb327e761953abe6bd1ce0c8f31a23dd38bdb89/game/state/city/scenery.cpp#L1662-L1665

Though I'm not sure what that `commonProperty` is, I believe it shouldn't affect Marsec/Megapol buildings.

FilmBoy84:
Tested, works as intended
LGTM so merging
Great work as always @emptyvoid

PR #853, opened 2020-05-03
Initial infiltration graph implementation

JonnyH:


FilmBoy84:
Every few hours so we can see how OG curves graphs
Day 1 12:00
image
Day 1 15:44
image
Day 1 20:05
image
Day 1 23:55 (just after UFO drop)
image
Day 2 00:02
image
Day 2 04:15
image
Day 2 08:05
image
Day 2 10:05
image
Day 2 14:35
image

Note how organisations have it raise and fall independently of missions (I ran no tac missions in this game, just let it flow naturally)

In the last screen one org has had it's infiltration drop naturally to 0 and has left the list

FilmBoy84:
It appears that the infiltration graph is not just tied to average infiltration level but also where aliens are located in buildings.

As alien agents move around, this affects the average infiltration level, possibly bringing it down to zero

Not all of this is relevant to this PR, but it's something we should consider going forwards when graphing infiltration

Otherwise, LGTM JonnyH

JonnyH:
From my eyes it just looks like the 'current' infiltration is modified live, the rest appears to be a static sample once per day?

Then they're control points for a curve?

My current implementation doesn't curve the results (that can be added later), but pretty much the same behavior.

One thing I'm not sure of is what the 'max' alien infiltration level is - I assumed it was '100', but it seems that once taken over it sets it to '200'? Is this just a "known high", or the actual max?

If 100 isn't the max it should be simple enough to change the constant.

FilmBoy84:
It may be something to do with scale

Remember OG has relations from -100 to 100 in short term, -100 to 100 in long term bias for each organisation in a two way relationship

It would not surprise me if that 200 value range is just an adaption of that scale applied to infiltration mechanics

After-all Apoc supports both hostile, direct, alien takeover and "aliens are friends because they do us less harm than you" loss of support from an org

The point at which an organisation is allied with the aliens will mean that their relationship is 100 (or 200 from the worst possible relationship with the aliens)

PR #854, opened 2020-05-03
Small format change

JonnyH:
I think this may be a difference in clang-format 10 handles it?


JonnyH:
Yes, this seems to be a difference between clang-format-10 and the -8 version currently used on the build machines.

As I don't really want to change the 'standard' clang-format version, parking this for now.

PR #855, opened 2020-05-03
Multiple warning fixes

JonnyH:
This now builds with clang-10 -Wmost without any warnings :)

Most seem to be switch statements not handling all cases, unused variables and constructor order.

This adds some LogError()s at switch cases that didn't seem to make sense, so may suggest a bad state, and a few other fixes (like handling more vehicle mission targets in the strategy view, more vehicle mission types handled in getName(), and what looks like a bug when trying to push a std::set<T> to lua)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ before_install:
 
 
 before_script:
-  - export CFLAGS="-Wall" CXXFLAGS="-Wall -Werror"
+  - export CFLAGS="-Wall" CXXFLAGS="-Wall -Werror -Wno-error=sign-compare"
   - $(which time) cmake . -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DCLANG_TIDY=clang-tidy-8 -DCLANG_FORMAT=clang-format-8 -DENABLE_TESTS=ON -DCMAKE_C_FLAGS="${CFLAGS}" -DCMAKE_CXX_FLAGS="${CXXFLAGS}" -GNinja
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ before_install:
 
 
 before_script:
-  - export CFLAGS="-Wall -Wextra" CXXFLAGS="-Wall -Wextra"
+  - export CFLAGS="-Wall" CXXFLAGS="-Wall -Werror"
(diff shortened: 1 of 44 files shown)


JonnyH:
So the construct used to count the number of iterate-able elements in UString (IE the number of codepoints) breaks if the range for loop argument is [[maybe_unused]] on visual c++ 19.16 and older. 19.20 seems to work. godbolt.org doesn't have any versions between to test.

JonnyH:
e.g. the following breaks in versions before 19.20 (Which is some update for msvc2019 - not sure if 2017 has any working versions)
`#include <array>
std::array<int, 5>  array {0,0,0,0,0};
int count()
{
    int i = 0;
    for ([[maybe_unused]] const auto &c : array)
        i++;
    return i;
}`

gotbolt link:
https://godbolt.org/z/JS-PKc

emptyVoid:
`19.1x` is Visual C++ 2017 (`19.16` - latest)
`19.2x` is Visual C++ 2019 (`19.25` - latest)

Is there any real use in such constructs aside from conditional compilation? (IMO it's an abomination)

Maybe it would be better to replace range-for-loops:
```c++
int i = 0;
for ([[maybe_unused]] const auto &b : state->player_bases)
{
    int cargoDelta = c->getCargoDelta(i);
    if (cargoDelta)
    {
        vecCargoDelta[i] += cargoDelta;
        vecChanged[i] = true;
    }
    i++;
}
```

With classic for-loops:
```c++
int i = 0;
for (size_t i = 0; i < std::size(state->player_bases); i++)
{
    int cargoDelta = c->getCargoDelta(i);
    if (cargoDelta)
    {
        vecCargoDelta[i] += cargoDelta;
        vecChanged[i] = true;
    }
}
```

JonnyH:
While that example has a valid std::size() overload, the UString example does not - as the utf8 the iterator is effectively consuming is a variable length encoding, there's no way of deciding how many codepoints there are without actually scanning the whole string.

But I suppose we could write that manually, instead of using the iterator, but why repeat code?

And as for your example, I agree that's nasty, as it's effectively deriving an order index from something not well ordered (a map keyed by a string ID) - as we're not really worried about performance here, I'd suggest that the 'correct' solution would be to make vecCargoDelta and similar maps keyed on the Base itself (or the ID as writing operator< for a while base seems a bit much :)

emptyVoid:
Shouldn't `std::distance` be sufficient to count the number of code points in UString?

JonnyH:
yeah, that looks pretty much what we want (unless I've screwed up the iterator interface :)

PR #856, opened 2020-05-03
Update appveyor to use msvc2019

JonnyH:
The -dev version (used on my own PR branches) already uses this, and
PR #854 causes an ICE in the 2017 version, so seeing if this helps?


JonnyH:
It seems even the msvc-2019 images on Appveyor don't have the version new enough to avoid the crash - so parking this

PR #861, opened 2020-05-03
Drop carried vehicle when dying.

### LINKED ISSUE #850: ID error
I keep getting a error saying "E void __cdecl OpenApoc::StateRef<class OpenApoc::Vehicle>::resolve(void) const No Vehicle object matching ID "VEHICLE_31" found"

emptyVoid:
Resolves #850.


FilmBoy84:
Thanks for the fix as always emptyVoid
Merging as tested here too

PR #862, opened 2020-05-03
Add support for appending game state from a directory.

### LINKED ISSUE #841: submod loading (gamestate difficulty) issues if using directories instead of archives (Windows x64)
For some reason, if you have the submod portion of the EWM laid out in directories instead of archives, it will CTD on new game start

image
image
image
image

If however, for example, we put difficulty0 into an archive instead of a directory of the same name, it will load correctly with (in the case of the EWM) it loading the new starting funds and other difficulty related items correctly

image

emptyVoid:
Resolves #841.

### DIFF
--- a/framework/fs.h
+++ b/framework/fs.h
@@ -48,6 +48,7 @@ class FileSystem
 	std::list<UString> enumerateDirectory(const UString &path, const UString &extension) const;
 	std::list<UString> enumerateDirectoryRecursive(const UString &path,
 	                                               const UString &extension) const;
+	UString resolvePath(const UString &path) const;
 };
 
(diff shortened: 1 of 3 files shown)


Atrosha:
Please, preserve log warning message, when file can't open.

emptyVoid:
`GameState::loadGame` already generates an error message in case the game state file/directory fails to load.

https://github.com/OpenApoc/OpenApoc/blob/4f5de943e42f9b6698b32181130f5d8bd6fe6df8/game/state/gamestate_serialize.cpp#L464-L468

FilmBoy84:
LGTM I'm good to merge if the other devs are? :)

PR #868, opened 2020-05-12
fix 786 researchselect message dlgs rises twice

### LINKED ISSUE #786: [interface] "Project too large", etc... message dialogues rises twice

Atrosha:
fix #786

### DIFF
--- a/game/ui/base/researchselect.cpp
+++ b/game/ui/base/researchselect.cpp
@@ -65,7 +65,6 @@ void ResearchSelect::begin()
 	this->redrawResearchList();
 
 	auto research_list = form->findControlTyped<ListBox>("LIST");
-	research_list->AlwaysEmitSelectionEvents = true;
 
 	research_list->addCallback(FormEventType::ListBoxChangeSelected, [this](FormsEvent *e) {
 		LogInfo("Research selection change");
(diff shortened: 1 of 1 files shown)


Atrosha:
It seems `AlwaysEmitSelectionEvents` in

https://github.com/OpenApoc/OpenApoc/blob/ff1db57d65abdb1ee0d4c5c2051801da254104bb/forms/listbox.cpp#L184-L195

isn't properly implemented.

Atrosha:
And now this dialogues doesn't shows up after clicking again before changing selection, but this is much better then was previously.

FilmBoy84:
Tested and appears to work, though with the caveat Atrosha mentions in #870 
I'm unsure whether to merge with the rest of the code still being a mess though?

game/ui/base/researchselect.cpp
```diff
@@ -80,31 +79,32 @@ void ResearchSelect::begin()
 			if (topic->isComplete())
 			{
 				LogInfo("Topic already complete");
-				auto message_box = mksp<MessageBox>(tr("PROJECT COMPLETE"),
-				                                    tr("This project is already complete."),
-				                                    MessageBox::ButtonOptions::Ok);
-				fw().stageQueueCommand({StageCmd::Command::PUSH, message_box});
+				fw().stageQueueCommand({StageCmd::Command::PUSH,
```
JonnyH, line 82: To my eye these blocks look functionally identical?

Am I missing a change?

If just a cleanup I don't know if it's necessary, or at least should be a separate commit to functional changes.

JonnyH:
Yeah, this all looks a bit spidery - I really hate magic undocumented bool flags changing behavior of things....

Might be worth digging down and seeing how much of this complexity is necessary?

Atrosha:
I just remove `AlwaysEmitSelectionEvents = true` and container variables for dialogues.

game/ui/base/researchselect.cpp
Atrosha: I just remove AlwaysEmitSelectionEvents = true and container variables for dialogues.

Atrosha:
@emptyVoid found better solution for this. Closing.

PR #869, opened 2020-05-13
Import dependencies from VCPKG by default on Windows.

### LINKED ISSUE #866: Are dependencies required to be built from sources?
From 8 dependency libraries found in `dependencies` 6 are available through `vcpkg` (except `libsmacker` and `tinyformat`).

Does OpenApoc use some custom/specific versions of those libraries and thus cannot import them from `vcpkg`?

If not I'd like to switch to `vcpkg` to reduce compilation times.

emptyVoid:
Here's a proposal to import uncustomised dependency libraries from VCPKG instead of building them from sources. Importing is enabled by default on Windows and could be enabled via `USE_VCPKG_*` options on other platforms, the same options could be used to disable importing on Windows.

Resolves #866.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,12 @@ option(MSVC_PDB "Always generate PDB files" OFF)
 option(ENABLE_TESTS "Build some unit tests" ON)
 option(EXTRACT_DATA "Run the data extractor as part of the default target" ON)
 
+option(USE_VCPKG_GLM "Use GLM library from VCPKG toolchain" ${WIN32})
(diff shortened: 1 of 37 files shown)


JonnyH:
I'm unsure if this is actually needed going forward - we need the sources for other platforms already, and I personally like having sources locally to aid debugging into libraries - maybe there's something wrong with my setup but it doesn't seem to do source-level debugging with vcpkg-imported libraries.

emptyVoid:
Have you tried updating `vcpkg`?
I've just tested stepping into `pugi::xml_document::load_file` -- works as expected.

PR #875, opened 2020-05-19
Fix a CTD on opening the tile view while a vehicle returns from UFO recovery.

emptyVoid:
After a UFO battle ends XCOM vehicle gets assigned an `Offer Service` mission without a target building, resulting in a CTD if player switches to the city tile view.

https://github.com/OpenApoc/OpenApoc/blob/e65b5fe4a1d997ed1d8bcbd3c2e313ae4307ab74/game/state/battle/battle.cpp#L3362


FilmBoy84:
Tested and working as intended
Thanks for the fix emptyVoid

PR #877, opened 2020-05-20
Vaporize popper's body on "self-destruct".

### LINKED ISSUE #745: [Battlescape] Poppers that explode should NOT leave behind a recoverable body
When a popper is hit by ammunition that does cause it to explode, or detonates it's own bomb, it should NOT leave behind it's body for recovery

emptyVoid:
Removes popper's body from the battlescape if it dies violently (i.e. self-destructing), and prevents X-COM from recovering it.

Resolves #745.


FilmBoy84:
Thanks for the fix emptyVoid

PR #879, opened 2020-05-20
Make toxigun, plasma, and laser damage non-violent.

### LINKED ISSUE #744: [Battlescape] Poppers hit by toxiguns/plasma/laser/disruptor/devastator should die, leaving their body, but not explode
When a popper is hit by any of the three toxigun munitions/plasma/laser/disruptor/devastator, it should die without exploding AND leave behind a recoverable body

Currently poppers explode and leave behind their body

Additionally, when a popper is hit by ammunition that does cause it to explode, it should NOT leave behind it's body (I shall make a separate issue for this)

emptyVoid:
Adjusts toxigun, plasma, and laser damage type not to detonate poppers.

Resolves #744.

### DIFF
--- a/tools/extractors/extract_agent_equipment.cpp
+++ b/tools/extractors/extract_agent_equipment.cpp
@@ -44,6 +44,11 @@
 #define DT_EXPLOSIVE 4
 #define DT_STUNGUN 5
 #define DT_PSIBLAST 6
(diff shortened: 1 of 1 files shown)


FilmBoy84:
Thanks for the fix emptyVoid

Issue #773, opened 2019-09-16
Replace tinyfmt with fmt (https://github.com/fmtlib/fmt)


JonnyH:
fmt:
- Allows positional arguments to ease translation (e.g. the string format("Do {0} to {1}", thing1, thing2) could be "translated" as "Do {1} to {0}", to allow different ordering in the sentance
- Currently in the draft c++20 standard, so pretty likely to be in standard libs soon (fewer dependencies and wider use and programmer exposure is good)
- Faster compile times
- Fast run times

compared to tinyformat.

emptyVoid:
Is this being worked on?

JonnyH:
I've got the initial port working locally, but haven't had time to polish it and properly test (e.g. all the translation strings need to be updated, so I was using this as a good point to make the strings extracted automatically from code with xgettext as a build target etc)

It also revealed a number of "bad" translation policies - e.g. stitching together translated text fragments instead of doing the wrong string at a time (e.g. something like "format(tr("%s, %s"), format(tr("%s thing is %d out of "), object, count), format(tr(out of "%d total"), number));

Which would result in the completely useless strings for translators:
"%s %s"
"%s thing is %d "
"out of %d total"

when it really should be translated as a single call, so the extracted string should be:
"%s thing is %d out of %d total" (or "{0} thing is {1} out of {2} total" in the fmtlib way of doing things)

emptyVoid:
From what I can gather there has to be two distinct formatting subsystems:
1. `{fmt}` based logging (since IMO there's little benefit in translating potentially ever-changing log messages, and would be a bit hit to performance routing them via `boost::locale`.
2. `boost::locale::format` for localizable UI strings (since it has support for plural forms).

I would propose implementing them in small steps: first switch to `{fmt}` then deal with localizable strings.

JonnyH:
I am unsure if boost::locale is actually going to be that useful for us going forward - it's great for static applications with static translation databases, but I have struggled to find where I can hook into it's message database to add in new messages from mods.

I think it's only possible to add a completely new database in a new translation domain, and then we get into issues trying to figure out which domain we should translate a string with, without doing something like trying /all/ mod domains in order until we find a match

JonnyH:
So my current plan is to try to separate each of the tasks - similar to what you mentioned:

- Use {fmt} for all logs and internal strings (it's used in some places for constructing IDs etc - something we explicitly don't want translating)
[...]

PR #881, opened 2020-05-25
Wip/fmt translate redux

JonnyH:
A work-in-progress at moving from a single tr(string) call to boost::locale's message formatting

RE the boost-locale discussion in issue #773 

This means that strings can be translated at as a while rather than piecemeal depending on how they're constructed in code.

Going forward it's always best to construct strings for display in sensible units, to allow translators some sanity.

e.g. if you want the string:
"Moving item {1} from {2} to {3} will take {4} hours"
doing that all in one tformat() call like this:
tformat("Moving item {1} from {2} to {3} will take {4} hours") % %item_name % source % destination % time;
will allow the string to be translated as a whole, possible reordered etc.

But in quite a few places, stings were constructed piecemeal, e.g:
string = "Moving Item "
string += item_name
string += "from"
etc. etc.

That means that even if each string section is wrapped in a translate-aware replacement call (e.g. tr(), now replaced by translate()) the translators only see the strings "Moving Item" and "from", can't reorder them, and any instance of "from" will be replaced by the translated version - even if it doesn't make sense in context.

### DIFF
--- a/forms/control.cpp
+++ b/forms/control.cpp
@@ -13,6 +13,7 @@
 #include "framework/sound.h"
 #include "framework/trace.h"
 #include "library/sp.h"
+#include "library/strings_translate.h"
 
 namespace OpenApoc
 {
@@ -730,7 +731,7 @@ void Control::configureSelfFromXml(pugi::xml_node *node)
 			{
 				ToolTipText = child.attribute("text").as_string();
 				if (!ToolTipText.empty())
-					ToolTipText = tr(ToolTipText);
+					ToolTipText = translate(ToolTipText);
 			}
 			else
 			{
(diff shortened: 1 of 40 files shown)


JonnyH:
Also going through this I noticed a large number of places where we translate /names/ just before they're displayed onscreen.

This is wrong, as if (for example) you named something to match a string that happened to be translated, it'll be translated too at display time - even if that's not what you changed the name to.

That likely means we shouldn't actually translate names of objects (buildings, equipment, organisations etc.) but instead just set their initial values at game start time based on language (maybe using the current modloading framework?)

JonnyH:
Also using boost-locale in a more direct manner shown here means we can use it's extended formatting stuff to allow things like plurals, currency, direct date translations etc.  - which we should probably be using.

And there's a number of places where table is implemented effectively as separate text boxes - which again limits the translations of their corresponding contents - can't reorder or resize columns based on content or translate any way but each column as independent text - maybe that's worth trying to implement some kind of tabulation-aware text box? Then columns can be split by a special character in translations (e.g. \t) instead of in-code

PR #883, opened 2020-05-26
Add VS launch settings for DataExtractor.

emptyVoid:
Adds default settings for DataExtractor utility to launch and debug it in Visual Studio.


JonnyH:
I don't use VS, so I'll just assume this works :)

PR #885, opened 2020-05-26
Allow large vehicles moving into tiles occupied by themselves.

### LINKED ISSUE #629: Alien Battleship Stuck
5 UFOs entered the map, however, the Battleship seemed to be stuck on the gate? I was able to knock it down with a debug command, but even destroying the building next to it didn't free it. Attacking it with craft did not help, either. The other UFOs went to it and just hovered next to it until I attacked.
[save_Battleship Stuck.zip](https://github.com/OpenApoc/OpenApoc/files/3327847/save_Battleship.Stuck.zip)

emptyVoid:
Resolves #629.

### DIFF
--- a/game/state/city/vehiclemission.cpp
+++ b/game/state/city/vehiclemission.cpp
@@ -57,30 +57,18 @@ static const std::map<UString, std::pair<unsigned, unsigned>> selfDestructTimer 
     {"VEHICLETYPE_ALIEN_MOTHERSHIP", {60 / 5 * TICKS_PER_MINUTE, 240 / 5 * TICKS_PER_MINUTE}}};
 } // namespace
 
-FlyingVehicleTileHelper::FlyingVehicleTileHelper(TileMap &map, Vehicle &v)
-    : FlyingVehicleTileHelper(map, *v.type, v.crashed, (int)v.altitude)
-{
-}
-
-FlyingVehicleTileHelper::FlyingVehicleTileHelper(TileMap &map, VehicleType &vehType, bool crashed,
-                                                 int altitude)
-    : map(map), crashed(crashed), altitude(altitude)
+FlyingVehicleTileHelper::FlyingVehicleTileHelper(TileMap &map, const Vehicle &v)
+    : map(map), v(v), altitude((int)v.altitude)
 {
 	int xMax = 0;
 	int yMax = 0;
-	for (auto &s : vehType.size)
+	for (const auto &s : v.type->size)
 	{
 		xMax = std::max(xMax, s.second.x);
 		yMax = std::max(yMax, s.second.y);
 	}
 	size = {xMax, yMax};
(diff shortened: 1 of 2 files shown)


game/state/city/vehiclemission.cpp
```diff
@@ -58,17 +58,11 @@ static const std::map<UString, std::pair<unsigned, unsigned>> selfDestructTimer
 } // namespace
 
 FlyingVehicleTileHelper::FlyingVehicleTileHelper(TileMap &map, Vehicle &v)
```
JonnyH, line 60: Can all these args be const? And presumably the members variablestoo

game/state/city/vehiclemission.cpp
```diff
@@ -58,17 +58,11 @@ static const std::map<UString, std::pair<unsigned, unsigned>> selfDestructTimer
...
-
-FlyingVehicleTileHelper::FlyingVehicleTileHelper(TileMap &map, VehicleType &vehType, bool crashed,
-                                                 int altitude)
-    : map(map), crashed(crashed), altitude(altitude)
+    : map(map), v(v), altitude((int)v.altitude)
 {
 	int xMax = 0;
 	int yMax = 0;
-	for (auto &s : vehType.size)
+	for (auto &s : v.type->size)
```
JonnyH, line 65: 's' can be const?

game/state/city/vehiclemission.cpp
```diff
@@ -132,14 +120,20 @@ bool FlyingVehicleTileHelper::canEnterTile(Tile *from, Tile *to, bool, bool &, f
 	// Allow pathing into tiles with crashes
 	bool foundCrash = false;
 	bool foundScenery = false;
+	auto self = v.shared_from_this();
```
JonnyH, line 123: 'self' can be const?

game/state/city/vehiclemission.cpp
emptyVoid: Sure.

game/state/city/vehiclemission.cpp
emptyVoid: Done, although I'm not a fan of making every unchanging local non-reference variable a const.

game/state/city/vehiclemission.cpp
emptyVoid: Can't make `TileMap` reference a const, since it drags too much refactoring in places unrelated to the PR (in `CanEnterTileHelper` and all of its descendants, in `Tile`, etc.).

I guess I will try argument constness refactoring through out the code in another PR.

JonnyH:
I generally like const-by-default, as especially in larger functions it means you know any subsequent references to that variable will be that known value - and it's not modified in the block in between.

Less important for smallish functions where you can kinda keep all the state in mind at one time, but IMHO a good habit nonetheless.

PR #886, opened 2020-05-26
Fix dumping of floating point options to log.

emptyVoid:
Fixes log strings for options of floating point type.


JonnyH:
Thanks - looks like a dumb typo and poor testing :)

PR #889, opened 2020-05-28
Destroy falling UFOs on touchdown in Alien Dimension.

### LINKED ISSUE #682: [Alien Dimension] UFOs do not search for crashsite and explode. X-COM Craft become stuck.
Raided alien zone and downed UFO didn't explode, so I tried investigating one with a crewed annihilator, but it then froze and wouldn't move, even when the other ships returned to the portal.
[save_Stuck in zone.zip](https://github.com/OpenApoc/OpenApoc/files/3473746/save_Stuck.in.zone.zip)

emptyVoid:
Falling UFOs no longer survive the crash in Alien Dimension.

Resolves #682.

### DIFF
--- a/game/state/city/vehiclemission.cpp
+++ b/game/state/city/vehiclemission.cpp
@@ -1405,8 +1405,15 @@ void VehicleMission::update(GameState &state, Vehicle &v, unsigned int ticks, bo
 		{
 			if (finished)
 			{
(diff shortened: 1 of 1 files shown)


FilmBoy84:
Thanks for this emptyVoid
LGTM, merging

PR #878, opened 2020-05-20
Remove all dead vehicles immediately after a battle.

emptyVoid:
Resolves an issue with "resurrecting" vehicles when loading a save game made right after a battle (e.g. UFO recovery).

The issue manifests if player saves the game right after a UFO recovery battle concludes and before unpausing the cityscape, and loads this save file -- the recovered UFO gets "resurrected".

Here's a repro save game -- unpause and wait for the battle to end, save the game in cityscape and load the saev:
[save_issue-recovered-ufo.zip](https://github.com/OpenApoc/OpenApoc/files/4656524/save_issue-recovered-ufo.zip)

Issue #788, opened 2019-10-25
Crash #333


Slaine050877:
This Constantly Happens - Its after a Purple UFO is shot down.

W class OpenApoc::SoundBackend *__cdecl `anonymous-namespace'::SDLRawBackendFactory::create(void) Creating SDLRaw sound backend (Might have issues!)
W __cdecl `anonymous-namespace'::SDLRawBackend::SDLRawBackend(void) Changing audio drivers is not currently implemented!
W __cdecl `anonymous-namespace'::SDLRawBackend::SDLRawBackend(void) Selecting audio devices not currently implemented! Selecting first available device.
W __cdecl `anonymous-namespace'::SDLRawBackend::SDLRawBackend(void) Audio output format: Channels 2, format: signed float LE 32bit, freq 22050, samples 492
W void __cdecl OpenApoc::Framework::setupModDataPaths(void) loading mod "base"
W void __cdecl OpenApoc::Framework::setupModDataPaths(void) Loaded modinfo for mod ID "org.openapoc.base"
W void __cdecl OpenApoc::Framework::setupModDataPaths(void) Appending data path "./data/mods/base/data"
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 2 entries from 140 to 141
W void __cdecl OpenApoc::GameState::validate(void) Validating GameState
W void __cdecl OpenApoc::GameState::validate(void) Validated GameState
E class std::shared_ptr<class OpenApoc::Vehicle> __cdecl OpenApoc::Vehicle::get(const class OpenApoc::GameState &,const class OpenApoc::UString &) No vehicle matching ID "VEHICLE_333"
E void __cdecl OpenApoc::StateRef<class OpenApoc::Vehicle>::resolve(void) const No Vehicle object matching ID "VEHICLE_333" found
E class std::shared_ptr<class OpenApoc::Vehicle> __cdecl OpenApoc::Vehicle::get(const class OpenApoc::GameState &,const class OpenApoc::UString &) No vehicle matching ID "VEHICLE_333"
E void __cdecl OpenApoc::StateRef<class OpenApoc::Vehicle>::resolve(void) const No Vehicle object matching ID "VEHICLE_333" found
E class std::shared_ptr<class OpenApoc::Vehicle> __cdecl OpenApoc::Vehicle::get(const class OpenApoc::GameState &,const class OpenApoc::UString &) No vehicle matching ID "VEHICLE_333"
[...]

PR #890, opened 2020-05-29
Remove dead vehicles from selection.

emptyVoid:
Resolves a CTD on `CityView::render` when the list of selected vehicles contains a non-present entry.

Although the issue was uncovered as a result of #878, it manifests under different circumstances as well; here's a repro:
1. Fly to Alien Dimension.
2. Order XCOM units to return to Human Dimension.
3. Select any UFO and leave it selected till cityscape changes to Human Dimension.
4. Wait for the selected UFO to invade Human Dimension.
5. Destroy it.
6. Save the game.
7. Reload the game.
8. Fly to Alien Dimension.

Related to https://github.com/OpenApoc/OpenApoc/issues/788#issuecomment-634602993.

### DIFF
--- a/game/state/gamestate.cpp
+++ b/game/state/gamestate.cpp
@@ -942,6 +942,23 @@ void OpenApoc::GameState::cleanUpDeathNote()
 		for (auto &name : this->vehiclesDeathNote)
 		{
 			vehicles.erase(name);
+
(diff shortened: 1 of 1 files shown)


FilmBoy84:
Thanks for fix emptyVoid
LGTM Merging

PR #894, opened 2020-06-01
Support keypad enter button.

### LINKED ISSUE #891: Enter button at right numpad dont work when i need press ok to close screen
Enter button at right numpad dont work when i need press ok to close screen
or press enter two times to new save and close screen

usually i use this right enter to faster close screen and its work in OG

emptyVoid:
Resolves #891.


FilmBoy84:
Thanks for fix emptyVoid
LGTM Merging

PR #901, opened 2020-06-20
c++20 preparation workarounds

JonnyH:
This allows us to build with c++20 enabled - the biggest changes are the introduction of char8_t for utf8 character streams, and char32_t and similar no longer being implicitly cast-able to other integer types (As they're meant to be utf32 codepoints, not "numbers")

Note this is a /workaround/ - a long-term solution would be to replace the entire UString type with u8string (as they effectively mean the same thing - "utf8-encoded bytes" - and move the helpers we use to free functions (if they can't be replaced by standard library functions).

Unfortunately, pre-c++20 didn't define all the types (e.g. char8_t being distinct from 'char'), so until we move completely to only even newer c++ versions we probably need some level of hackery.

### DIFF
--- a/framework/font.cpp
+++ b/framework/font.cpp
@@ -67,7 +67,7 @@ sp<PaletteImage> BitmapFont::getGlyph(UniChar codepoint)
 		// FIXME: Hack - assume all missing glyphs are spaces
 		// TODO: Fallback fonts?
 		LogWarning("Font %s missing glyph for character \"%s\" (codepoint %u)", this->getName(),
-		           UString(codepoint), codepoint);
+		           UString(codepoint), static_cast<uint32_t>(codepoint));
 		auto missingGlyph = this->getGlyph(UString::u8Char(' '));
 		fontbitmaps.emplace(codepoint, missingGlyph);
 	}
--- a/framework/modinfo.cpp
+++ b/framework/modinfo.cpp
@@ -53,7 +53,7 @@ std::optional<ModInfo> ModInfo::getInfo(const UString &path)
 	{
 		for (const auto node : requiresNode.children("entry"))
 		{
-			info.requires().push_back(node.value());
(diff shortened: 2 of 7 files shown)


Atrosha:
Builded and tested with `clang 10.0.0`

Atrosha:
Sorry, was too fast.
When I set c++20 compiler options I get an errors on building with `clang`

```
/home/atrosha/OpenApoc/game/state/tilemap/collision.cpp:49:19: error: use of overloaded operator '!=' is ambiguous (with operand types 'OpenApoc::LineSegmentIterator<int, true>' and 'OpenApoc::LineSegmentIterator<int, true>')
        for (auto &point : line)
                         ^
/home/atrosha/OpenApoc/library/line.h:131:7: note: candidate function
        bool operator!=(const LineSegmentIterator &other) { return !(*this == other); }
             ^
/home/atrosha/OpenApoc/library/line.h:126:7: note: candidate function
        bool operator==(const LineSegmentIterator &other)
             ^
/home/atrosha/OpenApoc/library/line.h:126:7: note: candidate function (with reversed parameter order)
/home/atrosha/OpenApoc/game/state/tilemap/collision.cpp:49:21: note: in implicit call to 'operator!=' for iterator of type 'OpenApoc::LineSegment<int, true>'
        for (auto &point : line)
                           ^~~~
/home/atrosha/OpenApoc/library/line.h:27:39: note: selected 'begin' function with iterator type 'LineSegmentIterator<int, true>'
        LineSegmentIterator<T, conservative> begin() const;
                                             ^
1 warning and 1 error generated.
make[2]: *** [game/state/CMakeFiles/OpenApoc_GameState.dir/build.make:970: game/state/CMakeFiles/OpenApoc_GameState.dir/tilemap/collision.cpp.o] Error 1
```

PR #904, opened 2020-06-23
Show total game days in save game menus.

emptyVoid:
Changes how save games are presented in the menus altering Days field to show total days since 2084-Mar-7 instead of day of month.


FilmBoy84:
Thanks for fix emptyVoid
LGTM Merging

PR #905, opened 2020-06-23
Clear background of an unselected form list item in debug form preview.

emptyVoid:


FilmBoy84:
Thanks for fix emptyVoid
LGTM Merging

Issue #906, opened 2020-07-03
Icons of UFOs on map not should be circle red


makus82:
![image](https://user-images.githubusercontent.com/11581207/86483052-e0d5c100-bd5b-11ea-9462-8f0c5e4c8f4c.png)

Icons of UFOs on map not should be circle red but red square romb
![image](https://user-images.githubusercontent.com/11581207/86483135-0cf14200-bd5c-11ea-98f1-311e43fd1eef.png)

FilmBoy84:
If i recall this is size related
So smaller UFOs use 
![image](https://user-images.githubusercontent.com/33913088/86535746-8747d080-beda-11ea-9901-ecf7d134a8a9.png)
Whereas larger UFOs use a red circle

makus82:
nop in OG all ufo use https://user-images.githubusercontent.com/33913088/86535746-8747d080-beda-11ea-9901-ecf7d134a8a9.png
only really big use Big circle but not small circle

makus82:
> If i recall this is size related
> So smaller UFOs use
> ![image](https://user-images.githubusercontent.com/33913088/86535746-8747d080-beda-11ea-9901-ecf7d134a8a9.png)
> Whereas larger UFOs use a red circle

nop in OG all ufo use 
 ![image](https://user-images.githubusercontent.com/33913088/86535746-8747d080-beda-11ea-9901-ecf7d134a8a9.png)
only really big use Big circle but not small circle

FilmBoy84:
That's what i thought; Battleship and Mothership use big red circles
All other craft are only one tile in size so use the single tile
![image](https://user-images.githubusercontent.com/33913088/86537875-232d0880-beea-11ea-82e5-cd49aca1c492.png)

TBH what we should have is a large, four tile version of the red arrow for big UFOs and keep all red circles for ground vehicles only

I can only assume this was not done as a result of time constraints

makus82:
im for left all as is) in OG

FilmBoy84:
The big problem with the OG is that you cannot see which direction Large UFOs are heading from the map
It may be worth trying to fix this with a new icon...

idshibanov:
Large UFO icon is a thing of preference. Personally, I like the big menacing circle, it fits.

FilmBoy84:
[Directional Large Flyer Map Icons.zip](https://github.com/OpenApoc/OpenApoc/files/4966760/Directional.Large.Flyer.Map.Icons.zip)

Preview
[...]

PR #912, opened 2020-07-23
Change small UFO icon

idshibanov:
Related to #906 .

Change is in data extractor so would have to re-run it to see the difference.


FilmBoy84:
LGTM, merging

PR #916, opened 2020-07-28
Implement organization and city economy

idshibanov:
Fully implemented city economy along with workers, wages and budgets. Based on actual data and algorithms in the original game. X-Com funding will be implemented separately as they use completely different logic there.

1. Added code extracting missing values related to Building, BuildingFunction and Organization
2. New game initialization logic: two weeks of funding and profitability of every building
3. Simulated civilian workforce that is paid salary by organizations. Civilians move between buildings and can switch jobs.


FilmBoy84:
LGTM
I also note your comment with regards to values from bld_cost_struc
These can always be amended later if needed
But, TBH, they seem to work in the build i tested as per OG at Medium difficulty
Maybe the offset was determined by difficulty at one time?

If JonnyH and any other devs don't have any queries, i'm good to merge

Thanks for a great update ShellStorm

game/state/gamestate.cpp
```diff
@@ -1215,7 +1229,141 @@ void GameState::updateEndOfDay()
...
+
+// Recalculates AI organization and civilian finances, updating budgets and salaries
+void GameState::updateHumanEconomy()
+{
+	// TODO: remove hardcoded references
+	auto humanCity = cities["CITYMAP_HUMAN"];
+
+	humanCity->populationWorking = 0;
+	// Game resets only Government income, it's not right logically but will keep it to match OG
+	organisations["ORG_GOVERNMENT"]->income = 0;
```
idshibanov, line 1247: I dislike using hardcoded string IDs for this kind of thing but can't think of other way right now. Any suggestions to how to do this better with current code?

idshibanov:
> 
> I also note your comment with regards to values from bld_cost_struc
> These can always be amended later if needed
> But, TBH, they seem to work in the build i tested as per OG at Medium difficulty
> Maybe the offset was determined by difficulty at one time?

We should be able to update those values via difficulty patch XML. I don't think they work that well in OG (Marsec gets income boost in base data for example), so we can look to change them later.

game/state/city/building.cpp
```diff
@@ -96,6 +97,48 @@ bool Building::hasAliens() const
...
+	return currentWorkforce * (incomePerCapita - currentWage) - maintenanceCosts;
+}
+
+unsigned Building::countActiveTiles() const
+{
+	unsigned relevantTiles = 0;
+	for (auto &p : buildingParts)
+	{
+		auto tile = city->map->getTile(p);
+		if (tile->presentScenery && tile->presentScenery->type->value > 5)
```
JonnyH, line 134: Not a big fan of the magic number "5" - if there's a limited set of types it makes sense to have a static set of values, or even modifying the scenerytype to have a flag that says "This is part of a building"

game/state/gamestate.cpp
JonnyH: Maybe add to the GameState the IDs of 'Interesting' organisations? (IE "player = ORG_X-COM", "enemy=ORG_ALIENS", player_funder="ORG_GOVERNMENT")

Then people can change that in mods - but it's a little out of scope here - put it down as a TODO, as there's hundreds of other hardcoded ORG_ALIENs/ORG_WHATEVER around the place :)

tools/extractors/extractors.cpp
```diff
@@ -314,6 +314,27 @@ const std::vector<int> InitialGameStateExtractor::buildingFunctionDetectionWeigh
 	100, 100, 100, 100, 100, 100, 100, 100, 100
 };
 
+// Data is pulled from UFO2P, bld_cost_struc (was hard to determine common offset)
```
JonnyH, line 317: for the other data tables we actually map in the .exe and read it directly in the extract step - can that be done here?

game/state/city/building.cpp
idshibanov: What about making it `constexpr`? Or is there a config file we can put this kind of variables into?

game/state/gamestate.cpp
idshibanov: Yes, I have a TODO with this in the beginning of this method.

game/state/city/building.cpp
JonnyH: Stuff like that should probably go in the GameState - as if mods add new scenery types they probably also want to alter this value.

But I think it may be better to add a new bool flag to the SceneryType object itself - and use that to key off. For the initial data files, you can add something to the extractor to set the flag if the type.value > 5 - as we 'know' the OG data files use that value, but any future mods that add new/modify scenery types can set that flag themselves as necessary.

tools/extractors/extractors.cpp
idshibanov: I couldn't figure out exact structure and offset. We'll see if @Skin36 can find it, already asked for help :)

game/state/city/building.cpp
idshibanov: I like this approach! Will add the boolean.

game/state/city/building.cpp
```diff
@@ -36,7 +37,7 @@ template <> const UString &StateObject<BuildingFunction>::getPrefix()
 }
 template <> const UString &StateObject<BuildingFunction>::getTypeName()
 {
-	static UString name = "AgentType";
+	static UString name = "BuildingFunction";
```
JonnyH, line 40: this should be const?

game/state/city/building.cpp
```diff
@@ -96,6 +97,48 @@ bool Building::hasAliens() const
...
+{
+	return currentWorkforce * (incomePerCapita - currentWage) - maintenanceCosts;
+}
+
+unsigned Building::countActiveTiles() const
+{
+	unsigned relevantTiles = 0;
+	for (auto &p : buildingParts)
+	{
+		auto tile = city->map->getTile(p);
```
JonnyH, line 133: This can be const?

game/state/city/building.cpp
```diff
@@ -96,6 +97,48 @@ bool Building::hasAliens() const
...
+
+int Building::calculateIncome() const
+{
+	return currentWorkforce * (incomePerCapita - currentWage) - maintenanceCosts;
+}
+
+unsigned Building::countActiveTiles() const
+{
+	unsigned relevantTiles = 0;
+	for (auto &p : buildingParts)
```
JonnyH, line 131: This can be const?

game/state/gamestate.cpp
```diff
@@ -1178,7 +1192,141 @@ void GameState::updateEndOfDay()
...
+	}
+	humanCity->averageWage =
+	    (humanCity->populationWorking) ? totalCivilianIncome / humanCity->populationWorking : 0;
+
+	// Step 2. Government additionally gets 10% of civilian income as taxes
+	government->balance += totalCivilianIncome / 10;
+
+	// Step 3. Calculate civilians leaving work because of the low wage
+	const int minimumWage = std::max(humanCity->averageWage, 30);
+	for (auto &pair : humanCity->buildings)
```
JonnyH, line 1239: Can you use a structured binding? (IE const auto& [id, build] : humanCity->buildings)?

game/state/gamestate.cpp
```diff
@@ -1178,7 +1192,141 @@ void GameState::updateEndOfDay()
...
+					humanCity->populationUnemployed -= workersJoining;
+				}
+			}
+		}
+		// Reduce demands by 10%
+		expectedWage -= humanCity->averageWage / 10;
+	}
+
+	// Step 5. Adjust the building wages to attract new workers
+	for (auto &pair : humanCity->buildings)
```
JonnyH, line 1299: Again structured binding?

game/state/gamestate.cpp
```diff
@@ -1178,7 +1192,141 @@ void GameState::updateEndOfDay()
...
+		if (humanCity->populationUnemployed <= 0)
+			break;
+
+		for (auto &pair : humanCity->buildings)
+		{
+			const auto build = pair.second;
+			if (build->currentWage > expectedWage)
+			{
+				int workersJoining = humanCity->populationUnemployed;
+				if (build->currentWage < 20)
```
JonnyH, line 1267: Generally I dislike the magic numbers - long-term it makes sense to push this into the GameState to allow mods to change the behavior?

tools/extractors/common/building.h
```diff
@@ -5,23 +5,58 @@
...
+	uint16_t income;
+	uint16_t agentSpawnType;
+	uint16_t investmentValue;
+	uint16_t respectValue;
+};
+#pragma pack(pop)
+
+static_assert(sizeof(struct BuildingCostData) == 12, "Unexpected bld_cost_struc size");
+
+#define BUILDING_COST_STRUCT_OFFSET_START 0x1405C0
```
JonnyH, line 57: Thanks for this :) - I know it's more work, but it acts as effective documentation on the binary too

tools/extractors/extract_buildings.cpp
```diff
@@ -21,6 +22,17 @@ void InitialGameStateExtractor::extractBuildingFunctions(GameState &state) const
 	{
 		auto f = mksp<BuildingFunction>();
 		f->name = data.building_functions->get(i);
+		if (i < data.building_cost_data->count())
+		{
+			auto costEntry = data.building_cost_data->get(i);
```
JonnyH, line 27: const?

game/state/city/building.cpp
```diff
@@ -96,6 +97,48 @@ bool Building::hasAliens() const
 	return false;
 }
 
+void Building::initBuilding(GameState &state)
+{
+	// Initialize economy data, done in the map/city editor or when game starts for the first time
+	// Not on save/load, that's why values are serialized
+	currentWage = 65;
```
JonnyH, line 104: Is this the default initial state? Maybe add it to the extractor instead (IE so it'll be in the 'initial gamestate' - as otherwise this in-code value will override anything set in mods

idshibanov: Yes, it was the default internally. I moved it to be a city variable, set in the extractor (only for human map). The rest of "magic numbers" is a rough proportion of this one, so I got rid of those.

FilmBoy84:
Excellent work Shellstorm!
My build from this branch seems to work with no obvious problems well into week 6 for organisation funding
I'm good to merge and JonnyH gave thumbsup to merge on Discord #GitHub_News last night
So here we go.... 🥳

makus82:
Wow im reading your description and cant belive my eyes) its looking really great 
thx for your work @idshibanov 

_X-Com funding will be implemented separately as they use completely different logic there._
From this i understand you will do this after the economy?

Anyway this your work and Jonny with language changes should be now go to OA news at FB =] right away!

Issue #922, opened 2020-08-01
CTD on the Hire/Fire Staff button


BeornTB:
2 logs rather than one:

[HireFireStaffCTD1.zip](https://github.com/OpenApoc/OpenApoc/files/5009611/HireFireStaffCTD1.zip)

[HireFireStaffCTD2.zip](https://github.com/OpenApoc/OpenApoc/files/5009612/HireFireStaffCTD2.zip)

This felt like it was surely a problem on my end, so : I updated visual C++, reinstalled Apocalypse, installed 2.0.2.310 with the exe changing nothing to the default settings. Still got 4 CTD in a row pressing the Hire button. Saved before attempts 2 and 3. Other buttons in the base seem to work just fine.

BeornTB:
Tried again with another build: used the uninstall exe, then installed 2.0.2-283. Same outcome.

[0202-283.zip](https://github.com/OpenApoc/OpenApoc/files/5009655/0202-283.zip)

FilmBoy84:
Confirmed, no need for save, this is easily replicated on a new game start
Just go to your base screen
Click Hire/Fire Agents
The game crashes with a hard CTD and NO log entry
Going to try and get a back trace...

FilmBoy84:
Fixed by #923

PR #923, opened 2020-08-02
Fix CTD on the Hire/Fire Staff button

Jarskih:
Fix to #922


makus82:
And here the fixes was delivered)
Thx Jarskih fo your work!

FilmBoy84:
Thanks for the fix as always Jarskih
Good spot on the typo!

Issue #915, opened 2020-07-27
Crash after losing base assault


JonnyH:
Crash after failing base assault. Backtrace:

#7  0x0000555555a7246a in OpenApoc::Base::die (this=0x7fff211892f0, state=..., collapse=false)                                                       
    at /home/jonny/code/projects/OpenApoc/game/state/city/base.cpp:122                                                                                        
#8  0x00005555559c0356 in OpenApoc::Battle::exitBattle (state=...) at /home/jonny/code/projects/OpenApoc/game/state/battle/battle.cpp:3417
#9  0x000055555589c87f in operator() (__closure=0x555558f8b390) at /home/jonny/code/projects/OpenApoc/game/ui/battle/battledebriefing.cpp:23         
#10 0x000055555589e4a2 in std::__invoke_impl<void, OpenApoc::BattleDebriefing::BattleDebriefing(OpenApoc::sp<OpenApoc::GameState>)::<lambda(OpenApoc::Event*)>
&, OpenApoc::FormsEvent*>(std::__invoke_other, struct {...} &) (__f=...) at /usr/include/c++/10.1.0/bits/invoke.h:60
#11 0x000055555589e374 in std::__invoke_r<void, OpenApoc::BattleDebriefing::BattleDebriefing(OpenApoc::sp<OpenApoc::GameState>)::<lambda(OpenApoc::Event*)>&, 
OpenApoc::FormsEvent*>(struct {...} &) (__fn=...) at /usr/include/c++/10.1.0/bits/invoke.h:110
#12 0x000055555589e16a in std::_Function_handler<void(OpenApoc::FormsEvent*), OpenApoc::BattleDebriefing::BattleDebriefing(OpenApoc::sp<OpenApoc::GameState>):
:<lambda(OpenApoc::Event*)> >::_M_invoke(const std::_Any_data &, OpenApoc::FormsEvent *&&) (__functor=..., __args#0=@0x7fffffffded0: 0x555557d59490)
    at /usr/include/c++/10.1.0/bits/std_function.h:291
#13 0x00005555556d8319 in std::function<void (OpenApoc::FormsEvent*)>::operator()(OpenApoc::FormsEvent*) const (this=0x555558f8b390, __args#0=0x555557d59490)
    at /usr/include/c++/10.1.0/bits/std_function.h:622
#14 0x00005555556d4617 in OpenApoc::Control::triggerEventCallbacks (this=0x555558a93ea0, e=0x555557d59490)
[...]

PR #924, opened 2020-08-02
Fixes for destroyed bases #915

Jarskih:
The agents and vehicles are marked dead/destroyed and they are cleaned up on next update after they die.
Should fix removing agents and vehicles during iterating through the lists.

### DIFF
--- a/game/state/city/base.cpp
+++ b/game/state/city/base.cpp
@@ -121,14 +121,11 @@ void Base::die(GameState &state, bool collapse)
 	}
 	for (auto a : building->currentAgents)
 	{
-		auto agent = a;
-		agent->die(state, true);
+		a->die(state, true);
 	}
-	auto vehicles = building->currentVehicles;
-	for (auto v : vehicles)
+	for (auto v : building->currentVehicles)
 	{
-		auto vehicle = v;
-		vehicle->die(state, true);
+		v->die(state, true);
 	}
 	building->base.clear();
 	building->owner = state.getGovernment();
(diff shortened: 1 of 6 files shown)


JonnyH:
Part of me is wondering why we have a ->die() method when there's a perfectly valid destructor. Some part of me doesn't really see it the concern of the object itself "killing" itself, so much as maybe the containing StateRefMap instead? Then make the destructor private and the StateRefMap a friend to enforce "All objects must be killed by the StateRefMap" which guarantees we never get into the situation of a hanging reference again?

Not asking for changes to this PR, it looks like a good fix, so much as thinking aloud of ways of trying to simplify and detangle the 'ownership' and separation of concerns that the StateRef<> hierarchy has become.

Jarskih:
Sounds like you are thinking about bigger refactoring task? Should the VehicleStatus and AgentStatus be serialized? I am guessing having the health serialized should be enough.

Seems that my clang format does not work again with Visual studio. Any tips how to get it working properly?

FilmBoy84:
Regarding serialisation of status
Are shields included in vehicle/agent health?
If not, those should probably be serialised too

As to the CLANG issue, are you on CLANG 8?
Jonny upgraded the version a while back from 4.0.1

Jarskih:
Thanks FilmBoy84 updating clang fixed it.

FilmBoy84:
NP Jarskih
The move to CLANG 8 got me once too 😁 

LGTM, if JonnyH is fine with the destructor i'm good to merge

We do need a separate PR at some point to deal with the StateRefMap and risk of hanging references as JonnyH has described though

emptyVoid:
I would suggest using Github auto-close patterns for such PRs, otherwise the number of no-longer-valid issues will eventually flood the tracker once again.

Issue #849, opened 2020-05-01
Civilians still appearing as hostile in turn based and real time


Proselyte-Arcturus:
Civilian units will still appear as hostile in turn-based and real-time if enough of them are killed by xcom on previous missions. Civilians are ignored in real-time but still appear as hostile units, in turn-based they are actually reaction fired upon if they move in LoS.

makus82:
Hey @Proselyte-Arcturus ty for report !
you use last build?
you have some saves ? or video

Proselyte-Arcturus:
I can get some together, and yes, the last build I downloaded was about a week ago, but I had heard from Film that this issue was supposedly fixed months ago.

makus82:
yes, well we need then continue to fix it)

Proselyte-Arcturus:
![Hostile Civilians](https://user-images.githubusercontent.com/52583706/80809368-88354c80-8b8f-11ea-960a-42f19f27890f.png)



[save_Psionic Rebalance Battlescape.zip](https://github.com/OpenApoc/OpenApoc/files/4564427/save_Psionic.Rebalance.Battlescape.zip)

Proselyte-Arcturus:
Sorry if that's not enough, let me know if you need any additional documentation.

FilmBoy84:
@Jarskih If you are able can you please look into this as it appears to have occurred despite your fix for civilians earlier in the year

Many thanks

Atrosha:
#816 solve only agents autoattack on civilians, but doesn't fix relationship between x-com and current mission site's organization and civilians from it.

dethstryke01:
Not sure if this is the same issue or not, but in this situation all civilians on the map became hostile and non-targetable for some reason. This did not affect mission victory. 

https://youtu.be/U8FNmmOGJZg?t=4728
[save_test 10.zip](https://github.com/OpenApoc/OpenApoc/files/14183090/save_test.10.zip)
[settings.zip](https://github.com/OpenApoc/OpenApoc/files/14183093/settings.zip)

PR #926, opened 2020-08-03
Civilians should not appear hostile and not cause reaction fire

Jarskih:
Fix to #849

Civilians are never considered enemies.
Added check for reaction fire to not shoot civilians.


FilmBoy84:
LGTM, great work as always Jarskih
Merging

PR #938, opened 2020-08-14
Readme shortened in Whats left and added Copyright notice

makus82:
Readme shortened in Whats left and added Copyright notice

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,11 +44,11 @@ All rights for the original game and its resources belong to their respective ow
 
 ## What's left?
 
-TO HAVE A TRULY PLAYABLE ALPHA STATE #1:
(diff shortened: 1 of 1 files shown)


FilmBoy84:
Probably worth mentioning that pirate ISOs are not supported and will cause issues such as crashes and map problems so the OpenApoc team cannot respond to any query regarding OpenApoc when using a pirate ISO

(This is because many pirate versions of apoc miss certain map files or the music, but obviously don't say *why* they don't work - only that they will not)

makus82:
You can do it after) 
I think all written includes all what you want add

FilmBoy84:
Now part of the existent readme
Closing

Thanks for originally doing this Makus 🍻

PR #943, opened 2020-10-06
Show TU costs for different aiming modes in turn-based battlescape

TreacherousOne:


forms/control.cpp
```diff
@@ -305,15 +305,25 @@ void Control::eventOccured(Event *e)
 			{
 				Vec2<int> pos = {e->forms().MouseInfo.X, e->forms().MouseInfo.Y};
 				e->Handled = true;
-
-				sp<Image> textImage = ToolTipFont->getString(ToolTipText);
-
+				
+				auto lines = split(ToolTipText, "\n");
```
JonnyH, line 309: lines can be const?

forms/control.cpp
```diff
@@ -305,15 +305,25 @@ void Control::eventOccured(Event *e)
...
+				std::list<sp<Image>> textImages;
+				unsigned maxTextWidth = 0;
+				for (auto &line : lines)
```
JonnyH, line 312: line can be const?

forms/control.cpp
```diff
@@ -305,15 +305,25 @@ void Control::eventOccured(Event *e)
...
+				{
+					auto textImage = ToolTipFont->getString(line);
```
JonnyH, line 314: textImage can be const?

forms/control.cpp
```diff
@@ -305,15 +305,25 @@ void Control::eventOccured(Event *e)
...
+					if (textImage->size.x > maxTextWidth)
+						maxTextWidth = textImage->size.x;
+					textImages.push_back(textImage);
+				}
+				
 				unsigned totalBorder = 0;
 				for (const auto &b : ToolTipBorders)
 					totalBorder += b.first;
 
+				auto totalTextHeight = ToolTipFont->getFontHeight() * textImages.size();
```
JonnyH, line 324: totalTextHeight can be const?

forms/control.cpp
```diff
@@ -326,8 +336,14 @@ void Control::eventOccured(Event *e)
 					                        b.second, b.first);
 					i += b.first;
 				}
-				fw().renderer->draw(textImage, {totalBorder, totalBorder});
-
+				
+				unsigned ypos = totalBorder;
+				for (auto &textImage : textImages )
```
JonnyH: textImage can be const?

JonnyH (changes requested):
I think this opens up exactly where word wrapping should be performed - should this be pushed into Font::getString()? If not maybe adding a check for '\n' there and splitting out a warning it's not supported would be useful to avoid future missteps?

TBH text rendering is a PITA to begin with - general word wrapping (where to 'split' lines) and non-latin glyphs can get super hard to layout - maybe we should abstract out all that (and maybe backend onto a library that actually does it properly, like harfbuzz or something).

But that's much more 'maybe long-term fix' than this PR requires - just adding a warning and adding 'const' to everything possible makes this look good to me right now :)

PR #944, opened 2020-10-09
Fix crash on manufacture completed event.

TreacherousOne:
Cause : invalid type specifier in message box format string,


FilmBoy84:
Many thanks for the fix @TreacherousOne confirmed here so merging 👍

PR #946, opened 2020-10-12
Interrupt player's battleunits missions on HostileSpotted event in TB

TreacherousOne:
Interrupt player's battleunits missions on HostileSpotted event in TB


FilmBoy84:
LGTM Treacherous One
Build checks out
Merging

Issue #737, opened 2019-09-07
[Battlescape] Agent out of ammo Notification


FilmBoy84:
A notification for when an agent is out of ammo on their current weapon and one for when an agent is out of ammo on ALL weapons in their inventory would be really useful

(Identified on QMs stream)

FilmBoy84:
TreacherousOne has implemented this feature
Closing

PR #948, opened 2020-10-22
Generate notification when agent is out of ammo.

TreacherousOne:
This addresses #737


FilmBoy84:
Thanks for implementing this earlier than i had expected it to be
Build checks out
Merging

PR #954, opened 2020-11-18
Destroy base if there is no agents to defend it

Jarskih:


game/state/gameevent.cpp
```diff
@@ -370,11 +370,11 @@ GameSomethingDiedEvent::GameSomethingDiedEvent(GameEventType type, UString name,
 		case GameEventType::BaseDestroyed:
 			if (actor.length() > 0)
 			{
-				messageInner = tr("X-COM base destroyed by hostile forces.");
+				messageInner = tr("X-COM " + name + " destroyed by hostile forces.");
```
JonnyH, line 373: This doesn't work well with translations - something like "format(tr("X-COM %s destroyed by hostile forces"), name)" should allow the string to be translated.

tr() should only take static strings, as that can then be extracted for translation.

JonnyH (changes requested):

game/state/gameevent.cpp
Jarskih: Oh right didn't think about that.

FilmBoy84:
@Jarskih Thanks for looking into this

One other query regarding the process; I see you have the base announced as destroyed and the building collapse

Can you confirm this also kills the workers assigned to the building by the parent organisation?

Or do they look for new work elsewhere or get reassigned?

Hopefully this will be dealt with by another part of the economy code

But, to prevent further issues, it may be worth checking out as i can see no part of this PR that deals with the effect on assigned workers and the parent company - If it's already catered for elsewhere, we need to know, if it is not, then it's probably something i need to add to the to-do list

Jarskih:
This mainly deals with the issue when base was undefended and aliens invaded it. It triggered a battle which immediately ended because there was no x-com agents.
I don't think this affects the worker economy because it is x-com base which gets destroyed?

FilmBoy84:
Thanks for the prompt response

Most buildings appear as Government Organisation until purchased by X-COM

When X-COM is booted out by destruction, government should own again as it starts to rebuild in the cityscape

This is where workers are relevant for X-COM buildings - ether before purchase or after they are destroyed

If it's already dealt with No Probs :)

PR #963, opened 2020-11-23
Implement a cap on number of concurrent samples

JonnyH:
Lots of samples together just sound like noise, and can have bad corner cases (e.g. after unpausing multiple of the same sample could perfectly overlap making it super loud and clip)


FilmBoy84:
Finally managed to test this out
LGTM and does resolve that awful loudness issue with stacked samples
Merging

PR #964, opened 2020-11-26
Fixed bugs related to firing scientists

Jarskih:
Added a fire agent method. Before they were killed which makes no sense.
The agents are properly removed from the lab when fired.
The agents who gets fired now goes back to their home building of their original organization.
It should be possible to hire agents again after firing them.


game/state/gamestate_serialize.xml
```diff
@@ -1,2064 +1,2065 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <openapoc_gamestate>
-	<object external="true">
```
JonnyH, line 3: I guess you used an auto-formatter on the xml file here? What tool did you use? It makes sense to have consistency, otherwise it'll cause pretty much every line to change any time someone edits anything, like here.

And even if reformatting the xml is a good idea, it should probably be a separate PR completely, as otherwise 'git blame' will point out this commit for every modified line, which is less than helpful.

game/state/shared/agent.cpp
```diff
@@ -387,12 +389,67 @@ void Agent::assignTraining(TrainingAssignment assignment)
...
+		owner = state.getGovernment();
+	}
+	else
+	{
+		owner = oldOwner;
+	}
+
+	if (!owner->buildings.empty())
+	{
+		homeBuilding = owner->buildings.front();
```
JonnyH, line 430: Is this correct? If a org has multiple buildings, all fired agents will return to the first one in the list?

What does the OG do? Random of each building? Store which building they were initially assigned? Just magically make the agent disappear as the "home building" of agents is a mechanic with little gameplay use?

JonnyH (changes requested):
Generally LGTM, though the xml reformatting should be another PR if it's intended, and if it is intended we probably need to discuss what formatting tools are available everywhere for consistency.

JonnyH:
For the xml formatting, I used libxml2's "xmllint" (http://www.xmlsoft.org/) - it's even wired up in the buildsystem as the 'format-xml' target. I think that's available pretty much everywhere? On linux, macos homebrew and windows vcpkg? Might be worth 'formalizing' that as the one-true-xml-format, maybe even add it to the lint pass like clang-format?

game/state/gamestate_serialize.xml
Jarskih: I used Visual studio and it formatted this file. I didn't mean to change the formatting only add one item to agent. I haven't worked much with xml so I will need to check if I can get the xmllint working with visual studio.

game/state/shared/agent.cpp
Jarskih: Yes that is how it works currently. 
I am not sure what happens in OG to be honest. But since this has almost no effect on gameplay I thought we could keep it simple. The scientists just travel between some building and x-com base when hired or fired.

There might be a problem with this solution though if there is no buildings available for the organization.  If you have better idea how to handle this I can make changes.

game/state/gamestate_serialize.xml
Jarskih: It seems we are using tabs and indentation of 8 for xml? gamestate_serialize.xml seems fixed now.

Jarskih:
I could make changes to this but I am not sure how OG handles firing agents and scientists. 

Does anyone know which buildings/organizations the agents go when they are fired?
Do they teleport to the new home building from x-com base after being fired?

PR #967, opened 2020-12-12
Use github actions for linux CI builds

JonnyH:
Now travis-ci isn't free it makes sense to move over

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,46 @@
+name: CMake
+
+on: [push]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Release
+
+jobs:
+  build:
+    # The CMake configure and build commands are platform agnostic and should work equally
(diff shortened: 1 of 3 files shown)


.github/workflows/cmake.yml
```diff
@@ -0,0 +1,61 @@
...
+    # well on Windows or Mac.  You can convert this to a matrix build if you need
+    # cross-platform coverage.
+    # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Checkout submodules
+      uses: textbook/git-checkout-submodule-action@master
```
SupSuper, line 21: This shouldn't need an extra action, you can just add `submodules: true` or `submodules: recursive` to the checkout action

SupSuper:
Looks good, might be worth setting up something like ccache.
You can see how OpenXcom does it if it helps: https://github.com/OpenXcom/OpenXcom/blob/master/.github/workflows/ci.yml

PR #980, opened 2021-01-30
Implement xcom funding and weekly reports

idshibanov:
Player organization now will receive government funding based on it's weekly rating. Funding is adjusted based on OG formula. Added a weekly funding assessment screen.

Weekly score now reset at the start of every week.

Also fixed weekly loop as it was running on Tuesday morning instead of Monday.

image


game/state/rules/weeklyrating.h
```diff
@@ -0,0 +1,28 @@
...
+
+namespace OpenApoc
+{
+
+class WeeklyRating
+{
+  public:
+	enum class Type
+	{
+		Terrible,     // -1600
```
JonnyH, line 11: Is it really necessary to have the names as part of the rating enum? Would just an integer be better and more flexible in the future? And just map them to names for user-visible text output?

game/state/gamestate.h
```diff
@@ -107,6 +109,7 @@ class GameState : public std::enable_shared_from_this<GameState>
 
 	int baseIndex = 1;
 	int difficulty = 0;
+	bool fundingTerminated = false;
```
JonnyH, line 112: Any time you add something to the GameState (or any child classes) you need to add it to the gamestate_serialize.xml, as otherwise this value won't persist over save/load

game/ui/city/scorescreen.cpp
```diff
@@ -156,11 +164,19 @@ void ScoreScreen::setFinanceMode()
 		    ->setText(format("$%d", agentsSalary + basesCosts));
 
 		int balance = state->getPlayer()->balance;
+		
+		// Special case: when shown during weekly upkeep balance was already adjusted by the game loop
```
JonnyH, line 168: The clang-format version in the CI wants this comment line to be broken, only a small format change but keeping that happy makes it easier to spot more important issues :)

game/ui/city/weeklyfundingscreen.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+	if (government->isRelatedTo(player) == Organisation::Relation::Hostile)
+	{
+		ratingDescription =
+		    tr("The Senate has declared total hostility to X-COM and there will be no further "
+		       "funding. Furthermore, the Senate will take any steps necessary to destroy the "
+		       "X-COM organization if it refuses to cease operation.");
+
+		currentIncome = 0;
+		labelAdjustment->setText(emptyString);
```
JonnyH, line 55: Having to keep around an empty string just to clear text boxes feels a little weird, we probably should have a Label::clear() method or something?

game/ui/city/weeklyfundingscreen.cpp
```diff
@@ -0,0 +1,141 @@
...
+		labelNextWeekIncome->setText(emptyString);
+	}
+	else
+	{
+		const auto rating = WeeklyRating::getRating(state->weekScore.getTotal());
+		const int mod = WeeklyRating::getRatingModifier(rating);
+		int adjustment = (mod == 0) ? 0 : player->income / mod;
```
JonnyH, line 71: Can be const?

JonnyH (changes requested):
generally looks cool, some serialization issues (IE you need to modify the xml if you alter GameState or any child classes), some things probably should be modd-able instead of constants in-code (but that can wait to be fixed later), and I love the const keyword a little too much.

game/ui/city/weeklyfundingscreen.cpp
idshibanov: Removed default text on the form iteself, no longer needed.

game/state/rules/agenttype.h
```diff
@@ -14,6 +14,15 @@
 namespace OpenApoc
 {
 
+static const int HIRE_COST_SOLDIER = 600;
```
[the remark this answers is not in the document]
idshibanov: Potential problem here is that old saves won't have any values in them. Then I have to hardcode the check in `GameState::initState()` which will look something like this:
```
if (this->agent_salary.empty())
{
	this->agent_salary = {{AgentType::Role::Soldier, 600},
                                           {AgentType::Role::BioChemist, 800},
                                           {AgentType::Role::Physicist, 800},
                                           {AgentType::Role::Engineer, 800}};
}
```

idshibanov:
Feedback addressed, moved all hardcoded values to serialized game state. Don't forget to re-run DataExtractor before testing the change.

game/state/rules/agenttype.h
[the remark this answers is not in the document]
FilmBoy84: TBH, we're at Alpha, whilst it would be **nice** to have those values; they're not going to come from nowhere

For testing and bug-whacking, users should be starting a new game everytime a feature is added to trunk anyway

Once merged, I'm tempted to say we bump to the Alpha 0.3 series and tell all users to start new games just to reinforce that mindset

Thanks for all your effort with this by the way
I'll review the code fully once I've finished with the current sprite stuff I'm testing for our 3D modellers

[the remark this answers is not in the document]
JonnyH: I certainly don't worry about save compatibility right now - especially for values that aren't really useful for mods (as mods are just 'saves' we load on top of the base state, but stuff like this is new info exposed so no current modding stuff could use it anyway)

It doesn't seem out of the question to have a "save version" - maybe increment a number every time we /know/ we've broken something terrible and tell the user, but I fear people will take that as a "If the number hasn't incremented I expect saves to work", which isn't really something I want to put work into right now.

Maybe post-1.0 we can do better trying to keep this stable, but right now I wouldn't put any effort into compatibility.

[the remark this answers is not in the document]
idshibanov: Alright, I'll keep it in mind in the future. Regarding this particular PR I already added defaults to gamestate init. We can remove them in the future.

FilmBoy84:
Spent the day trying to break it, and not errors encountered other than those already listed in other issues
Merging; if anything arises; we can try to fix then

Many thanks for your work @idshibanov

PR #981, opened 2021-01-31
Forward unhandled game state events to previous stage

idshibanov:
Sometimes game state events are being ignored if currently open Stage isn't Battleview or Cityview.

Simplified fix for the issue. Suggestions?

### DIFF
--- a/framework/framework.cpp
+++ b/framework/framework.cpp
@@ -483,6 +483,21 @@ void Framework::processEvents()
 			case EVENT_WINDOW_CLOSED:
 				shutdownFramework();
 				return;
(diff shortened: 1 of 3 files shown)


FilmBoy84:
Have you closed this because it's resolved elsewhere?
If so, will stop trying to think of alternate solutions

idshibanov:
> 
> Have you closed this because it's resolved elsewhere?
> If so, will stop trying to think of alternate solutions

I think we need a better implementation of it; a secondary event queue just for GameEvent type, so they won't be cleared on every loop.

Forcing it the way I did here added some side effects.

PR #984, opened 2021-02-06
Fix crash due to invalid cargo log format

### LINKED ISSUE #929: [Cityscape] CTD when targetting UFO
First UFO incursion, multiple vehicles selected. Target the first UFO, everything works. 5/6 vehicles have landing pads to takeoff.

2nd UFO comes into play, change all of them to the 2nd target, using the Target UFO button. Instant CTD.

[CTD Target UFO.zip](https://github.com/OpenApoc/OpenApoc/files/5017097/CTD.Target.UFO.zip)

### LINKED ISSUE #976: Game crash after selecting a vehicle
Hi,
the game crashes in certain situations. I assume all those I've experienced will have the same root cause. I'm attaching a savefile with the following steps to replicate:
1) After loading the game simply resume the time in the tactic mission
2) Mission will end after a little while as there is no other enemy in the area.
3) Confirm the summary screen
4) You should see the selection of Valkyre leaving the Temple of Sirius. The game should be paused (I didn't test continuing the time)
5) Click on the Valkyre (on the main screen in the temple, not on UI or anywhere else)
image
6) Game crashed

**Note**: It seems to be the problem with sort of layering as the problem occurs only when you click in the **center** of the rectangle - the corners work as expected I'd say.

[save_1.zip](https://github.com/OpenApoc/OpenApoc/files/5886816/save_1.zip)

I'm sorry if it's a duplicate already.
This type of crash happened to me several times already so it doesn't seem to be as rare as it might look like.
If anything else you need let me know I'll try to help

Cheers,
Jouki

idshibanov:
Fixes #976 . Fixes #929 .

### DIFF
--- a/game/ui/tileview/cityview.cpp
+++ b/game/ui/tileview/cityview.cpp
@@ -3298,7 +3298,7 @@ bool CityView::handleMouseDown(Event *e)
 					}
 					for (auto &c : vehicle->cargo)
 					{
-						LogWarning("Cargo %dx%s", c.id, c.count);
+						LogInfo("Cargo %sx%d", c.id, c.count);
 					}
 					if (modifierLAlt && modifierLCtrl && modifierLShift)
 					{
@@ -3347,7 +3347,7 @@ bool CityView::handleMouseDown(Event *e)
 					}
 					for (auto &c : vehicle->cargo)
 					{
-						LogWarning("Cargo %dx%s", c.id, c.count);
+						LogInfo("Cargo %sx%d", c.id, c.count);
 					}
 				}
 			}


JonnyH:
Seems a pretty slam-dunk fix - the there may be other similar issues hanging around since the {fmt} change, as before we used tinyformat which was more permissive with %format specifiers.

idshibanov:
> 
> 
> Seems a pretty slam-dunk fix - the there may be other similar issues hanging around since the {fmt} change, as before we used tinyformat which was more permissive with %format specifiers.

Yes, I quickly checked the backlog and identified that those two crashes definitely happened because of the format mistake. I'm sure there could be others, have to keep an eye on this in the future.

PR #998, opened 2021-02-22
An illegal flyer has been detected (implement organisation raids)

idshibanov:
Related to #1002. Bug should be fixed now, extra suggestions can be done later.

Refactored Organisation missions to split into two types to support new code. Now AI will fight with each other. I tried to keep the logic (timers, chance %) very close to the original.

Base defense event will be launch if they "Raid" or "Storm" player owned building with a base.

Added a building attack ammo check to avoid vehicles hanging for too long.


FilmBoy84:
Fantastic work!

Just going through the code now whilst I make a build

Couple of questions regarding the vehicles sent on missions
I notice that you have `int attack_vehicle_count = 2;` under organisationsraid.h
Is this ALWAYS two vehicles in the OG?
Or should the number vary with manpower and finances?

I also note that you have under organisations.cpp several vehicles that the organisations pick according to manpower (which is great)
`state.organisation_raid_rules.attack_vehicle_types = {
	    {&state, "VEHICLETYPE_PHOENIX_HOVERCAR"},
	    {&state, "VEHICLETYPE_HOVERBIKE"},
	    {&state, "VEHICLETYPE_VALKYRIE_INTERCEPTOR"},
	    {&state, "VEHICLETYPE_HAWK_AIR_WARRIOR"}};`
Are raids in the OG always airborne?
Do they ever send ground vehicles such as Griffons, Bikes or Stormdogs?

What considerations are in place for MegaPol?
Should they raid with Police Hovercars and Police Cars in greater numbers (as police generally do)?

FilmBoy84:
Thanks for finding out for us Skin36

So it does indeed vary with a maximum value

I also note that it is dependant on number of owned vehicles in the organisation - are we currently tracking this in OpenApoc?

idshibanov:
> Couple of questions regarding the vehicles sent on missions
> I notice that you have `int attack_vehicle_count = 2;` under organisationsraid.h
> Is this ALWAYS two vehicles in the OG?
> Or should the number vary with manpower and finances?

This variable should be exposed for modding.
In OG in practice it almost always two vehicles and limited by what's available in the randomly picked building.
Right now in OpenApoc organisations don't spend money on vehicles, we can add a check later 

> Are raids in the OG always airborne?
> Do they ever send ground vehicles such as Griffons, Bikes or Stormdogs?

 Yes, it specifically looking for flyers and priority order is how I wrote it down. If you manually remove smaller vehicles (edit savegame) from organisation's vehicle park then they will start sending Hawks.

> What considerations are in place for MegaPol?
> Should they raid with Police Hovercars and Police Cars in greater numbers (as police generally do) with the Valkyrie and Hawk as support?

This is interesting. In fact Megapol does send 'Illegal vehicles', but it's always Police Hovercars and those doesn't trigger the event, they are just silently ignored. Raids currently happen without sending a vehicle over (as in OG); we can make it happen later.

idshibanov:
> I also note that it is dependant on number of owned vehicles in the organisation - are we currently tracking this in OpenApoc?

Yes, vehicles are distributed in `Organisation::updateVehicleAgentPark` call. Organisations don't buy new vehicles though, that's a TODO.

FilmBoy84:
Thanks for the responses Shellstorm
I have opened issues #999 and #1000 to keep track of the To-Do stuff
I shall continue trying to break things in my build
Otherwise, LGTM so far
Pending any bugs between now and play tomorrow, I'm happy to merge if @JonnyH is happy with the code

makus82:
Strange but during whol game week i not get any alert about "Unautorized vehicle detected"

idshibanov:
> 
> 
> Strange but during whol game week i not get any alert about "Unautorized vehicle detected"

It depends on game's difficulty and random chance. Without Alien control only Psyke, Diablo and Osiron will send illegal flyers, have to piss off one of those.

makus82:
there was Nutrivend or something was under control of aliens after day 1 or 2)...
got it thx

FilmBoy84:
Thanks for updates Shellstorm
Will make a build and check out
I'm clear to merge by the way if noone finds any more issues :)

PR #1005, opened 2021-03-05
Added Alert header to Alert screen - Update alert.form

makus82:
Added Alert header to Alert screen


FilmBoy84:
Thanks for the form fix Makus
Merging

makus82:
ye that was easy =]

PR #1015, opened 2021-03-16
Revert "Readme update 2"

FilmBoy84:
Reverts OpenApoc/OpenApoc#1014
Build was failing but was merged
Reverting till we have a clean branch to merge so that Appveyor is not a sea of red


FilmBoy84:
Going to need a more comprehensive solution
As just reverting this is not going to fix the mess - #1014 and #1012 were both merged with failing builds
So cancelling this request

Will aim to make a fix PR over coming days

PR #1017, opened 2021-03-17
Remove vcpkg install lines from appveyor

JonnyH:
This should be handled automatically by cmake


JonnyH:
The current appveyor builds seem to be timing out in the cmake vcpkg handling - possibly it's just taking that long (~1 hour) to update that stuff?

Perhaps it's doing everything twice, as we "manually" vcpkg install'ed as well as the cmake integration?

So this is more a "testing" PR to kick the CI instead of actually expecting to be merged.

PR #1030, opened 2021-04-09
Nightly Score & Finance screen

gnegno84:
As in title
I've also included a call to an hook for LUA script that i don't think exists (but i've added it for simmetry with end of week handling)

### DIFF
--- a/game/state/gametime.cpp
+++ b/game/state/gametime.cpp
@@ -128,7 +128,7 @@ unsigned int GameTime::getMonth() const
 	const date currentDate = getPtime(this->ticks).date();
 
 	const int months = (currentDate.year() - GAME_START.date().year()) * 12 + currentDate.month() -
-	             GAME_START.date().month();
+	                   GAME_START.date().month();
 	return months;
 }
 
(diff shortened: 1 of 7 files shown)


game/state/gametime.cpp
```diff
@@ -140,9 +140,32 @@ unsigned int GameTime::getWeek() const
 	return duration.days() / 7 + 1;
 }
 
+unsigned int GameTime::getFirstDayOfCurrentWeek() const
+{
+	date today = getPtime(this->ticks).date();
```
JonnyH, line 145: I think these local variables can be const?

game/state/gametime.cpp
```diff
@@ -140,9 +140,32 @@ unsigned int GameTime::getWeek() const
...
+
+		date first_day_of_week = today - days(days_to_monday);
+		return first_day_of_week.year_month_day().day;
+	}
+}
+
 unsigned int GameTime::getLastDayOfCurrentWeek() const
 {
-	unsigned int daysBeforeWeekEnd = 7 - getPtime(this->ticks).date().day_of_week();
+	unsigned short dayOfWeek = getPtime(this->ticks).date().day_of_week();
```
JonnyH, line 164: More local variables can be const?

JonnyH (approved):
Generally LGTM - though I tend to prefer every variable that can be marked as "const" to be marked as such - I think it's easier to "know" that nothing in that function modifies that variable later.

gnegno84:
> Generally LGTM - though I tend to prefer every variable that can be marked as "const" to be marked as such - I think it's easier to "know" that nothing in that function modifies that variable later.

Thanks for the hints and checks. Totally agree with your points. I’ll update the PR asap.

FilmBoy84:
LGTM, if the build error is solely down to Appveyor being a pain (which it appears from the log that it is), I'm good to merge if JonnyH is?

I've found no issue when building myself

JonnyH:
Looks like the build issue was just a transitory failure - maybe due to chocolatey updating nsis or something. Seems fine after being re-kicked.

PR #1032, opened 2021-04-21
Weekly report fix

gnegno84:
There are some issues with the weekly report due to racing conditions of the update of the GameState that is used as reference in the weekly report. 

The idea is to create a small snapshot of the GameState that can be used to fix this issue.
In order to do that we need to add a new event, update the WeeklyFundingScreen and ScoreScreen, and refactor the GameState handling of the weeklyUpdate event.

Please, double check if what i've done is legit/correct from a c++ point of view (pointers etc...) as I still have a lot to learn.

Ps: in this PR i've added also the label coloring in WeeklyFundingScreen that was depending on.

### DIFF
--- a/data/forms/city/weekly_funding.form
+++ b/data/forms/city/weekly_funding.form
@@ -55,7 +55,7 @@
       </label>
       <label id="FUNDING_ADJUSTMENT">
         <position x="52" y="212"/>
-        <size width="200" height="15"/>
+        <size width="0" height="15"/>
         <alignment horizontal="left" vertical="centre"/>
         <font>smalfont</font>
       </label>
--- a/data/forms/city/weekly_funding.form
+++ b/data/forms/city/weekly_funding.form
@@ -37,7 +37,13 @@
 	  
       <label id="FUNDING_CURRENT" text="Current income>">
         <position x="52" y="140"/>
-        <size width="400" height="15"/>
+        <size width="0" height="15"/>
+        <alignment horizontal="left" vertical="centre"/>
+        <font>smalfont</font>
+      </label>
+      <label id="FUNDING_CURRENT_VALUE">
+        <position x="0" y="0"/>
+        <size width="0" height="15"/>
         <alignment horizontal="left" vertical="centre"/>
         <font>smalfont</font>
       </label>
@@ -49,17 +55,29 @@
       </label>
       <label id="FUNDING_ADJUSTMENT">
         <position x="52" y="212"/>
-        <size width="400" height="15"/>
+        <size width="200" height="15"/>
+        <alignment horizontal="left" vertical="centre"/>
+        <font>smalfont</font>
+      </label>
+      <label id="FUNDING_ADJUSTMENT_VALUE">
+        <position x="0" y="0"/>
+        <size width="0" height="15"/>
         <alignment horizontal="left" vertical="centre"/>
         <font>smalfont</font>
       </label>
       <label id="FUNDING_NEW">
         <position x="52" y="240"/>
-        <size width="400" height="15"/>
+        <size width="0" height="15"/>
         <alignment horizontal="left" vertical="centre"/>
         <font>smalfont</font>
       </label>
-
+      <label id="FUNDING_NEW_VALUE">
+        <position x="0" y="0"/>
+        <size width="0" height="15"/>
+        <alignment horizontal="left" vertical="centre"/>
+        <font>smalfont</font>
+      </label>
+      
       <graphicbutton id="BUTTON_OK">
         <position x="602" y="443"/>
         <size width="35" height="34"/>
--- a/game/state/playerstatesnapshot.cpp
+++ b/game/state/playerstatesnapshot.cpp
@@ -1,0 +1,63 @@
+#include "game/state/playerstatesnapshot.h"
+#include "game/state/city/base.h"
+#include "game/state/city/facility.h"
+#include "game/state/shared/organisation.h"
+
+namespace OpenApoc
+{
+
+PlayerStateSnapshot::PlayerStateSnapshot(sp<GameState> stateToSnap)
+    : fundingModifier(stateToSnap->calculateFundingModifier()),
+      fundingTerminated(stateToSnap->fundingTerminated), gameTime(stateToSnap->gameTime),
+      totalScore(stateToSnap->totalScore), weekScore(stateToSnap->weekScore),
+      weekly_rating_rules(stateToSnap->weekly_rating_rules),
+      agent_salary(stateToSnap->agent_salary), playerBalance(stateToSnap->player->balance),
+      playerIncome(stateToSnap->player->income),
+      governmentBalance(stateToSnap->government->balance),
+      government_player_relationship(stateToSnap->government->isRelatedTo(stateToSnap->player))
+{
+	int soldiers = 0, biochemists = 0, engineers = 0, physicists = 0;
+	for (auto &a : stateToSnap->agents)
+	{
+		if (a.second->owner == stateToSnap->player)
+		{
+			switch (a.second->type->role)
+			{
+				case AgentType::Role::BioChemist:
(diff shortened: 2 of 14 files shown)


data/forms/city/weekly_funding.form
```diff
@@ -49,17 +55,29 @@
...
       <label id="FUNDING_NEW">
         <position x="52" y="240"/>
-        <size width="400" height="15"/>
+        <size width="0" height="15"/>
         <alignment horizontal="left" vertical="centre"/>
         <font>smalfont</font>
       </label>
-
+      <label id="FUNDING_NEW_VALUE">
+        <position x="0" y="0"/>
```
JonnyH, line 75: Having an empty position doest seem right here, the goal of the form xml files is that they can be altered to reskin the game, and so just setting the position in code, instead of using a value relative to an anchor in the xml, limits that.

Possibly the same with the zero sized stuff above, I don't think this screen can reflow (ie each line has a set contents and can't flow to a new line and push everything else down anyway, so having a zero size just makes it less clear to whoever's reading the xml expecting it to relate to the final output)

game/state/playerstatesnapshot.cpp
```diff
@@ -0,0 +1,63 @@
...
+      fundingTerminated(stateToSnap->fundingTerminated), gameTime(stateToSnap->gameTime),
+      totalScore(stateToSnap->totalScore), weekScore(stateToSnap->weekScore),
+      weekly_rating_rules(stateToSnap->weekly_rating_rules),
+      agent_salary(stateToSnap->agent_salary), playerBalance(stateToSnap->player->balance),
+      playerIncome(stateToSnap->player->income),
+      governmentBalance(stateToSnap->government->balance),
+      government_player_relationship(stateToSnap->government->isRelatedTo(stateToSnap->player))
+{
+	int soldiers = 0, biochemists = 0, engineers = 0, physicists = 0;
+	for (auto &a : stateToSnap->agents)
```
JonnyH, line 20: If the agent isn't being modified it can be const

game/state/playerstatesnapshot.cpp
```diff
@@ -0,0 +1,63 @@
...
+					break;
+			}
+		}
+		agent_qty.insert_or_assign(AgentType::Role::BioChemist, biochemists);
+		agent_qty.insert_or_assign(AgentType::Role::Engineer, engineers);
+		agent_qty.insert_or_assign(AgentType::Role::Physicist, physicists);
+		agent_qty.insert_or_assign(AgentType::Role::Soldier, soldiers);
+	}
+
+	for (auto &base : stateToSnap->player_bases)
```
JonnyH, line 46: Const base?

game/state/playerstatesnapshot.cpp
```diff
@@ -0,0 +1,63 @@
...
+	{
+		int basesCosts = 0;
+		for (auto &f : base.second->facilities)
```
JonnyH, line 49: Const facility?

game/state/playerstatesnapshot.h
```diff
@@ -0,0 +1,40 @@
...
+#include "game/state/gamestate.h"
+#include "game/state/gametime.h"
+#include "game/state/shared/organisation.h"
+#include "library/strings.h"
+#include "library/sp.h"
+
+namespace OpenApoc
+{
+
+class PlayerStateSnapshot : public std::enable_shared_from_this<PlayerStateSnapshot>
```
JonnyH, line 12: As this class is relatively small and copy able, it might make sense to avoid using shared ptr for access and just copy as needed? I want to avoid using shared_ptr as much as possible, as it often ends up being a crutch instead of actually thinking about ownership of the data

game/ui/city/scorescreen.cpp
```diff
@@ -10,18 +10,19 @@
...
 #include "game/state/gamestate.h"
+#include "game/state/playerstatesnapshot.h"
 #include "game/state/shared/agent.h"
 #include "game/state/shared/organisation.h"
 #include "game/ui/base/recruitscreen.h"
 
 namespace OpenApoc
 {
-ScoreScreen::ScoreScreen(sp<GameState> state, bool showWeeklyUpkeep)
+ScoreScreen::ScoreScreen(PlayerStateSnapshot state, bool showWeeklyUpkeep)
```
JonnyH, line 20: If this doesn't alter the snapshot, that can be const. Also it can be a reference, as theres no need for a copy here.

game/ui/city/scorescreen.cpp
```diff
@@ -161,25 +152,17 @@ void ScoreScreen::setFinanceMode()
...
 
-		int balance = state->getPlayer()->balance;
-
-		// Special case: during weekly upkeep balance was already adjusted by the game loop
-		if (isWeeklyUpkeep)
-		{
-			// revert balance value to original for display
-			balance += agentsSalary + basesCosts;
-		}
+		int balance = state.playerBalance;
```
JonnyH, line 165: Can be const?

game/ui/city/weeklyfundingscreen.cpp
```diff
@@ -4,22 +4,29 @@
...
-#include "game/state/gamestate.h"
+#include "game/state/playerstatesnapshot.h"
 #include "game/state/shared/organisation.h"
 #include "game/ui/city/scorescreen.h"
 
 namespace OpenApoc
 {
 
-WeeklyFundingScreen::WeeklyFundingScreen(sp<GameState> state)
+WeeklyFundingScreen::WeeklyFundingScreen(PlayerStateSnapshot state)
```
JonnyH, line 17: The snapshot can be a const reference again?

game/ui/city/weeklyfundingscreen.cpp
```diff
@@ -31,25 +38,25 @@ WeeklyFundingScreen::~WeeklyFundingScreen() = default;
...
-	menuform->findControlTyped<Label>("TEXT_DATE")->setText(state->gameTime.getLongDateString());
-	menuform->findControlTyped<Label>("TEXT_WEEK")->setText(state->gameTime.getWeekString());
+	menuform->findControlTyped<Label>("TEXT_FUNDS")->setText(state.getPlayerBalance());
+	menuform->findControlTyped<Label>("TEXT_DATE")->setText(state.gameTime.getLongDateString());
+	menuform->findControlTyped<Label>("TEXT_WEEK")->setText(state.gameTime.getWeekString());
 
 	menuform->findControlTyped<Label>("TITLE")->setText(tr("WEEKLY FUNDING ASSESSMENT"));
 
 	UString ratingDescription;
+	const Colour deepBlueColor = {4, 73, 130, 255};
```
JonnyH, line 54: I need to think of a way to expose things like this to the modding system. Maybe allow named colours in the form files or something? But that's on me not this PR

game/ui/city/weeklyfundingscreen.cpp
```diff
@@ -31,25 +38,25 @@ WeeklyFundingScreen::~WeeklyFundingScreen() = default;
...
+	const Colour cyanColor = {113, 219, 255, 255};
 
-	const auto player = state->getPlayer();
-	const auto government = state->getGovernment();
-	int currentIncome = player->income;
+	int currentIncome = state.playerIncome;
```
JonnyH, line 57: Can be const?

JonnyH (commented):
Done a quick initial pass, still need to actually build and test and possibly poke more nitpicks :p

game/ui/city/weeklyfundingscreen.cpp
gnegno84: This one could change during the execution of the code (e.g: zeroed when funding are cut due to hostile relationships)

data/forms/city/weekly_funding.form
gnegno84: Agree. It's not the best solution. 
What do you suggest?

PR #1041, opened 2021-06-01
Reverts accidental logging change in OpenApoc#1040

zigmar:


FilmBoy84:
Thanks for the quick fix
No worries about the mistake - we all make them
Merging

PR #1047, opened 2021-06-11
Crash fixes

### LINKED ISSUE #1046: Crash when psi-controlled unit dies
When psi-controlled unit dies, the game crashes. 

To reproduce: load this [save game](https://github.com/OpenApoc/OpenApoc/files/6639770/save_psi_crash.zip), get psi-controlled unit die (e.g. suicide by grenade), observe game crash.

Seems to be caused by iterating over std::map while modifying it content at battleunit.cpp:4430 (BattleUnit::dropDown)

zigmar:
- Fix #1046: CTD when psi-controlled unit dies (iterating over container while modifying it)
- Fix crash in debug on bad formatting argument


FilmBoy84:
Thanks for the quick fix
LGTM merging

PR #1048, opened 2021-06-16
Implements a unix-based utility to extract and autoformat contents of game saves

zigmar:
Useful for debugging saved gamestate issues.

### DIFF
--- a/tools/extract_save_game.sh
+++ b/tools/extract_save_game.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+XML_FORMATTER="xmllint"
+XML_FORMATTER_ARGS="--format"
+
+if [ -z "$1" ]; then
+    echo "Usage: extract_save_game.sh <save_game> [output_dir]"
+    echo "       output_dir is current directory by default"
+    exit
+fi
+
+savegame=$1
+outputdir=${2:-'.'}
+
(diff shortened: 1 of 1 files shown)


JonnyH:
The plan was that the tools/serialization_tool would do stuff like this - pack/unpack saves, pretty-print the result, may even lint later (e.g. checks like no un-read nodes or things that may show duplication/typos).

As it uses the game core serialization logic it 'should' work with any save archive (directory/zip/whatever wrappers and internal serialization format we end up actually using if the current xml format is bottlenecking us)

I think I'm the only one who ever used it though, so it could probably do with some love to round it out and make it useful :) I know it currently has an issue with serializing things that contain resource links - if they aren't available at serialization in time (eg not setting up the data/cd directories for the serialization_tool itself) it'll replace them with blank nodes. Which is less than idea.

zigmar:
Ah nice. At the time I was debugging an issue I could not get SerializationTool to do what I needed, but revisited now and after some experimentation and reading the code, managed to do exactly what the script in PR does. I guess it makes the script redundant, especially given that SerializationTool is portable while the script isn't.  Though I could see one instant where it might be useful: in case gamestate fails to read game save (e.g. due to version incompatibility) but this is probably rare enough so it could be done manually.

PR #1049, opened 2021-06-16
Fix crash when agent in cityscape dies

zigmar:
- Fix crash caused by agent being removed during GameState::update by delaying agents removal and avoid modifying STL container while iterating over it. TODO: check if any other containers might be handled incorrectly.
- Updated .gitignrore to ignore case for xcom.bin

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,7 +80,7 @@ CMakeSettings.json
 [Dd]ata/UFOPAEDI/
 [Dd]ata/MUSIC
 [Dd]ata/cd.iso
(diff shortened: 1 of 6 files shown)


FilmBoy84:
Normally, would not merge without further consultation
But as the PR checks out in terms of doing what it's meant to
It fixes several crashes that need resolving
And I cannot see a better way to implement myself
Merging

We can always improve upon codestyle or method if a more efficient way of doing things comes to light via another PR

Tested on Win32 and Win x64 builds 👍

PR #1052, opened 2021-06-24
Fix expired cargo

zigmar:
- Prevents game quitting with error when cargo that wasn't picked up
  expires


FilmBoy84:
Thanks for the quick fix :beers: 
Merging

PR #1074, opened 2021-11-21
Mission briefing texts

Grichmann:
Added loading mission briefing texts from appropriate localisation files.


FilmBoy84:
Many thanks for putting this together.

LGTM, if this is okay with @JonnyH with regards for future translation related updates I'm good to merge

PR #1079, opened 2021-12-12
Added support for Psionic damage type

Jarskih:
I tested this by changing the smoke grenade damage type to Psionic Blast. The psionic attack will use psi defense when calculating if the unit takes damage or not. The grenade damage is used for two things now: determining if it will go through psi defense & morale damage if it goes through.


makus82:
uuu niceee thx @Jarskih

PR #1081, opened 2021-12-14
Remove damage from psionic damage type

Jarskih:


Jarskih:
Sorry. I will remove that xml format commit and make a new one.

PR #1082, opened 2021-12-14
Remove damage from psionic blast

Jarskih:


FilmBoy84:
Much appreciated

To clarify, this ensures only morale damage is done - which still allows for the various panic effects to occur on an afflicted agent - beserk, flee, freeze, etc (anything attached to the panic state basically)

Issue #1084, opened 2021-12-15
marketplace: clicking on one of the arrows on the side of the scroller decrements/increments by two...


darkspyre:
Hi,

in the marketplace screen a click on one of the arrows on the side decreases/increases by two instead of one. Is that intended?

I'm running it on Ubuntu 20.04...

TIA and KR,
Nils

FilmBoy84:
You beat me to it, making videos of the Original Game behaviour and reporting this has been on the cards for a while, thanks for report, I will label this as confirmed and update with video of the expected behaviour and what OpenApoc actually does soon

darkspyre:
Thanks for looking into that...

darkspyre:
Many thanks for fixing - looks good now... :+1:

PR #1086, opened 2021-12-19
Scroll bar fixes #1085 #1084

### LINKED ISSUE #1085: marketplace: use of mouse scroll wheel scrolls the list and moves the amount slider at the same time...
Hi,

in the marketplace screen if I use the scroll wheel, it scrolls the list of items as well as moves the amount slider at the same time. No ideas how to fix that...

I running on Ubuntu 20.04

TIA and KR,
Nils

Jarskih:
- Fix accidentally moving items when scrolling up or down the list
- Fix incrementing by 2 when clicking to buy or sell items

I assumed that we only want to scroll the vertical scroll bars by using the mouse wheel. Original Apoc did not have any scrolling using the mouse wheel.

#1085 #1084


FilmBoy84:
Many thanks for the fix, merged 👍

PR #1092, opened 2022-02-10
Fix switched arguments when extracting strafing when aiming animations

JonnyH:
This looks wrong, passing an integer to a bool argument and bool to integer argument - noticed when looking over the code.

Not 100% on what this actually does or could break, but the types make it look wrong.


FilmBoy84:
Have tested a build with this this morning

Ran a couple of tests in game and nothing immediately wrong - even in a heated cityscape battle where the animation cycling is generally at it's peak and the ImagePacks are hit hard

Merging, as agreed, the arguments appear to be the wrong way round and if, in some unlikely case, it does cause a problem somewhere, it'd be better to fix the cause as a separate pull...

PR #1093, opened 2022-02-25
Weapons with 0 ammo (laser) didn't fire

Jarskih:


FilmBoy84:
Thanks for spotting this and fixing Jarskih

Merging

PR #1096, opened 2022-06-07
Fix the way to specify dl library

AMDmi3:


FilmBoy84:
Good spot!

Thanks for the first submission to the project too!
🍻

Issue #1100, opened 2022-08-19
Bugged text / images when opening equip agent screen with no agents available


dl471:
**Describe the bug**
The Equip agent menu is filled with placeholder text and broken images if the player opens it while having no agents available

**To Reproduce**
1. Fire all agents
2. Let the game time advance for a few seconds
3. Click the "Equip agent" button

**Expected behavior**
In the original game the menu would refuse to open under these circumstances.

**Screenshots**
![openapoc_agent_screen](https://user-images.githubusercontent.com/32369415/185526105-f8f8aa8c-5de0-4806-a79f-1ccdea9206ce.png)

**Desktop (please complete the following information):**
OS: Ubuntu 18.04.3 LTS
OpenApoc Version: Built from master at commit f1a3ecb

FilmBoy84:
Can confirm this error with a new save and this build https://github.com/OpenApoc/OpenApoc/releases/tag/2022-v0.2-539

1) Agents appearing normally on game start
![image](https://user-images.githubusercontent.com/33913088/185604520-c5c0dff3-90eb-4cdd-a5d7-d5a5f2c19c8f.png)

2) All Agents (not scientists) fired
![image](https://user-images.githubusercontent.com/33913088/185604622-15ac23bd-f9c9-47f0-b8aa-a99aaaee9233.png)

3) Additional Bug with Agents NOT immediately disappearing from the equip screen if time is not progressed - they should. For some reason their health appears as 0 though
![image](https://user-images.githubusercontent.com/33913088/185604796-0906a720-9820-4317-ad63-5d8be1bc0161.png)

4) Move forward time 1-2 seconds to force a gamestate update
![image](https://user-images.githubusercontent.com/33913088/185604942-a80b66e0-1434-4948-94e6-45d531515eb1.png)

5) The Agent Equip Screen is now glitched
![image](https://user-images.githubusercontent.com/33913088/185605042-70a78ec7-213f-48ab-9d81-65b512834d75.png)

FilmBoy84:
Thanks for the report @dl471 :beers:

FilmBoy84:
Should be resolved by #1116

PR #1102, opened 2022-08-20
Avoid opening agent/vehicle equip screens if player has no soliders/vehicles

dl471:
At current trying to open the vehicle equip screen with no vehicles leads to an error message and seg fault, and trying to open the agent equip screen with no soldiers leads to issue #1100. I have attempted to fix this by adding preliminary checks to the callback functions for opening the equip screens that check if the player has soldiers/vehicles that they can equip before letting them be opened.

In terms of the implementation I'm not the biggest fan of the duplicated code in the callbacks but I'm not sure how I would go about puling it out into a separate function. Could perhaps be generalized and thrown in gamestate or organisation file but I am not sure if it would be worth it.

### DIFF
--- a/game/ui/base/basescreen.cpp
+++ b/game/ui/base/basescreen.cpp
@@ -15,6 +15,7 @@
 #include "game/state/city/base.h"
 #include "game/state/city/building.h"
 #include "game/state/city/facility.h"
+#include "game/state/city/vehicle.h"
 #include "game/state/gamestate.h"
 #include "game/state/rules/city/ufopaedia.h"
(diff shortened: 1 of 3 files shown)


FilmBoy84:
Many thanks for this PR @dl471 

Apologies for the delay in getting to review it

I'll set about looking at it now :)

Another thing to consider, just glimpsing over the changes, is do we need a prompt with a new string of text - for example

"No agents are present at this location" or "No vehicles are present at this location" so we don't get "I click and nothing happens" type reports on the tracker

PR #1108, opened 2022-10-23
Disable autoscale for macOS

BabyWolf:
Disabled autoscale. With this option all visible. Without only black screen.


FilmBoy84:
Thanks for this quick resolution to getting the game running on MacOS with the defaults and a good spot on the issue

We should probably also add a new Issue to the tracker to resolve the underlying scaling problem on MacOS at a later date (unless you have another PR in the works to resolve it already?)

PR #1114, opened 2022-10-30
Add support for linking to system PhysFS

h3xx:
Introduces new CMake option `USE_SYSTEM_PHYSFS` (default OFF).


JonnyH:
Just to confirm this all works? We were having to patch physfs due to some issues reading the (steam-distributed) Apoc ISO, and some issues on platforms with case sensitivity.

Though from what I remember the issues we patched were reported upstream, so they may have fixed things since the (rather ancient) version we started using.

PR #1116, opened 2023-01-13
Fix double equip screen stage being pushed

luiscamara:
I noticed that the equip screen was requiring two clicks on the ok button to close it. So I investigated and discovered that two stages were being pushed when the BUTTON_EQUIP_AGENT was clicked. 
This PR removes the extra equip screen stage being pushed to the stage queue. There is only one stage push left, which is the one that checks if the player has soldiers, so this should also fix the issue #1100.


FilmBoy84:
Much appreciated

We're getting Lint / Appveyor errors so will merge once those are fixed 🍻

Issue #993, opened 2021-02-12
Agent number should be moved 1 or 2 pixel left not to touch the frame


makus82:
Agent number should be moved 1 or 2 pixel left not to touch the frame
![image](https://user-images.githubusercontent.com/11581207/107780787-d8148700-6d4f-11eb-8154-393c813ef49e.png)

FilmBoy84:
Thanks for the report Makus
This should be a simple form correction providing it is present in all languages
Verified in English too
Updating Label

makus82:
found this tab in forms but there no such element (agents number)
https://github.com/OpenApoc/OpenApoc/blob/70cbddfa109379441fc8509f3dd485c2ac1e5734/data/forms/city/tab2.form

PR #1124, opened 2023-02-24
fix: move the passenger number icon one pixel left.

Kurtsley:
As mentioned in #993, the icon showing the number of passengers in a vehicle was off by one pixel. This PR fixes that.


FilmBoy84:
Thanks for the fix and good spot

Not sure why we are getting a linting error with just two characters difference though

Issue #484, opened 2019-01-15
Crashed Vehicles sometimes do not get recovered, Destroyed Vehicles sometimes do not clear


FilmBoy84:
See attached save, there are several police and civilian cars that have crashed, remain static, but will never be cleared

Ideally, with crashes enabled, the city should send out emergency/construction vehicles to airlift them out of the tile and return to parent organisations, this clears them from the map

If crashes are NOT enabled, then we need to make them "self destruct" after a few hours and be cleared from the map

At present, all that happens is we end up with a map of destroyed vehicles that never get cleared... This slows the game

[save_SuperHuman 2.zip](https://github.com/OpenApoc/OpenApoc/files/2760423/save_SuperHuman.2.zip)
![image](https://user-images.githubusercontent.com/33913088/51193279-65225300-18e0-11e9-8e67-a99776dc9bb6.png)

PR #1129, opened 2023-03-02
Bug Fixes

### LINKED ISSUE #1073: Building Investigate Count <0? on starting a mission at a building where agents are already present
**Describe the bug**
Agents are already commanded to move to an apartment and wait for a alien alert (load the game and ultra fast time)
When the alert appears I select the agents already there to investigate
The following error occurs on confirming the order to investigate
image
Clicking okay will continue the investigation
But something is amiss to cause this

**Savegame**
[save_Economy 1.zip](https://github.com/OpenApoc/OpenApoc/files/7446764/save_Economy.1.zip)

kgd192:
Fix #1073 

Unfortunately due to how Event handling works I am currently unabel to provide a more elegant solution

### DIFF
--- a/game/state/city/vehicle.h
+++ b/game/state/city/vehicle.h
@@ -327,6 +327,8 @@ class Vehicle : public StateObject<Vehicle>,
 	bool canDamageBuilding(StateRef<Building> target) const;
 	bool isIdle() const;
 
+	bool wasAlreadyAtTgtBuilding;
+
 	// This is the 'sum' of all armors?
 	int getArmor() const;
 	int getAccuracy() const;
--- a/game/state/city/vehiclemission.cpp
+++ b/game/state/city/vehiclemission.cpp
@@ -1427,6 +1427,12 @@ void VehicleMission::update(GameState &state, Vehicle &v, unsigned int ticks, bo
 		}
 		case MissionType::InvestigateBuilding:
 		{
+			//prevent double update when vehicle was already at tgt building at time of alert
+			if (v.wasAlreadyAtTgtBuilding)
+			{
+				v.wasAlreadyAtTgtBuilding = false;
+				return;
+			}
 			if (finished && v.owner == state.getPlayer() && v.currentBuilding->detected)
 			{
 				v.currentBuilding->decreasePendingInvestigatorCount(state);
--- a/game/state/city/vehiclemission.cpp
+++ b/game/state/city/vehiclemission.cpp
@@ -927,6 +927,26 @@ Reachability VehicleTargetHelper::isReachableTarget(const Vehicle &v, Vec3<int> 
 	}
 }
 
+Reachability VehicleTargetHelper::isReachableForRecovery(const Vehicle& v, Vec3<int> target) {
+	
+	auto &map = *v.city->map;
+	auto targetTile = map.getTile(target);
+
+	// Check if target tile has no building parts permanently blocking it
+	for (auto &obj : targetTile->ownedObjects)
+	{
+		if (obj->getType() == TileObject::Type::Scenery)
+		{
+			auto sceneryTile = std::static_pointer_cast<TileObjectScenery>(obj);
+			if (sceneryTile->scenery.lock()->type->isBuildingPart)
+			{
+				return Reachability::BlockedByBuilding;
+			}
+		}
+	}
+
+}
+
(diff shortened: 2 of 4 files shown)


kgd192:
Fix #484 (Partially)

Recovery Vehicles could not reach other crashed vehicles since their tiles wher marked unreachable due to having no landing Pad, this is now fixed.
Additionaly, AI vehicles which gave up on routefinding during their way home do now crash to initiate recovery by the AI.
No Fixed were attempted regarding the issue that destroyed vehicles are not removed correctly

FilmBoy84:
Much appreciated on your continued PRs 🍻 

I apologise for the continued Linting errors that stop us merging - I've asked another member of the team to look into it when they are able.

I agree on the fix for #1073 and can also not see a better solution at present

kgd192:
> I apologise for the continued Linting errors that stop us merging - I've asked another member of the team to look into it when they are able.

No worries it will be figured out eventually.

For the #484 , One addition I could think of is to give the vehicles a certain time they have to be in a "lost can not find way home state" before they are flagged as damaged and eventually recovered. I honestly didn't check (if even feasible) how the original game handled stuck vehicles.

FilmBoy84:
> For the #484 , One addition I could think of is to give the vehicles a certain time they have to be in a "lost can not find way home state" before they are flagged as damaged and eventually recovered. I honestly didn't check (if even feasible) how the original game handled stuck vehicles.

From what I remember of last time i looked into this: The original game differs in that it only generates missions from point a to b (typically garage to garage or hanger to hanger) rather than the variety of missions we have as optional in OpenApoc (for example via the "Realistic Transportation" feature)

If the target garage or hanger is destroyed whilst the mission is in progress, the vehicle is simply rerouted to it's point of origin or an alternative friendly garage or hanger to the original target within 20 tiles (decision is RNG based with a bias to minimal pathfinder cost)

Given OpenApoc users may be playing without "Realistic Transportation" enabled, that should be replicated for those users.

For users with "Realistic Transportation" enabled we have a few more options as follows

* Behaviour as per above (OG)
* Create entirely new mission, changing direction accordingly
* Attack the unit or organisation that caused damage to the target tile until target destroyed, vehicle destroyed, or it is forced to crash or return to original point of origin
* Choose to "Raid" the organisation (or allies of the organisation) that attacked the target garage or hanger
* Crash the vehicle (maybe they just had a traffic accident) and allow the recovery mechanic to order a recovery unit to collect it and return to the parent organisation

kgd192:
> * Crash the vehicle (maybe they just had a traffic accident) and allow the recovery mechanic to order a recovery unit to collect it and return to the parent organisation.

That was my thought process as well that the vehicle did basicaly break down at the side of the road and then be recovered so that the city would not fill up w/ stuck vehicles.

Otherwise if their current mission is canceled the current behavior that they simply return home remains.

But I really like the Idea to expand on that and create follow on missions depending on why the mission was canceled.

FilmBoy84:
Getting a linting error still

Are you running a CLANG format pass with CLANG / LLVW 14 before submitting each commit?

`for /R forms %%f in (*.cpp,*.h) do clang-format -i "%%f"
for /R framework %%f in (*.cpp,*.h) do clang-format -i "%%f"
for /R game %%f in (*.cpp,*.h) do clang-format -i "%%f"
for /R library %%f in (*.cpp,*.h) do clang-format -i "%%f"
for /R tools %%f in (*.cpp,*.h) do clang-format -i "%%f"

del *.TMP /F /S`

kgd192:
> Getting a linting error still
> 
> Are you running a CLANG format pass with CLANG / LLVM 14 before submitting each commit?

Actually I did not, I will recheck and probably upload the changes to sepperate Branches

FilmBoy84:
> > Getting a linting error still
> > Are you running a CLANG format pass with CLANG / LLVM 14 before submitting each commit?
> 
> Actually I did not, I will recheck and probably upload the changes to sepperate Branches

Thanks, yeah, seperate commits for each fix/change makes things much easier if we have to review at a later point

I'm around for the next hour or so so fingers crossed if we can resolve the linting errors I can review and merge today :)

kgd192:
Have been separated in different Branches / Pull Request #1136 and #1137

kgd192:
Haven been separated into individuall pull requests #1136 & #1137

PR #1132, opened 2023-03-11
Fix message history screen scrolling

### LINKED ISSUE #546: In message history screen message list not auto scroll to newest message at bottom
In message history screen message list not auto scroll to newest message at bottom
(as this was in vanilla) _and if list is long need to list aaaaaallllll over to bottom_
Or it should be reversed where new messages at the top

**ITS FOR BOTH cityscape and for battlescape**

Kurtsley:
Fixes #546. 

It would appear that when the ScrollBar is created in the MessageLogScreen, it uses the original Maximum value of 10 that is assigned by default. When calling the scrollMax function, it was only scrolling down by 10. This was verified by increasing this initial Maximum value and observing the ScrollBar scroll further down the message history screen. This PR fixes this problem by calling the MessageLogScreen::update function to obtain the correct Maximum value before calling scrollMax.

### DIFF
--- a/game/state/city/agentmission.cpp
+++ b/game/state/city/agentmission.cpp
@@ -386,11 +386,11 @@ void AgentMission::start(GameState &state, Agent &a)
 					                                  {&state, a.shared_from_this()}));
 					a.recentlyHired = false;
 				}
-				if (a.recentryTransferred)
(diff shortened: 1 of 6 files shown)


Kurtsley:
Closing because I screwed up... Disregard.

PR #1133, opened 2023-03-14
Fix removing passenger module left excess passengers on vehicle

### LINKED ISSUE #694: [Vehicle Equipment] Adding lots of agents to craft and removing passenger modules allows them to stay in craft
If you add lots of passengers to an aircraft by adding extra passenger modules, then remove those passenger modules, the craft still allows the extra passengers to stay assigned to the craft

In the OG, excess passengers were automatically removed from their assignment to the craft starting from the bottom of the list and moving up

image

Kurtsley:
Fixes #694. My OG experience differs from what is explained in the issue. When trying to remove the passenger module with extra soldiers on board, the game throws up a dialog box explaining that you can't remove the module while in use.

Example from OG:
xcomremove

With that being the case, this PR implements this in OpenApoc by checking if the new number of passengers will be higher than the max passengers when the module is removed. If this is the case, the same dialog box is shown.

Example from OpenApoc:
openapocremove

### DIFF
--- a/game/ui/base/vequipscreen.cpp
+++ b/game/ui/base/vequipscreen.cpp
@@ -51,10 +51,12 @@ VEquipScreen::VEquipScreen(sp<GameState> state)
 	    paperDollPlaceholder->Location, paperDollPlaceholder->Size, EQUIP_GRID_SLOT_SIZE);
 
 	// when hovering the paperdoll, display the selected vehicle stats
-	paperDoll->addCallback(FormEventType::MouseEnter, [this](FormsEvent *e [[maybe_unused]]) {
-		highlightedVehicle = selected;
-		VehicleSheet(formVehicleItem).display(selected);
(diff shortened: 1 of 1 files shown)


game/ui/base/vequipscreen.cpp
```diff
@@ -277,6 +278,20 @@ void VEquipScreen::eventOccurred(Event *e)
 		    std::dynamic_pointer_cast<VEquipment>(this->selected->getEquipmentAt(mouseSlotPos));
 		if (equipment)
 		{
+			// Check if a passenger module can be removed
+			if (this->highlightedVehicle->getPassengers() >
+			    (this->highlightedVehicle->getMaxPassengers() - 4))
```
Jarskih, line 283: You can replace the hardcoded `4` with something like `equipment->type->passengers`

PR #1135, opened 2023-03-25
Fix Issue #749

### LINKED ISSUE #749: Destroyed Terrain May Not Be Actually There 100%
I noticed terrain that is destroyed appears to be repaired when the day changes, but something in the game or save file remembers what was destroyed and re-destroys it on a game load, even immediately after saving the game. Also, city walls that get destroyed are still able to be driven through by ground vehicles (or at least the Griffon for now) even after they appear to be repaired. Flying vehicles don't seem to be able to do this, however. Please see the following videos: https://youtu.be/hQ__bNFvIWM     https://youtu.be/9xRrLQN6AcY

Roads/ramps seem to be affected as well as my Griffon wouldn't travel on one of the ramps to get up to a higher level road.

This was done in OpenApoc-v0.2.0.2-12-g8f8b94c1 with V6.1 of EWM. I have also attached a save file to show this.

[save_City Wall Bug.zip](https://github.com/OpenApoc/OpenApoc/files/3592772/save_City.Wall.Bug.zip)

kgd192:
Fix #749 
Fix reset destroyed state of tile

### DIFF
--- a/game/state/city/scenery.cpp
+++ b/game/state/city/scenery.cpp
@@ -1641,6 +1641,7 @@ void Scenery::repair(GameState &state)
 	}
 	damaged = false;
 	falling = false;
+	destroyed = false;
 	if (tileObject)
 	{
 		tileObject->removeFromMap();


FilmBoy84:
Thanks for the fix
Good spot kgd192
🍻

Issue #1072, opened 2021-10-27
Proximity Weapons (Proximity Mine and Boomeroid etc.) Have ambiguous wording and controls for use


FilmBoy84:
**Describe the bug**
On a recent stream it became clear that proximity mines such as the MarSec Proximity Mine and Alien Boomeroid use ambiguous terminology when priming them for the intended use

Currently they can be set to "Activates Now"  - which the game treats as exploding on landing (proximity trigger disabled causing it to operate as a grenade that detonates on impact) and "Delay = 0"  - which means that the device will explode immediately on detecting something within the proximity range set by the player (the expected behavior for most users, but NOT the default)

This is very confusing as "Activates now" suggests that it will activate with the proximity trigger enabled immediately, not that it will detonate on landing/dropping. "Delay = 0" equally doesn't really explain that it will wait for proximity trigger and then immediately explode

It would be wise to think of a different naming convention for these two modes
I propose that we have something like "Detonates Now - Proximity Disabled" and "Detonates on Proximity Trigger" as replacements for "Activates Now" and "Delay = 0" respectively

**To Reproduce**
See stream from 2:29:32 onward where Quicktime tests a stun mind (EWM mod, though the behaviour is the same with the vanilla mine and boomeroid)

This test also features an instance of #1057 as a thrown mine goes off the world (despite visually being stuck on the tile in the place it should be)

https://www.twitch.tv/videos/1184682999?t=02h29m32s
1. Select Mine in agent inventory
2. Click to set detonation conditions
3. "Activates Now" is set by default and "Delay = 0" can be set via scroll
4. These do not clearly tie to the behaviour of the device when thrown/dropped

**Screenshots**
![image](https://user-images.githubusercontent.com/33913088/139122030-794ee6ef-49c6-43f1-828a-204ca31fa5da.png)
![image](https://user-images.githubusercontent.com/33913088/139122225-01ed9acb-0c95-4c00-a9fb-5f6cf2c0293a.png)

FilmBoy84:
This has been partially resolved in #1145 that fixes the Delay = 0 aspect of this issue

However, the ambiguous naming will likely warrant a new string to resolve - as the translation structure and mechanism is something being worked on by JonnyH we cannot resolve that aspect yet until we are confident there will not be a conflict with his work

Therefore, this issue will remain open until fully resolved

PR #1145, opened 2023-03-28
Fix the confusing naming of proximity mine delay

Kurtsley:
Addresses #1072.

I'm putting this PR forward as a step in the right direction, but more work has to be done. As discussed in issue 1072, the wording of the proximity mines is a bit confusing. This changes the wording of Activates Now to Detonates Now.

Ex.
det1

 The issue of delay == 0 was some sort of rounding or math error and has been corrected. The delay slider now reads correctly in seconds and counts up by one quarter second until 7.5. To match the original game timer length of 7.5 seconds, I had to add one to the max value of the delay slider. The delay slider now maxes out at 30 instead of 29.

Ex.
det2

I noticed that the "Activates Now" string is present in the ufo_string translation files and adding another phrase without translating it could be a problem. This was not tested in another language, so using the tr() function with the new string may cause a crash. If this change is permanent, "Detonates Now" will probably have to be translated and added to those files. I left a comment that addresses this.

### DIFF
--- a/data/forms/battle/battle_rt_tab3.form
+++ b/data/forms/battle/battle_rt_tab3.form
@@ -62,7 +62,7 @@
       <scroll id="DELAY_SLIDER">
         <position x="263" y="75"/>
         <size width="116" height="15"/>
-        <range min="0" max="29"/>
+        <range min="0" max="30"/>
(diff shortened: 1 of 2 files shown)


FilmBoy84:
@JonnyH as you are working on translation updates, are you please able to advise how those changes will interact with this Pull please?

Kurtsley:
I'm going to remove the new text as I feel that the timer change is important in itself. I I will leave a comment about changing the phrase in the future.

Issue #1127, opened 2023-02-25
Clicking an icon does not bring up the equip menu


Kurtsley:
In the original game, when the icon of a vehicle or player is clicked, you are taken to the equip screen. Clicking the name will highlight the vehicle or player and not open the equip menu. In OpenApoc currently, clicking either side will only highlight the vehicle or player.

For example, clicking the icon on the left opens the equip menu for the vehicle, while clicking the name highlights the vehicle.
![apocscreen](https://user-images.githubusercontent.com/73447098/221334609-1a599c4b-8eb9-4871-8626-792127c7f922.png)

kgd192:
The Equip Screen now opens on a RIGHT Clicks, looks like quality of life feature

PR #1146, opened 2023-03-30
Add option to left click icon to open equip menu

Kurtsley:
Addresses #1127 by adding an option to open the agent or vehicle equip screen by left clicking the icon. This restores the functionality from the original game where it was possible to right OR left click the icon to open the equip menu. To be clear, this option does not remove the ability to open the equip menu using right click, it just adds the left click and is disabled by default.


Kurtsley:
Actually, should this be lumped in with the vanilla cityscape controls?

FilmBoy84:
> Actually, should this be lumped in with the vanilla cityscape controls?

Probably an idea to
The options menu is not the best ordering at the moment
But if we try and get some coherency going forward it will save the job of doing a massive reordering later :D

Kurtsley:
I will look into that reorder. Probably a separate PR though.

FilmBoy84:
Agreed, best as another PR

Just a heads up, this now has conflicts with the changes added by #1150 on merger

PR #1149, opened 2023-03-31
Add mouse scroll wheel changes vertical levels in battlescape

Kurtsley:
This PR adds an option that lets the mouse scroll wheel change the vertical level in the battlescape. Since this wasn't in vanilla, I decided to disable it by default but would like some input on whether this should be the default option.


FilmBoy84:
The mouse scroll is not used anywhere in Original Game IIRC

As it's incredibly handy to move up and down with this feature, i'm tempted to say it should be ON by default with the option to turn it off if you really don't like it

Very much a QoL feature that's not invasive and doesn't conflict with the Original Game so having it on by default is the way forward IMHO

PR #1151, opened 2023-04-01
Remove options menu from main menu

Kurtsley:
Removes options menu from main menu. Since I didn't get a ton of feedback on this issue, I'll just drop this here in case a consensus is reached. Code commented out instead of deleted to seem more temporary.

### DIFF
--- a/data/forms/mainmenu.form
+++ b/data/forms/mainmenu.form
@@ -33,18 +33,19 @@
         <size width="300" height="32"/>
         <font>smalfont</font>
       </textbutton>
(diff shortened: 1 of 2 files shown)


FilmBoy84:
Looks good to me - I agree, commenting out rather than fully removing is the best approach

Although, it's probably an idea to add a note to explain why it's commented out. Something like

`// options menu commented out until a more useful purpose for it is returned to the game`

This way we don't lose why it's not visible further down the line

PR #1156, opened 2023-04-04
Fix transaction screen scrollbar multiple items per click

Kurtsley:
In the original game, the horizontal scroll bars in the transaction screen moved one item at a time when clicked. In OpenApoc, the scrollbars move depending on the amount of items in the list. This PR fixes this issue by separating the vertical and horizontal scrollbars in the transaction screen and setting the button click to move add or remove exactly one item each time.


FilmBoy84:
Thanks for doing this 🍻 

I was reminded on my stream on Sunday of it and we just had another report in saying it was a problem

PR #1162, opened 2023-04-07
Fix enforce cargo limits not working in cityscape

Kurtsley:
When enforce cargo limits is turned off, it only counts cargo coming back from a mission. This PR extends the option to include picking up cargo in the cityscape. When off, a single vehicle can pick up an unlimited amount of cargo from a building. I'm looking for some feedback on how exactly this should work. At the moment, the vehicle that is carrying the extra cargo shows exactly how much cargo they have on board.

Example:
cargo
You can see the number shows a larger number than is possible for the vehicle. 

While this makes sense in a way, I wonder if it would be better to cap the cargo number to the max capacity of the vehicle.

Secondly, I took the liberty to limit this ability to only vehicles owned by the player. If not, a small transtellar transport will be able to pick up everything in one trip. Thoughts?

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -1633,11 +1633,15 @@ void Vehicle::provideServiceCargo(GameState &state, bool bio, bool otherOrg)
 			continue;
 		}
 		// How much can we pick up
-		int maxAmount = std::min(spaceRemaining / c.space * c.divisor, c.count);
+		int maxAmount = (!config().getBool("OpenApoc.NewFeature.EnforceCargoLimits") && this->owner == state.getPlayer())
+		                    ? c.count
+		                    : std::min(spaceRemaining / c.space * c.divisor, c.count);
+
 		if (maxAmount == 0)
 		{
 			continue;
 		}
+
 		// Here's where we're going
 		if (!destination)
 		{
--- a/game/ui/general/vehiclesheet.cpp
+++ b/game/ui/general/vehiclesheet.cpp
@@ -6,6 +6,7 @@
 #include "game/state/rules/battle/damage.h"
 #include "game/state/tilemap/tilemap.h"
 #include <list>
+#include <framework/configfile.h>
 
 namespace OpenApoc
 {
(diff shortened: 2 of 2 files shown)


FilmBoy84:
Agreed on limiting it only to the player controlled vehicles

From an optimum design point of view we'd want the following for the player

If the option for Enforce Cargo Limits is OFF/Deselected/Cross

* Cargo limits for transport are not obeyed, regardless of cargo capacity of vehicle
* Cargo limit is displayed in vehicle equipment screen / profile as the total value when all cargo modules and any craft native capacity are added together

If the option for Enforce Cargo Limits is ON/Selected/Tick

*Cargo limits for the transport are obeyed up to the total cargo capacity of the vehicle
* Cargo limit is displayed in vehicle equipment screen / profile as the total value when all cargo modules and any craft native capacity are added together

I believe that later (ON) option is what the computer controlled vehicles use with transtellar often sending multiple transports for a delivery to be made...

Kurtsley:
So just to be clear, in the above example, with cargo limits NOT enforced it should read 50/50+? And you want to add an option to allow the computer controlled vehicles to ignore these limits?

FilmBoy84:
> So just to be clear, in the above example, with cargo limits NOT enforced it should read 50/50+?

Yep, that sounds good - does it seem readable to most end users to you?

>  And you want to add an option to allow the computer controlled vehicles to ignore these limits?

If you wish to add one yes, personally, i'm wondering if we dare mess with the computer organisations - at present, they work

But equally, I can see players wanting the same rules for the computer as they have for the player

Then again, if we separate the options into something like "Enforce Cargo Limits for Player" and "Enforce Cargo Limits for Organisations" it would give the end user maximum control over the way things play

I'm just weary about "unforeseen circumstances" if we change the other orgs 😅 

What are your thoughts on this?

Kurtsley:
Personally, I agree with the first part and don't think we should mess with the other orgs. As far as the plus symbol, it might make more sense to replace the max cargo number with an infinity sign or simply the word infinite or inf. It would read like 1232/infinite or 1232/inf.

FilmBoy84:
I'm trying to think of the symbols available in the character sheet
(Don't believe we have moved over to True Type Font text yet)

The issue with using Inf. or Infinite is that, unless there is already a suitable text string, we'd be looking at adding another one which may cause issue with translations (One reason I haven't merged #1145 until we hear from JonnyH regarding his updates)

Personally, given the small space available in Apoc screens I'd love the `1234/Inf.` format but again, if no string for this, but one already existent for Infinite, use Infinite

FilmBoy84:
Addition: The font sets are located here

https://github.com/OpenApoc/OpenApoc/tree/master/data/fonts

Kurtsley:
Or we could remove the limit number entirely and just list the amount of cargo on board.

Kurtsley:
I screwed this branch up again... I'm going to open an issue to discuss this further and put in a PR when it is decided what to do.

PR #1165, opened 2023-04-08
Fix cargo limits not being enforced in the cityscape

### LINKED ISSUE #1164: Enforce cargo limits does not apply to craft in cityscape
As discussed before, cargo limits are still enforced in the cityscape even with the "enforce cargo limits" option turned off. My new idea is to completely remove the max cargo number when cargo limits are not being enforced.

For example, this is what an unloaded vehicle will look like:
item1

Any cargo added will simply be shown as the current cargo on board:
item2

Kurtsley:
Fixes #1164.

When enforce cargo limits is turned off, it only counts cargo coming back from a mission. This PR extends the option to include picking up cargo in the cityscape. When off, a single vehicle can pick up an unlimited amount of cargo from a building. This is limited to only player owned vehicles and not the transtellar transports in the city. When cargo limits are disabled, the vehicle will only show the cargo that is onboard and no max cargo number.

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -763,29 +763,29 @@ class GroundVehicleMover : public VehicleMover
 								vehicle.goalPosition.z = vehicle.position.z;
 							}
 							else
-							    // If we're on flat surface then first move to midpoint then start
-							    // to
-							    // change Z
-							    if (fromFlat)
-							{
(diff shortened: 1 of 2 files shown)


FilmBoy84:
It's taken a week to test, my apologies

But yes, this works nicely - and computer vehicles work unaffected by it

Great stuff sir! Merging

Issue #997, opened 2021-02-21
[Core] Assumed Ticks Per Second / Frames Per Second is not unified across all parts of the codebase


FilmBoy84:
At present, several areas of the code deviate from the original game in assuming various ticks per second values; often these are tied to erroneous frames per second values

Since the Alpha 0.2 series of OpenApoc versions, the frame rate has been fixed at 60fps as opposed to the original game's assumed 18/36.

However, ticks per second is still all over the place with some parts of the code, for example, weapons ballistics assuming 144TPS on a 144Hz Monitor - this causes very real problems with the ballistics calcs making matching the original game impossible as many values are based on integers which do not permit decimal values

Some time ago tentative discussion was made on discord about fixing this

The ideal would be to have the game assume 180 ticks/sec for everything - this is because 18/36/60 based calculations from the original game can be easily adapted into whole number multiplications/divisions when calculating certain game mechanics and we can keep using int instead of float - it's a lot less work to update the codebase and fix any calculations if 180 is the unified value

Though this would present an issue in meeting the 5ms optimisation target on none-desktop hardware

Whilst I'm inclined, given that any mobile forks of OpenApoc would need massive changes to the UI as well as other optimisations, to go "leave it to the mobile devs to sort out" and shift to the ideal of 180TPS. We should probably try not to break compatibility if a "better" solution can be found

Please see discussion on discord for further details:
https://discord.com/channels/142798944970211328/143424162998910976/642830538117349385
https://discord.com/channels/142798944970211328/143424162998910976/702306923894145025

FilmBoy84:
Further discussion as of April 2023 can be found here on the discord thread

https://discord.com/channels/142798944970211328/1093236322837217490

Hambones | Josh has also kindly made a Trello page here

https://trello.com/b/tCJi0GrR/issue-997-related-code-locations

PR #1166, opened 2023-04-10
Decouple fps tps

Hambones82:
This pull request attempts to address https://github.com/OpenApoc/OpenApoc/issues/997.  First, in framework.cpp, a new whileloop is added around currentstage.update() which runs on a ticks per second basis, rather than being in lockstep with stage.render().  The code also includes various changes related to ticks_per_second and related constants, throughout a number of files.  The code also moves the ticks_per_second constants into framework.h, in order to allow the state update loop to run on a ticks_per_second basis.  This assumes a 1:1 correspondence between in-game seconds at the slowest speed and real-time seconds.  Later, we might consider moving the time constants back into gametime.h and having a single number that is ticks per real second (or something), as well as ticks_per_sim_second, with a relation between the two.

Re the change to doodad, it now has a magic number which is bad.  I'm not sure what it should be, though...  It was originally ticks_multiplier, which I thought I knew what it meant but now I'm not so sure.

### DIFF
--- a/framework/framework.cpp
+++ b/framework/framework.cpp
@@ -311,9 +311,8 @@ void Framework::run(sp<Stage> initialStage)
 	auto target_frame_duration =
 	    std::chrono::duration<int64_t, std::micro>(1000000 / Options::targetFPS.get());
 
-	//to do: replace magic number 36 with a constant -- this is for testing purposes only.
-	auto target_tick_duration = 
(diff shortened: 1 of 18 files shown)


FilmBoy84:
Regarding Doodad.cpp do we want to hardcode this value of four if we move to a variable simulation rate?

Hambones82:
You are right.  I think it will make sense to create a new constant for doodads in doodad.h.  I'll update it.

PR #1172, opened 2023-04-16
Fix CTD when selecting X-Com Vehicle Tab

### LINKED ISSUE #1171: CTD When Selecting X-Com Vehicle Tab
**Describe the bug**
When Selecting the X-Com Vehicle List in CityView the following exception is created
image

**To Reproduce**
1. Load a savefile or Start a new game
2. Press the "X-Com Vehicle Tab" Button

**Expected behavior**
View should display the X-Com Vehicle List

**Savegame**
N/A

kgd192:
Fix #1171 the CDT part but now the UI incorrectly displays two green "X"'s instead of One.
image

Maybe @Kurtsley could have a look into this

### DIFF
--- a/game/ui/components/controlgenerator.cpp
+++ b/game/ui/components/controlgenerator.cpp
@@ -155,6 +155,10 @@ VehicleTileInfo ControlGenerator::createVehicleInfo(GameState &state, sp<Vehicle
 		t.headedHome = v->missions.back().targetBuilding == v->homeBuilding ||
 		               v->missions.back().type == VehicleMission::MissionType::OfferService;
 	}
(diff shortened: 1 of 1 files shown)


Kurtsley:
I can check later. The second X is only drawn when headedHome is true.

kgd192:
headedHome defaulted to true therefore was always displayed is fixed with the last commit

kgd192:
Not required in a Release Build (see #1171) however still recommend to keep the change to avoid exceptions in the future since calling back() on a empty list is not defined (reference https://cplusplus.com/reference/list/list/back/)

Issue #1131, opened 2023-03-11
Long messages in the message history menu are cut off.


Kurtsley:
**Describe the bug**
When a message in the message history menu is long enough to require multiple lines, the message is cut off on the top and bottom.

Example:
![menucutoff](https://user-images.githubusercontent.com/73447098/224473858-69635215-153c-4f37-bb2e-ea48af293259.png)

**To Reproduce**
1. Have any of your vehicles be destroyed by a UFO.
2. Observe the message in the message history menu.

**Possible Solution**
I haven't played the original game enough to know for sure, but it seems this was not an issue because the messages were shorter. In the following example, the "Vehicle Destroyed" message prefix is not present in the original game.

Example:
![menuorig](https://user-images.githubusercontent.com/73447098/224474252-4655d615-f8db-4c51-9b2b-a38a5b5937de.png)

Perhaps the simple solution could be implemented first while a more permanent solution is worked on. I spent some time looking at changing the height of the messages, but have been unable to sort it out.

Kurtsley:
My apologies, I just noticed this already been reported. I would still like feedback on shortening the messages though.

FilmBoy84:
> My apologies, I just noticed this already been reported. I would still like feedback on shortening the messages though.

Agreed, mimicking the Original by culling the extra prefix is the way forward

Ideally, we should be updating the forms to allow text wrapping as a long term solution - but for now, Original functionality should be priority :)

Kurtsley:
I have addressed the vehicle destroyed message in a PR because it is the most common. There could be a situation where a vehicle recovered message will become too long and cause issues. I think this issue should be left open until a solution like the one above is implemented.

FilmBoy84:
Resolved by #1175

PR #1175, opened 2023-04-19
Fix multi-line messages being cut-off

Kurtsley:
Addresses #1131.

This is the solution that I have come up with so far, but I would like to get some feedback. The ideal solution would be to figure out a way to do this without editing the listbox file. While I haven't found any obvious bugs or changes to the game, editing this file opens up a can of worms that I feel needs more testing. 

Here are the changes:

---
```
	case Orientation::Vertical:
		ctrl->Size.x = (scroller_is_internal ? scroller->Location.x : this->Size.x);
		ctrl->Size.y;
		break;
```
In the listbox.cpp file, this section renders the individual items in the list. In the vertical orientation, this set the y value of the items to a variable called ItemSize. The entire point of this is to be able to have multiple items of different sizes in a listbox, attaching the ItemSize variable to the y value of the list items made this impossible in my view. This became clear when changing the ItemSize value in the messagelogscreen.cpp file. Even in a loop like so:
```
	auto listbox = menuform->findControlTyped<ListBox>("LISTBOX_MESSAGES");
	for (EventMessage message : state->messages)
	{
                ItemSize = HEIGHT;
		listbox->addItem(createMessageRow(message, state, cityView));
	}
```
This is just a rough example of what was tried, the point is I tried to change the ItemSize value while looping through each message. The result was not correct, whatever value was last, e.g. a value of 44 for the double size message, was the item size for the entire message history list. Since the listbox is rendering every item according to the ItemSize value, having multiple sized items did not work. 

Update func:
```
		if (ItemSize == 0)
		{
			for (const auto &i : Controls)
				scrollerLength += i->Size.y;
		}
		else
		{
			scrollerLength += Controls.size() * ItemSize;
		}
```
In the update function of the listbox.cpp file, this code determines the total length of the scroller. There is even a helpful comment:
```
// assume item sizes are variable if ItemSize is 0 (ie: need to calculate manually)
```
In my experience, if the ItemSize is set to 0, nothing shows up in the list. I'm not sure where this is used, so I changed it to calculate by Size.y every time. Without this change, the scroller size was not correct.

The messagelogscreen.cpp changes are more straightforward. The HEIGHT variable is changed depending on the size of the message.text string. There should be a better way to do this as I picked 55 through trial and error. Maybe some way to tell if the string has wrapped to the next line. The location button location is also changed depending on if the message.text string size is over 55. The important part is last, the size of the control is changed to the HEIGHT value. Since the listbox is rendering each item according to the ctrl->Size.y value, each item can be a different size in the list.

As stated earlier, I have not noticed any problems with any other vertical list in the game. That being said, I haven't played the game very much.

Sorry for the rant. Comments are appreciated!

### DIFF
--- a/forms/listbox.cpp
+++ b/forms/listbox.cpp
@@ -62,7 +62,7 @@ void ListBox::onRender()
 				{
 					case Orientation::Vertical:
 						ctrl->Size.x = (scroller_is_internal ? scroller->Location.x : this->Size.x);
-						ctrl->Size.y = ItemSize;
+						ctrl->Size.y;
 						break;
 					case Orientation::Horizontal:
 						ctrl->Size.x = ItemSize;
@@ -202,15 +202,8 @@ void ListBox::update()
 		{
 			case Orientation::Vertical:
 			{
-				if (ItemSize == 0)
-				{
-					for (const auto &i : Controls)
-						scrollerLength += i->Size.y;
-				}
-				else
-				{
-					scrollerLength += Controls.size() * ItemSize;
-				}
+				for (const auto &i : Controls)
+					scrollerLength += i->Size.y;
 				scrollerLength = scrollerLength > Size.y ? scrollerLength - Size.y : 0;
 				break;
 			}
--- a/game/ui/general/messagelogscreen.cpp
+++ b/game/ui/general/messagelogscreen.cpp
@@ -71,7 +71,7 @@ sp<Control> MessageLogScreen::createMessageRow(EventMessage message,
 {
 	auto control = mksp<Control>();
 
-	const int HEIGHT = 21;
+	int HEIGHT = message.text.size() > 55 ? 44 : 21;
 
 	auto date =
 	    control->createChild<Label>(message.time.getShortDateString(), ui().getFont("smalfont"));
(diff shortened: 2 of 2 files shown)


FilmBoy84:
I agree with your solution, it does a rather nice job of rectifying the message list issue and having objects of different sizes in the listbox will definitely be required for some purposes (I have a nagging feeling equipment screens use them along with the ItemSize 0)

I'm probably not going to be able to build and test myself before weekend - however, can you do the following please?

* Check that items in the economy (buy / sell / hire / hire / transfer) are still working correctly

* Download the Extended Weapons Mod from the github repot and see if it resolves things like Agent Names, Item Names, Vehicle Names, etc. being trimmed out of bounds

* Check that all text in the vehicle equipment and agent / vehicle assignment menus is correct

Many thanks as always for your work :beers:

Kurtsley:
Every appears to be working as intended, I didn't notice anything out of the ordinary. As for the extended weapons mod, none of the long item names issues have been fixed. The listbox change allows the ability to fix some of those issues, but this PR only addresses the message history screen. Each problem will have to be solved individually as it requires editing the forms of that particular screen. I may look at the agent names next as I noticed that there is sometimes a space between the name and the skill label. That PR may fix the issue of the long agent names being cut off in the weapons mod.

FilmBoy84:
Thanks for checking

If other screens require their own fixes, I'll say those should be via separate PRs just so we can keep track of each change made

This fix appears to be working perfectly - I'll merge now

PR #1194, opened 2023-04-22
Empty base will now be destroyed with message

### LINKED ISSUE #1067: New (Unfinished) Bases improperly spawn a defence mission when attacked instead of being destroyed
My new bases get attacked before I've finished any facilities. Which means, I can't station any units there. The attack on my empty base triggers an error message ("WTF, no player units found"), the battle begins without any units, which is unwinnable, and I lose the base. 
I can even fly units to the building but they don't appear as available for battle.
Seems like a bug. Maybe either let me station units at empty new bases, or have aliens not invade empty bases.

Kurtsley:
Partially fixes #1067.

Every case of an empty base being attacked was covered except the case in Building::detect. This adds a check for an empty agent list and automatically destroys the base if true. I also added a city notification to alert the player that the base has been destroyed. I believe this adds the same functionality as the original game.

### DIFF
--- a/framework/options.cpp
+++ b/framework/options.cpp
@@ -74,7 +74,6 @@ void dumpOptionsToLog()
 	dumpOption(optionPauseOnVehicleRefuelled);
 	dumpOption(optionPauseOnNotEnoughFuel);
 	dumpOption(optionPauseOnUnauthorizedVehicle);
-	dumpOption(optionPauseOnBaseDestroyed);
 	dumpOption(optionPauseOnHostileSpotted);
 	dumpOption(optionPauseOnHostileDied);
 	dumpOption(optionPauseOnUnknownDied);
@@ -292,9 +291,6 @@ ConfigOptionBool optionPauseOnNotEnoughFuel("Notifications.City", "NotEnoughFuel
                                             "Not enough fuel to refuel vehicle", true);
 ConfigOptionBool optionPauseOnUnauthorizedVehicle("Notifications.City", "UnauthorizedVehicle",
                                                   "Unauthorized vehicle detected", true);
-ConfigOptionBool optionPauseOnBaseDestroyed("Notifications.City", "BaseDestroyed",
-                                            "X-COM base destroyed by hostile forces.", true);
-
 ConfigOptionBool optionPauseOnHostileSpotted("Notifications.Battle", "HostileSpotted",
                                              "Hostile unit spotted", true);
 ConfigOptionBool optionPauseOnHostileDied("Notifications.Battle", "HostileDied",
--- a/framework/options.cpp
+++ b/framework/options.cpp
@@ -74,6 +74,7 @@ void dumpOptionsToLog()
 	dumpOption(optionPauseOnVehicleRefuelled);
 	dumpOption(optionPauseOnNotEnoughFuel);
 	dumpOption(optionPauseOnUnauthorizedVehicle);
+	dumpOption(optionPauseOnBaseDestroyed);
 	dumpOption(optionPauseOnHostileSpotted);
 	dumpOption(optionPauseOnHostileDied);
 	dumpOption(optionPauseOnUnknownDied);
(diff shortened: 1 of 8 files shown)


kgd192:
Does this show both the news ticker as well as the notification window (see below), when the base is destroyed?

*Notification Windows*
image

Kurtsley:
I may have misinterpreted the meaning of "notification." I've never had a second base destroyed in the original game.

Kurtsley:
This is from ufopedia (can be wrong I know):


An [undefended](https://www.ufopaedia.org/index.php/Tactical_Combat_Missions_(Apocalypse)#Base_Attack) X-Com base will be destroyed immediately when [attacked](https://www.ufopaedia.org/index.php/Base_Defense_(Apocalypse)) by hostile ground forces (alien craft infiltration or raid). Move/Fly armed agent(s) to the building immediately after purchase - living quarters does not need to be completed since they are [visitors](https://www.ufopaedia.org/index.php/Base_Defense_(Apocalypse)#Building_Garrison_And_Transfers). If the new base is under attack and no facilities have been placed, a base defense battlescape will only contain the bare corridors and the access lift.

In the save I am working with, there is a vehicle with agents on board parked at the base that is being constructed. According to this post, instead of automatically destroying the base there should be a base defense mission in the empty base with the agents in the craft. If this is true, this PR will need to be completely changed. I think the system that decides if a vehicle is parked in a building is not recognizing when a vehicle is parked in a base under construction.

FilmBoy84:
> This is from ufopedia (can be wrong I know):

It means that in the OG, if a base has no agents, it will be destroyed immediately

HOWEVER

If agents are stationed in the building above (all X-COM bases are basement structures) then a mission will be spawned with only the access lift (and any corridors forming the shape of the base as it appears in the Bases screen)

Basically, if you don't want your base insta-destroyed in the OG, you can station agents there until a facility is complete and it will just generate a map with a load of empty corridors allowing you a chance to fend off the invaders and avoid an insta-nuke of your freshly purchased basement 😅

Kurtsley:
This is an interesting issue. I took the agents that were sitting in the vehicle at the base under construction and put them in the building. This spawned a base defense but the agents still did not show up to defend. So in OG they would be used in the base defense whether they were in the vehicle or in the building? It appears that they don't defend the base when on either side. It is strange because the base defense should be triggered if the list of agents in the building is not empty, and yet they aren't there when the battle starts. Will look at some debug crap.

FilmBoy84:
That's correct, final release OG behaviour is that they will appear in the base defence regardless of whether they are assigned to vehicle at the building or assigned to the building itself - the building pool is collective with no differentiation in the final release

The map will always be access lift and corridor sections out to the pattern of the parent base

The base is destroyed if the agents at the building above lose the combat mission or leave the combat area via the access lift

Kurtsley:
I will try to get the agents to appear in the defense and change the notification to a prompt instead. I will consider this done at that point.

Kurtsley:
Alright, this seems to be working better now. To start with, the notification was changed to a messagebox. It displays the name of the destroyed base with the generic base destroyed message:

base

Getting the rest working was fun. I created a new function called occupied in the building file that checks if any agents, or agents in vehicles are present at a building. When a base invasion is triggered, the game will check if the building is occupied or not and destroy the base if empty. If any agents are present in the building, they are added to the list of agents available for the base defense. Likewise, if any agents are on vehicles in the building, they are also added to the list of available agents. Lastly, the original code for checking for agents available for the defense was if agent->homebuilding != base then ignore. This does not work for a second base that is not completed. This was changed to if agent->currentBuilding != base->building then ignore. This allows all agents currently in the building to participate in the defense. Hopefully this doesn't break other base invasion situations, it has only been tested with one save file.

Kurtsley:
I removed the messagebox because I was getting linking errors when running the rebuild all command in VS.

FilmBoy84:
> I removed the messagebox because I was getting linking errors when running the rebuild all command in VS.

Do you want me to merge this commit as is (seems to work on a test I just did) then create a new issue for the message box?

Kurtsley:
Might as well just merge this as is, I don't know what the problem is. It sounds like it could be something simple like forgetting a header file but it doesn't look any different than any other file using messagebox. Hopefully someone else fixes it.

Issue #1195, opened 2023-04-23
Soldiers not calling Taxi Service when unable to proceed on Foot


kgd192:
**Desktop (please complete the following information):**
 - OS:Windows 10 x64 22H2
 - OpenApoc Version 
![image](https://user-images.githubusercontent.com/15697480/233835488-c9c3a743-7ed7-4c4a-ba1c-9241200516a7.png)

**Describe the bug**
When trying to send a soldier to walk to a building, either via the alert screen or the go to building button they will not call a taxi even though the option is enabled and expected per OG behavior

**To Reproduce**
1. Load Save and fast forward and wait for Alert
2. select soldier
3. try to investigate
4. after this failed, try to send via "Go To Building Button" to the building with the alert (now centered)

**Expected behavior**
When unable to reach the target building via the peoples tube network the soldier should call a taxi and use it to reach their destination

**Savegame**
[save_agentOnFoot391.zip](https://github.com/OpenApoc/OpenApoc/files/11303486/save_agentOnFoot391.zip)

**Screenshots**
*Investigate Building*
![image](https://user-images.githubusercontent.com/15697480/233835743-8536c017-c196-4a25-a5cf-ccb76c75e0ba.png)

*go to Building manually*
![image](https://user-images.githubusercontent.com/15697480/233835771-ca9c770d-e768-4579-93f6-1492b738e856.png)

`W void __cdecl OpenApoc::AgentMission::start(class OpenApoc::GameState &,class OpenApoc::Agent &) Implement agent pathing to closest building when in the field and unable to path to building`
`W bool __cdecl OpenApoc::Agent::popFinishedMissions(class OpenApoc::GameState &) Agent Steve Brown mission "GotoBuilding BUILDING_BAKUNIN_BLOCK" finished`
`W bool __cdecl OpenApoc::Agent::popFinishedMissions(class OpenApoc::GameState &) No next agent mission, going idle`

FilmBoy84:
@kgd192 this appears fixed by Kurtsley's PR

Can you confirm please? It seems to work with the save on my system

kgd192:
Will recheck this today, keep you posted

kgd192:
Yes, the being unable to proceed on foot is solved however it does not bring up the message box that all selected units have arrived at the building and whether the investigation should be started or not.

![image](https://user-images.githubusercontent.com/15697480/234349781-a14ca058-9a4b-413c-b142-22ee83e89cad.png)

[...]

PR #1197, opened 2023-04-25
Fix soldiers can call taxi from alert screen

Kurtsley:
Addresses #1195. 

So this is both an actual problem and a control problem. As for the first part, it was not possible for an agent to call a taxi from the alert screen when clicking on the investigate button. This PR fixes that issue. With the option on, all soldiers will now call a taxi to take them from the base to the target building. The second part was when the soldier was stuck at the building halfway to the target. If you hold either control and click the target building, the soldier will call a taxi. I think this is a bit obscure. Maybe a better option would be to alert the user and give an option to call the taxi from a messagebox? I thought about letting the soldier automatically call a taxi if the option is checked, but this would happen no matter how short the distance to the target and would make no sense in some cases.

Edit: On second read, I may be wrong about the ctrl key. From the controls:
```
Cityscape Mouse:

[Ctrl]
 - when giving move orders to Agents, forces Agents to move on foot (never call a taxi), and allows use of personal teleporter
 - when giving move orders to Vehicles, allows manual use of teleporter
 - when giving attack orders forces Vehicles to attack their target (instead of recovering UFOs or escorting owned vehicles)
 - note that in vanilla, attacking owned vehicles was impossible, as such, a feature toggle is required for it to work (on by default)
```

### DIFF
--- a/game/ui/city/alertscreen.cpp
+++ b/game/ui/city/alertscreen.cpp
@@ -3,6 +3,7 @@
 #include "forms/graphic.h"
 #include "forms/label.h"
 #include "forms/ui.h"
+#include "framework/configfile.h"
 #include "framework/event.h"
 #include "framework/framework.h"
 #include "framework/keycodes.h"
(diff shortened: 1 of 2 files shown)


Kurtsley:
Based on reading the controls closer, everything should now work correctly. If the taxi and teleporter options are checked, which they are by default, the following is true:

- Soldiers will automatically call a taxi if the investigate button is clicked in the alert screen and not loaded on vehicle.
- Soldiers will automatically call a taxi if the go to building button is clicked unless control is held. They will still use personal teleporters if control is held.

I don't know if this matches the OG game but it matches the controls page on ufopedia.

Issue #1202, opened 2023-04-26
Mouse List Scrolling Issue


Nurulinde:
This is not really an "Addition" to the game, nor is it really a bug, but rather a humble request for a small change to a small feature of the current game.  Right now, the player can use the Mouse Wheel to scroll down lists in the game, which is awesome... but... when you reach the bottom, an INSANELY loud beeping occurs warning the player they are at the bottom.  I wear headphones and this makes me want to smash my headphones onto the floor every time I reach the bottom.  Yes, I know there are volume control sliders in the Options, but every other sound volume in the game is about right.  Do we really need to have any sound at all when the mouse scrolling reaches the bottom of the list ?  It is not like reaching the bottom is an error or something.

Kurtsley:
Could add this as an optionional feature I suppose.

Kurtsley:
This is a game changer, I may never turn this off now.

makus82:
I like that sound and its not so loud not INSANELY

FilmBoy84:
Resolved as an option in #1204 closing this issue :)

PR #1204, opened 2023-04-26
Add option to disable scrollbar error sound

Kurtsley:
Addresses #1202. 

Adds an option to disable scrolling error sounds. After turning this on, I may never go back.


FilmBoy84:
Nice little optional feature :)

Issue #1208, opened 2023-04-26
Single Tap selection of an Agent Group


Nurulinde:
In OG Apoc, the Player could select/center on any group in the Battlescape by tapping the associated group hot-key (1-6).  In OpenApoc, double-tapping is required.  Is there a way to restore this to the OG functionality ?

FilmBoy84:
This appears to have been a QoL change for allowing easier merger of squads compared to the OG

For example, first press selects the new squad as priority (solid yellow line when pressing 2 with the default 1 squad selected and centred, but leaves squad 1 partially selected with a dotted line, you can then select the merge group button to merge them)

![image](https://user-images.githubusercontent.com/33913088/234605385-5f0eaf23-63cf-48ec-92b4-3018a79a48fd.png)

But I agree, the option to select the Original Game behaviour (off by default) would be a more "friendly" option to allow the end user to choose the older, less useful, behaviour if required

FilmBoy84:
Select with Single Click functionality (off by Default) is now available under More Options thanks to Kurtsley
Closing this issue

PR #1217, opened 2023-05-24
Added the option to select squads with a single click.

Kurtsley:
Addresses #1208.

Adds the option to select squads with a single click. This not quite what the vanilla game did however... In OG Apoc, clicking a squad with the mouse acted as OpenApoc does now while using the number keys instantly zoomed to the squad. In this case, I felt it wasn't necessary to bring this functionality back as we can have the best of both worlds with the in-game options.


FilmBoy84:
Thanks as always for this PR
Reviewed and LGTM
Merging
🍻

PR #1219, opened 2023-05-27
Change the cursor to normal when inventory or options menu are opened

Kurtsley:
Just a small UI tweak that changes the battlescape cursor to normal when the inventory screen is opened. This also does the same when the options menu is opened using the button instead of the escape key.


FilmBoy84:
Thanks for the fix
Merging

Issue #1218, opened 2023-05-24
Bombs and grenades that will cause OpenApoc to CRASH


XCOM-HUB:
- OS: Windows 11 Pro 22H2
- OpenApoc Version: 0.0.731
- List of Mod files in use: None

If you prime a bomb or a grenade, throw it close to you, quickly pick it up before it blows up, and also kill the last enemy before it blows up. End the mission, and then go to a new Battlescape event. When you start the next Battlescape event, the bomb/grenade will detonate and crash OpenApoc. This bug is preventable if you go to your base and unequip the primed and thrown bomb/grenade before starting a new Battlescape event.

**Expected behavior**
The bomb or grenade should not remember being primed, thrown, and picked up. It should forget being primed in the first place, or at least there should be an option to unprime a bomb/grenade as soon as you pick it up. If the enemy threw the bomb/grenade, then you shouldn't be able to unprime it. But it shouldn't cause OpenApoc to crash if you killed the last enemy before the bomb/grenade goes off, and then start a new Battlescape event without unequipping the bomb/grenade in your base.

**Savegame**
You will find the primed, thrown, and picked up bomb (Marsec High Explosive) on Mattias in this save file named "Maybe bug".
https://www.dropbox.com/s/ipv50pe5fzxaxpc/save_Maybe%20bug.save?dl=0

**Copy of your OpenApoc_settings.conf**
"Explosions damage instantly" is turned off.

FilmBoy84:
Fixed by #1220

Appveyor builds will be ready in a couple of hours for you

PR #1220, opened 2023-05-28
Fix grenade related crashes

Kurtsley:
Addresses #1218.

There are two problems happening here. If you prime a grenade for any amount of time, drop it, then pick it up the game will crash when the timer runs out. If you pick up the grenade before it explodes and then end the mission, it will crash the game at the beginning of the next mission unless unequipped. The main problem stems from a unit update function in battleunit.cpp and the explode function in aequipment.cpp. It appears the update function is called every tick and iterates through each units inventory as part of the update process. The explode function will remove the grenade once it has exploded. When the grenade is place back into the inventory and explodes, the explode function will remove the equipment from the inventory and cause the crash.

The update function now iterates over a copy of the inventory instead of the actual inventory. This is a pretty broad change and could possibly lead to a slight performance hit or unintended results. Feedback appreciated.

The second fix was to simply unprime all grenades when the battle ends.

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -2546,12 +2546,14 @@ void Battle::finishBattle(GameState &state)
 	state.current_battle->unloadResources(state);
 
 	// Remove active battle scanners and deactivate medikits and motion scanners
+	// and unprime any grenades or proximity mines
 	for (auto &u : state.current_battle->units)
 	{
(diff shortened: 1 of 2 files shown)


FilmBoy84:
Nice catch :beers:

XCOM-HUB:
> Nice catch 🍻

That was my catch. Your welcome!

PR #1222, opened 2023-05-30
Add visible debug menu to all relevant screens

Kurtsley:
For the sake of convenience, I decided to add the debug commands at the top of the screen when F1 is pressed. As of now, only the cityscape is done. 

1280x720
debug1

OG Res
debug2

Not all commands currently showing, will fix later.

Would like feedback on format of the text and any other thoughts.

Regards.


FilmBoy84:
Great idea

A few things to consider

1) Maybe put a line in `OpenApoc_settings.conf` for `DebugCommandsVisible=True/False` or similarly worded to allow switching of this functionality. For example, I quite often have to fix issues on the fly with debug in a stream, i don't necessarily want viewers to have to see that unless it's something like a walkthrough on how to fix an issue

2) Regarding your question about fonts, i'm inclined to say have it use default system font - this covers us across all platforms and means that when OpenApoc is finally at a point where we can drop the Bitmap fonts and move to TrueType we don't have to consider that some extra work may be needed for reworking debug feedback

3) Ensure that the boundaries for the Debug function text do not exceed a 640x480 pixel space but are scaled by the scaler if set to a value other than 100% - this ensures that text will never be off screen for users that are running at OG 480i resolution or scaling up on high DPI displays

Kurtsley:
Added the option. My question actually had more to do with the formatting of the block of text as far as the wording goes and how well it fits. I am just using the default smalfont that is used for most buttons and text.

debug3
Scaling seems to work, the max width for the menu is 640 pixels so it should always fit. This is at 640x480 200% scaling.

Kurtsley:
Decided to put the base and research screen options on the cityview debug menu. It is only one button, F10, so I think people can remember that. Was thinking that the menu might look better stacked vertically? There probably won't be enough room at the 480i resolution.

FilmBoy84:
How many lines are you short for 480i?

Worth having an autoscroll up/down on mouse hover or timer?

Kurtsley:
I will try it tonight. Probably not worth trying to implement auto scroll if it doesn't fit.

Kurtsley:
Changing to vertical was a good call, looks much better.

City High Res
debug1

Battle High Res
debug2

City 640x480 200 Scaling
debug3

Battle 640x480 200 scaling
debug4

I am happy with how this works, will mark complete.

RedRobin84:
This looks nice!

Issue #1210, opened 2023-04-26
Agent unable to Fire Main Hand Weapon in Battlescape


Nurulinde:
I have experienced this problem a number of times before, but never got a "save" of it. 
One of my Agents does not fire his Main Hand weapon (Off-hand weapon = Stun Grapple), at an NME well within range, His adjacent buddy Agents all are firing at this same target.
Tried to manually force him to fire via "Targeting" mode.
Targeting Mode for Main Hand does NOT function as intended (No white rectangle surrounds weapon when clicked on, No crosshairs for a cursor) (Acting like it is bugged)
Off-hand weapon for this Agent DOES work in Targeting Mode (White Rectangle Appears, Crosshairs) (Obviously this weapon isn't firing automatically because target is well out of range).

I attached a Save Game with this Agent in combat exhibiting the aforementioned problem.  It is Agent 1 of Group 3.  I play in Real Time Mode, and his intended target is the Spitter to his right, although that is immaterial since he simply just cannot go into Main Hand Target Mode nor Auto fire his weapon.

[save_Group 3 Agent 1 cannot fiire.zip](https://github.com/OpenApoc/OpenApoc/files/11330351/save_Group.3.Agent.1.cannot.fiire.zip)

FilmBoy84:
I cannot replicate this bug under 20230422

On loading the game, I can select the agent, run to within 5 tiles of the spitter, he fires his stun grapple (losing one charge as you can see) and the spitter falls stunned

![image](https://user-images.githubusercontent.com/33913088/234609463-9a39786c-e750-4cd1-8bfb-d761b06cd6d6.png)

FilmBoy84:
Tried loading again, and yep, i can get within range for the stun grapple and manually use it too
![image](https://user-images.githubusercontent.com/33913088/234610321-d8685687-9e8f-4338-80b4-4ea4d8306d99.png)
![image](https://user-images.githubusercontent.com/33913088/234610304-8ce31cc7-a352-4864-90e3-993ced0a7f1f.png)

Nurulinde:
[...]

Issue #607, opened 2019-06-15
Personal Disruptor Shields Do Not Recharge After Mission


Quickmind01:
After completing a mission, two agents had personal disruptor shields on, but they did not recharge to 100% once back to the Geoscape. They also did not recharge after hours had gone by.

![image](https://user-images.githubusercontent.com/51022165/59552996-666daf80-8f54-11e9-94c1-9fd3e5adc65a.png)

Quickmind01:
When back at base, if the PDS is dropped into the base inventory, it can immediately be placed back onto the agent fully charged.

Quickmind01:
Rechargeable weapons can be added to this list as well. In build OpenApoc-v0.1-606-gd53c36dc with the weapons mod, I saw laser pistols did not recharge after I finished a mission, but they would recharge as soon as I entered one. I have not tested other weapons but would assume they would follow this pattern.

FilmBoy84:
This should be resolved by Kurtsley's PR #1223 which also affects rechargeable equipment
As the lingering issue with clip handling does not apply to rechargeable items, i am closing this issue
Please report any further problems with another report, citing this one if required

PR #1223, opened 2023-06-03
Empty weapons reload at start of battle

Kurtsley:
Looking at the weapon and ammo related bugs.

- [x] Empty weapons at start of battle (if actually a bug)
- [x] Shields not recharging

I didn't see a final decision regarding if clips should be filled after a mission or left partially filled and dealt with accordingly. I see there is an advanced inventory controls option that mentions unloading clips, but I don't see this put to use.

This would reload empty weapons at the start of the battle, which would solve #1210. Is this OG behaviour? If not it could be an option.

This also addresses #607 by recharging all equipment that needs to be recharged after a mission. I tested this on the disruptor shields and the rechargeable sniper from the EWM.

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -2503,6 +2503,8 @@ void Battle::enterBattle(GameState &state)
 	for (auto &u : state.current_battle->units)
 	{
 		u.second->updateCheckBeginFalling(state);
+		// Reload empty weapons at start of battle if needed
+		u.second->reloadWeapons(state);
 	}
 
 	// Find first player unit
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -2556,6 +2556,16 @@ void Battle::finishBattle(GameState &state)
 			e->battleScanner.clear();
 			e->inUse = false;
 			e->primed = false;
+
+			// Recharge all equipment
+			auto payload = e->getPayloadType();
+			if (payload->recharge)
+			{
+				if (e->ammo < payload->max_ammo)
+				{
+					e->ammo = payload->max_ammo;
+				}
+			}
 		}
 	}
 
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -2504,7 +2504,10 @@ void Battle::enterBattle(GameState &state)
 	{
 		u.second->updateCheckBeginFalling(state);
 		// Reload empty weapons at start of battle if needed
-		u.second->reloadWeapons(state);
+		if (config().getBool("OpenApoc.NewFeature.AutoReload"))
+		{
+			u.second->reloadWeapons(state);
+		}
 	}
 
 	// Find first player unit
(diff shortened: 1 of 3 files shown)


FilmBoy84:
Sorry for the delayed reply to this
Had to go back to OG to check behaviour

Okay, from what I can tell, this is what happens in OG

If an agent returns to base from a battle with partial clips the following happens:
1) All partial and empty clips are taken from the agent inventory and refilled from available base inventory - this ensures that agents never leave the base without full ammo clips unless there is nothing to replenish them, in which case, they will retain partial clips firstly outside the weapon elsewhere in their inventory (priority is given to full weapon reload) and only leaving partial weapon load if there is nothing in the base and nothing elsewhere in their personal inventory to replenish it
2) Any partial clips left in the base are listed next to the full clips at base - inherently, there will only ever be 1 partial clip at most as all partials are consolidated
3) There is an Original Game bug, that is not always applicable, where sometimes "extra ammo" can appear out of nowhere to replenish partial clips, but this is rare, and of course not intentional

In the case of agents who leave a battle, then travel to another without returning to base, the following happens:
1) 1) All partial and empty clips are taken from the agent inventory and refilled from available CRAFT inventory pool - this ensures that agents never leave the enter the next mission without full ammo clips unless there is nothing to replenish them, in which case, they will retain partial clips firstly outside the weapon elsewhere in their inventory (priority is given to full weapon reload) and only leaving partial weapon load if there is nothing in the craft inventory and nothing elsewhere in their personal inventory to replenish it
2) Any partial clips left in the craft are listed next to the full clips in the craft's cargo module(s) when accessing agent inventory for a vehicle outside of the base - inherently, there will only ever be 1 partial clip at most as all partials are consolidated
3) The OG bug with mysteriously appearing ammunition does not seem to occur with craft and agents between missions that do not return to base.

FilmBoy84:
Ah, I haven't covered vehicle weapons... Most certainly something for another PR, but I'll post here and link later whilst it's still in my mind.

Essentially it's a simplified version of the above:

1) On return to base, ammunition is automatically replenished from the pool of ammunition at the base.
2) I am not aware of any weapons in the OG that use "clips" so partial ammunition is not something that inherently happens - there's just a straight deduction of the amount required from the base pool into the craft. If there is not enough to refill every weapon or vehicle, it prioritises the vehicles with the lowest vehicle ID and then each weapon attached to that in a similar manner to that of agents 1) there is room here for improving the system - should all available ammunition be split, as evenly as possible, between all remaining craft and their weapons? 2) Should this be by default? 3) Should it be an end-user toggle option? 4) Are there a variety of priority systems that should be implemented as options?
3) We probably need to consider that mods may add clip based vehicle weapons - particularly with Burst weapons already being a new addition to OpenApoc. 1) Should this be dealt with similarly to Agent clips? 2) How will we represent partial clips in the UI - both on the equipment screens and in the cityscape?

Kurtsley:
Seems like that could take a bit to figure out. Might just mark this as complete and open a separate PR.

PR #1226, opened 2023-06-12
Moreoptions menu

Kurtsley:
Would like some feedback on something I've been working on. I think it is time to ditch the more options list and move to a different options menu. When trying to add features such as a map scroll speed scrollbar, there is no room left on the original options menu. 

Here is the rough example I'm working with:
menu

So far the options list has moved to this new menu, but you can see there is now room to add different options besides a checkbox in a list.

If a consensus agrees this would be a good idea moving forward, the menu could be spruced up with a frame around the list. I'm also thinking of moving some of the more important options out of the list and adding a vanilla mode toggle that would toggle all options to simulate a vanilla experience.

Let me know what you think.

TODO:

- [x] Menu design done
- [x] Menu functionality done
- [x] Clean up old menu

### DIFF
--- a/data/forms/moreoptions.form
+++ b/data/forms/moreoptions.form
@@ -68,40 +68,6 @@
 					<image>PCK:xcom3/ufodata/icons.pck:xcom3/ufodata/icons.tab:15:ui/menuopt.pal</image>
 					<imagedepressed>PCK:xcom3/ufodata/icons.pck:xcom3/ufodata/icons.tab:16:ui/menuopt.pal</imagedepressed>
 				</graphicbutton>
-				<label text="Map Scroll Speed">
-					<position x="56" y="166"/>
-					<font>smalfont</font>
-					<size width="148" height="16"/>
-					<alignment horizontal="left" vertical="centre"/>
-				</label>
-				<graphic>
-					<position x="42" y="182"/>
-					<size width="224" height="18"/>
-					<image>PCK:xcom3/ufodata/newbut.pck:xcom3/ufodata/newbut.tab:61:xcom3/ufodata/research.pcx</image>
-					<scroll id="MAP_SCROLL_SPEED">
-						<position x="14" y="2"/>
-						<size width="148" height="15"/>
-						<range min="0" max="50"/>
-					</scroll>
-					<graphicbutton scrollprev="MAP_SCROLL_SPEED">
-						<position x="0" y="0"/>
-						<size width="16" height="20"/>
-						<image/>
-						<imagedepressed>BUTTON_SLIDER_LEFT_DEPRESSED</imagedepressed>
-					</graphicbutton>
-					<graphicbutton scrollnext="MAP_SCROLL_SPEED">
-						<position x="161" y="0"/>
-						<size width="17" height="20"/>
-						<image/>
-						<imagedepressed>BUTTON_SLIDER_RIGHT_DEPRESSED</imagedepressed>
-					</graphicbutton>
-					<label id="TEXT_MAP_SCROLL_SPEED" text="1">
-						<position x="184" y="2"/>
-						<size width="36" height="14"/>
-						<alignment horizontal="centre" vertical="centre"/>
-						<font>smalfont</font>
-					</label>
-				</graphic>
 			</graphic>
 		</style>
 	</form>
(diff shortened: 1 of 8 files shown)


FilmBoy84:
I really like this, and agreed, it resolves a number of issues with the current screen

It also brings consistency with the Cheats menu that uses the same background

With regards to the moving out of some options - I'm inclined to say don't do this, I often want to change things into a neither vanilla nor "all the QoL changes" state. Being able to customise to the end user is one of the appeals of OpenXCOM for example

However, I agree that a single toggle for switching whatever the user has set to a "vanilla" experience with a click is welcome

What were you thinking of removing?

Kurtsley:
I was just thinking of taking some options and moving them from the list to a dedicated check to the left. Maybe stuff that doesn't have anything to do with gameplay like the debug menu toggle. I am also going to try to organize the list to read a bit better. I would like some help with the vanilla option as I'm not sure exactly how the original played so I don't know which options would return the game to OG.

FilmBoy84:
I'd have to go through and check

I know there's some errors - for example the "X-COM 1 Damage Model" is actually EU/UK Damage Model the US Damage Model is 50-150% so it's regional not just game dependant

If I get time tomorrow I'll dive into it for you :)

Ah, sorry, when you said move some options out of the list, i misinterpreted that as remove entirely
I agree, actually organising things to be more intuitive in placement and grouping would be very welcome :)

Kurtsley:
First of all, thanks for the quick work on the vanilla list.

The design is coming together. I have split the options into three categories, cityscape, battlescape, and general. This is how it looks now:

menu2

I have left room on the right for any slider type options that need to be added. The general options may need to be put into a list at some point, but I like the look for now. Having them outside a list makes adding a general option a bit less user friendly.

The vanilla toggle is still a work in progress. Also, adding this to the main menu may be harder than I thought. Might need a gamestate? Let me know your thoughts on the design.

FilmBoy84:
Looking very smart

Given the amount of options and the 480p screen it's very efficient but readable 👍 

I like that you have space for sliders and list buttons for General

As to vanilla toggle and gamestate, that may be something for @JonnyH to comment on as I'm not sure if there are any caveats to gamestate additions we need to consider that I may not be able to think of

Otherwise, looking forward to seeing this merged :)

Kurtsley:
Removed the vanilla options toggle for now, will add later. I'm happy with the look of this for now.

makus82:
options

Kurtsley:
What am I looking at here?

makus82:
> What am I looking at here?

more options bg with logic blue color

makus82:
> First of all, thanks for the quick work on the vanilla list.
> 
> The design is coming together. I have split the options into three categories, cityscape, battlescape, and general. This is how it looks now:
> 
> menu2
> 
> I have left room on the right for any slider type options that need to be added. The general options may need to be put into a list at some point, but I like the look for now. Having them outside a list makes adding a general option a bit less user friendly.
> 
> The vanilla toggle is still a work in progress. Also, adding this to the main menu may be harder than I thought. Might need a gamestate? Let me know your thoughts on the design.

also lists can be in 2 columns not 3 
coz general opntions can be considered cityscape because the game starts from the cityscape
and in that column may be such an inscription : general and cityscape opntions

just a thought

Kurtsley:
I consider options that affect the cityscape and battlescape as general options. That being said, all options added that aren't a checkbox, such as a slider, will not be in either listbox. For example, I will be adding a map panning speed feature, but it will have to be in general options.

makus82:
then general column can be attached to vanila options at bottom of list...

FilmBoy84:
> 

Is the blue-screen from the game files Makus?

If so, what's the PCX called please?

If it's a custom one, it'd need to be implemented as a mod because OpenApoc cannot distribute any original game assets that end-users do not provide themselves. Remixed assets are still technically owned by Take2 so we enter a dubious area if we start putting them with the builds

I agree though, the blue is a very nice background and matches the main options menu so if it is an existing PCX from the CD-ROM/ISO/CUE/BIN we should call it instead of the brown for options menu colour scheme continuity

The brown, as I said in an earlier post, nicely matches the cheat menu though so I'm not too worried if the blue is not in the original game files

PR #1229, opened 2023-07-03
Fixed a formatter crash when pressing F3 in debug mode (show walkmode)

Andy51:
A random fix for a crash I stumbled upon when playing with debug hotkeys


FilmBoy84:
Nice catch, thanks for the PR
Will merge as soon as checks pass :)

Issue #1064, opened 2021-10-05
Alien Dimension Map 3 - Burning/Collapsing Tiles cause a nullptr crash


gyunaev:
**Describe the bug**
Game crashes in alien farm map (Alien Building 3) - when you drop a vortex mine to explode the alien portal, there is major explosive scenery above it, which starts collapsing. Once this happens, the following error pops up on the screen:

`E OpenApoc::Tile* OpenApoc::TileMap::getTile(int, int, int) Incorrect tile coordinates 8,38,-1`

and the game crashes in `OpenApoc/game/state/battle/battle.cpp:1299`:

        if (map)
        {
                auto tile = map->getTile(position);
                // Cannot add non-fire hazard if tile is blocked or fire hazard if nothing is there to burn
                // at
                // HERE WE CRASH BECAUSE tile is nullptr
                if ( (!fire && tile->height * 40.0f > 38.0f) || (fire && tile->height * 40.0f < 1.0f))
                {
                        return nullptr;
                }

It crashes because tile is null. Suppressing the check by returning nullptr in this case doesn't resolve the crash, it just happens later in some other place (I assume map is corrupt at this moment).

**To Reproduce**
1. Load the attached save.
2. Click the "Normal speed" and wait until the thrown Vortex mine explodes and the scenery starts falling.
3. Game crashes.

**Savegame**
[save_progress-2.zip](https://github.com/OpenApoc/OpenApoc/files/7283614/save_progress-2.zip)

gyunaev:
as a "workaround" when playing this map, enable debug more, press R to reveal all aliens, press and hold S and click on all of them to stun them. Once they're all stunned the mission ends (another bug #1065 , but works for us)

Andy51:
I have found out that this particular map section is buggy and does not seem to have ground tiles below level 2 as any normal section does, so when ground on level 2 is blown up the tiles start to fall below level 0 because there is nothing to stop them.
![Screenshot 2023-07-10 010736](https://github.com/OpenApoc/OpenApoc/assets/4835002/87227346-3ba3-4264-9e7a-802e2cca609d)
Could not confirm if this is really an issue with original data but suggesting a workaround in #1230

Artytoo:
Even a Sword-Bot with Marsec Armour had this problem in the Original Apoc.

PR #1230, opened 2023-07-09
Fixed collapsing tiles falling through the ground

Andy51:
and causing a crash with some buggy map sections that are missing ground 0 layer of tiles, like the one with a portal and phallic pillars around it in Maintenance Factory Alien building (#1064)


game/state/battle/battlemappart.cpp
```diff
@@ -1235,6 +1237,23 @@ void BattleMapPart::updateFalling(GameState &state, unsigned int ticks)
 					break;
 			}
 		}
+
+		if (newPosition.z < 0)
+		{
+			// This can happen on some bugged map sections that are missing ground tiles on level 0
+			// so that can let tiles fall through into the abyss. Stop the ground tiles from falling
+			// further and let die() convert them to destroyed level 0 ground to close the gap.
+			if (!toDestroy && type->type == BattleMapPartType::Type::Ground)
```
Andy51, line 1246: This one is a questionable workaround because it is an attempt to close the bottomless pit with falling debris, maybe a better solution would be to somehow patch the section during the loading instead to have ground tiles on level 0

FilmBoy84: This does present an interesting problem

IIRC, some maps intentionally use space with no ground tiles - for example, the Dimension Gate Generators on the last mission, the "black void" has no bottom in the OG and destroys anything that passes out of bounds (for example bodies)

I agree that this tile appears to be an inherited original game bug

A better solution may be modifying the offending tile's XML to add in ground tiles for that section that way we're not patching everything that may use/need voids but am open to suggestion

IIRC, the map XML is under data/maps/<**offending map pack**>

Andy51: I specifically checked Dimension Gate having the same idea about it, and it turns out the abyss has ground tiles too and they are even walkable. Not sure if it was the same in OG though.
Yes, looks like data/maps/47maint_02 is the bugged section, and that maybe the only one. I thought that including even derivative files from OG is a big no-no, so how bad would it be to hardcode a patch into map data converter specifivally for this section?

FilmBoy84: All the <mapsector.xml> does is assign OG assets provided by the end user to specific coordinates and with specific properties IIRC

As long as you don't include anything from the OG - just reference it - we should be okay

I hope it's just a case of adding enough ground tile entries at the appropriate coordinates in that XML

I wonder if @Istrebitel can advise?
I have done very little with the map array and maps
But I know his work was core to the current functionality and how things are handled in OpenApoc

Andy51: So I confirmed this is really an issue with original data.
Just including a fixed mapsector file into the repo does not make much sense to me because these files are generated as part of the build process from OG and would be rewritten. My next thought was that maybe we can include an XML data patch for the defective mapsector similar to how gamestate patches are applied from https://github.com/OpenApoc/OpenApoc/tree/master/data/common_patch . Or include an already patched mapsector as part of default mod, but I am not sure if mapsectors are picked up from mod folders.

FilmBoy84: I do not believe battlescape map sectors are picked up from mod folders currently - the city maps are however - you can find them in the base mod submods serialised in gamestate

Personally, I'm erring on XML data patch as we already use XML to handle the maps and it saves us distributing an entire fixed mapsector

It also would be easier to mend/alter if there are any further issues or expand to cover other mapsectors that may be problematic

Andy51: I updated my attempt at fixing this by adding resource_patch data folder with a thought that all potential resource patches that aren't part of the gamestate should go there. These patches are only picked up for mapsector extraction for now and does the trick for fixing 47maint_02, but the side effect of this is that extractor spams errors in the log like these:
```W bool __cdecl OpenApoc::ZipDataProvider::openArchive(const class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &,bool) Failed to init zip file "data/resource_patch/maps/01senate_01" for reading
W class std::unique_ptr<class OpenApoc::SerializationArchive,struct std::default_delete<class OpenApoc::SerializationArchive> > __cdecl OpenApoc::SerializationArchive::readArchive(const class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &) Failed to open archive at "data/resource_patch/maps/01senate_01"
E bool __cdecl OpenApoc::BattleMapSectorTiles::loadSector(class OpenApoc::GameState &,const class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &) Failed to read "data/resource_patch/maps/01senate_01"
W bool __cdecl OpenApoc::ZipDataProvider::openArchive(const class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &,bool) Failed to init zip file "data/resource_patch/maps/01senate_02" for reading
W class std::unique_ptr<class OpenApoc::SerializationArchive,struct std::default_delete<class OpenApoc::SerializationArchive> > __cdecl OpenApoc::SerializationArchive::readArchive(const class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &) Failed to open archive at "data/resource_patch/maps/01senate_02"
```
should I quiet down these somehow?

Reverted the workaround for ground tiles falling through and now just issuing an error to identify any other problematic sectors.

FilmBoy84: Is that serialisation error because unpatched map data is currently in `/data/maps/XXnnnn_XX` style directory structure where XX is a value and n is the map name?

Only patches are in `/data/resource_patch/maps/XXnnnn_XX`

It looks like it's assuming everything is there?

Andy51: Just before forming an initial set of XMLs in /data/maps/XXnnnn_XX from original game data it now looks up if there is a matching patch in /data/resource_patch/maps/XXnnnn_XX to merge, and when it is not found (which is the common case) it logs a number of error messages. I did not want to hardcode a check for a specific sector with path to a single patch file, but maybe that was the right thing to do instead?

FilmBoy84: It's very much your call

Instinctively I say if possible, go with the solution that silences the errors properly - but I appreciate the extra work could be put to other things

On the other hand, if we have a worked solution that results in no log errors it makes things cleaner for bug whacking and reduces risk of false positives

I believe that the modding structure already takes the more complex route - as comparisons have to be made in gamestate between "base mod" XML and any additional or replacement XML provided by mods to have some form of unity between the way mods in `data/mods/` are loaded and anything in `data/resource_patch/` is loaded would be potentially cleaner down the road - particularly if we get to the point of having map editors where people may be providing additional patched or altered mapsectors

FilmBoy84:
I finally got round to testing this the other day
Apologies for delay in merging
Thanks @Andy51

PR #1231, opened 2023-07-16
Auto detect number of displays in the launcher

Kurtsley:
This removes the MAX_DISPLAYS variable and instead checks the number of attached displays when the launcher starts.

Two questions:

- Was the MAX_DISPLAYS variable chosen for a reason?

- Should this be kept but replaced with the displays.count() variable?
https://github.com/OpenApoc/OpenApoc/blob/e3a41f96078237cf31aee3337afa6dfcd52674c8/tools/launcher/launcherwindow.cpp#L218-L221

I figured no for both so I removed them.

### DIFF
--- a/tools/launcher/launcherwindow.cpp
+++ b/tools/launcher/launcherwindow.cpp
@@ -205,17 +205,17 @@ void LauncherWindow::setupScreenModes()
 
 void LauncherWindow::setupDisplayNum()
 {
-	constexpr int MAX_DISPLAYS = 4;
+	const auto displays = QGuiApplication::screens();
 	auto &comboBox = *ui->displayNumBox;
 	comboBox.clear();
 
-	for (int i = 0; i < MAX_DISPLAYS; ++i)
+	for (int i = 0; i < displays.count(); ++i)
 	{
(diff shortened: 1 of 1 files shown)


Andy51:
curDisplayValue check is probably needed just in case an option had invalid number in it?

FilmBoy84:
I believe Andy51 is correct and curDisplayValue is used in cases of invalid or missing values

I agree that MAX_DISPLAYS is an odd choice - I also don't like that hardcoded `4`, especially with some users now exceeding that number of displays for gaming / streaming walls or when using daughter streaming devices with a screen, so feel removing it and replacing was a good move going forward

We also need to check behaviour on LINUX builds - anyone here running linux?

Kurtsley:
So leave the check in place but replace with the new displays variable? I can check it on Linux later today.

FilmBoy84:
> So leave the check in place but replace with the new displays variable? I can check it on Linux later today.

I think that's the logical way forward - we're covered for systems with more displays than the old MAX_DISPLAYS value and we have a way to handle null values or bad numbers

Kurtsley:
It seems to work the same on Linux. Keep in mind I only have two monitors.

Issue #696, opened 2019-08-16
[Cityscape] ATVs should be able to approach crashed UFOs


FilmBoy84:
Proselyte Ethan suggested on discord that ATVs should be able to approach crashed UFOs; after-all UFOs always crash to floor and it would give another use for the wolfhound APC

Checks would be needed to ensure that only ATV vehicles could approach UFOs unless you want to go super-complex and allow none-ATV vehicles to approach craft only if the UFO crashes on a road tile

In either case, adding this here as a feature request as it applies to functionality already partially added to OpenApoc

Quickmind01:
I don't know if this should be a separate issue, but allowing multiple craft to land at a UFO would be nice. The code seems to be there in vanilla already for Alarms. When responding to Alarms, if multiple craft are sent, the Begin Mission box (or whatever it says) doesn't pop up until all agents have arrived. But for some reason, when attacking a UFO, the first craft to make it is the only one allowed.

Paddywhacker:
Is this just for getting agents there?

Then any way of moving agents should be supported, with X-Com able to raid the UFO at any stage in case an agent or two loses their way.

FilmBoy84:
Updated to a low level bug as off-road vehicles are now implemented
So they should be able to attend UFO missions

PR #1232, opened 2023-07-20
Allow ATV type vehicles to initiate UFO missions.

Kurtsley:
Addresses #696. 

This allows ATV type vehicles to initiate UFO missions, but not recover UFOs. If an ATV type vehicle approaches an unmanned UFO, it just returns to base. This could be done better, maybe not even try in the first place. Should they also be able to recover the UFOs?

### DIFF
--- a/framework/options.cpp
+++ b/framework/options.cpp
@@ -135,6 +135,7 @@ void dumpOptionsToLog()
 	dumpOption(optionLeftClickIcon);
 	dumpOption(optionBattlescapeVertScroll);
 	dumpOption(optionSingleSquadSelect);
+	dumpOption(optionATVUFOMission);
 
 	dumpOption(optionStunHostileAction);
 	dumpOption(optionRaidHostileAction);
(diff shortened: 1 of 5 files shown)


FilmBoy84:
Oh, nice, thanks for picking this up <3

As to your question - given that this option allows for ATVs to enter missions (which would result in UFO recovery of much larger UFOs) I propose that unmanned UFOs are also recovered as normal

It's entirely possible for a Hovercar to recover a battleship in both OG and OA, so I don't think we should make a distinction in OA just because a vehicle is on the ground - if the user has gone to the trouble of driving something to a UFO, arguably, they should be able to recover it

I fear anything breaking consistency with vehicles that can recover UFOs will result in more bug reports for little immersion gain

Kurtsley:
Gotcha. I'll change it back so the vehicles act as normal then. I added this as an option and made it false because the pathing is still questionable for ground vehicles. I had issues where the vehicles would choose the wrong path and explode if they couldn't get to the UFO. Do you think this should be on by default?

FilmBoy84:
I'd say it's best off by default as you have it - we can always change it to default on if we get reports from users saying it doesnt feel right. And with the explosion risk - until pathfinder issues with RVs are resolved that's another reason to have off by default :)

Your new More Options screen clearly displays it

And any user going to look in there will easily find it and toggle as desired

Kurtsley:
And just to be clear, road vehicles still can't initiate a UFO mission even if the UFO somehow crashed on a road. This is just for ATVs.

FilmBoy84:
Sounds good

FilmBoy84:
Finally got to test this and LGTM
Merging
Thanks, as always <3

PR #1234, opened 2023-07-29
Fix selected agent dying in battle causes error

Kurtsley:
Fixes this annoying issue that crops up sometimes after an agent dies in battle:

noagent

This was happening if a selected agent in the cityscape died in battle. The gamestate was updating the agents and expecting this selected agent to still exist. This clears the selected agents before a battle to make sure this can't happen.

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -2535,6 +2535,9 @@ void Battle::enterBattle(GameState &state)
 
 	// Remember time
 	state.updateBeforeBattle();
+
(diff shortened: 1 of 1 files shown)


makus82:
> Fixes this annoying issue that crops up sometimes after an agent dies in battle:
> 
> noagent
> 
> This was happening if a selected agent in the cityscape died in battle. The gamestate was updating the agents and expecting this selected agent to still exist. This clears the selected agents before a battle to make sure this can't happen.

Thx for fix Kurtsley

Issue #991, opened 2021-02-09
No Vehicle Matching ID / when entering alien dimension


pyhton:
** Reproducible error like #940 , but different **

**Problem:**
An error occurs when entering the alien dimension. Game is blocked by error loop.

**Description:**
After a few days of trial and error and reloading various savegames, the error seems to appear when the alien ships are shot down 4.April in my game (save game attached)

- If I don't shoot any of them down and skip the day, I can enter the alien dimension as normal.
- If I shoot down one of them (tested with shooting down the transporters) I get the error at midnight when I try to enter the alien dimension.
- This is very annoying ofc. I was much further in the game when entering the a.d. and realized it very late.

So it seems it has something to do with shooting down the enemy ships but ofc. I could be wrong.

[Error+Save.zip](https://github.com/OpenApoc/OpenApoc/files/5953782/Error%2BSave.zip)

makus82:
Hi and welcome @pyhton and thx for your retailed report =]

Kurtsley:
This has been mentioned before, but this issue is caused by projectiles in the alien dimension. The game is checking these projectiles for collisions and not finding the craft they were fired from. This could be tackled a few ways:

- Remove the projectiles when the craft is destroyed
- Remove projectiles from the dimension when the craft leaves
- Try to remove this check and create a LogWarning stating the owner of the projectile no longer exists
- Probably a better way I'm not thinking of

I think the bottom line is the projectiles can't just disappear when the craft that fired them is destroyed.

FilmBoy84:
> This has been mentioned before, but this issue is caused by projectiles in the alien dimension. The game is checking these projectiles for collisions and not finding the craft they were fired from. This could be tackled a few ways:
> 
> * Remove the projectiles when the craft is destroyed
> * Remove projectiles from the dimension when the craft leaves
> * Try to remove this check and create a LogWarning stating the owner of the projectile no longer exists
> * Probably a better way I'm not thinking of

I've been doing some playing with this in the OG

One thing I have noticed is that when all X-COM craft leave the alien dimension, it doesn't always switch immediately if projectiles (or crashing UFOs and aircraft) are still present

Instead, it will run the simulation until everything has crashed, collided, or reached it's TTL (Time to Live) value

Only then will it switch

You can test this by entering the alien dimension with a dimension missile weapon, firing it, then leaving, as long as the missile is going to hit something, the simulation continues until it does then there is a switch of dimensions

[...]

PR #1235, opened 2023-08-03
Delay the switch from alien dimension to clear projectiles

Kurtsley:
Addresses #991 among others.

This delays the switch from the alien dimension until all projectiles have died and all UFOs have crashed. Has been lightly tested and seems to function.

Should this rule apply to anything else? XCOM craft too?

### DIFF
--- a/game/ui/tileview/cityview.cpp
+++ b/game/ui/tileview/cityview.cpp
@@ -1921,6 +1921,17 @@ void CityView::update()
 				switchDimension = false;
 				break;
 			}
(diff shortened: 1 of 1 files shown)


FilmBoy84:
LGTM and it seems to work nicely

I would say a follow projectile feature would probably be for another PR and made optional

The difficulty i foresee is how we prioritise what the tracking camera actually follows where there are, for example, two UFOs of different classes looking to crash, a couple of missiles and a handful of beam shots

Issue #1212, opened 2023-04-26
Agent Stat / Agent Equipment Screen - Agent Awards and Combat History not yet implemented / functional


FilmBoy84:
In the original game, the agent history could be accessed via their equipment / stat screen through a up/down arrow button at the top right of the portrait area

![image](https://user-images.githubusercontent.com/33913088/234592583-4bc50551-e51a-46b2-b3cf-56cb62aba9b6.png)

![image](https://user-images.githubusercontent.com/33913088/234592503-b8b4b079-f4aa-42c7-b21b-96097701f464.png)

The agent medals and awards structure also needs to be implemented and this is directly relevant to the combat history also

I am in two minds as to whether a separate issue is needed, but as they are co-dependant on one another in terms of record keeping in the gamestate, I shall leave it as one report for now - we can always separate later if needed

stewartmatheson:
Hi @FilmBoy84 

Are you happy to use https://www.ufopaedia.org/index.php/Agents_Medals_(Apocalypse) as a guide for the medals? Seems to have all the required details. There is also a pic of how they should be rendered here https://lparchive.org/X-COM-Apocalypse/Update%2057/ 

![xcoma002](https://github.com/OpenApoc/OpenApoc/assets/31781/5badaa80-6558-4663-bd8b-e512c086dfa4)

Seems like that's everything.

PR #1236, opened 2023-08-06
Issue 1212: Added agent history screen

stewartmatheson:
Hi OpenApoc Team,

First up  I apologise in advance for my dyslexia.

Thanks for the awesome work you guys are doing with this project, it's amazing. I started hacking a couple of weeks ago and I was able to get something working for #1212 the agent history screen.  I'm not a C++ dev so I'm sure there are many issues here but I figured I'd throw up a PR just for shits and gigs.  I have had a lot of fun working on this.  Throw the book or ignore it altogether. I won't be offended.

---

A breakdown of the changes I have made:

 * Fixed a id attribute in `aequipscreen` form
 * Created a new agent history form
 * Broke a profile form out from agent stats so it can be shared between the new history form and the existing stats form
 * Added an activatedInSquad member function to battle unit in order to count missions
 * Added a `recordKill` member function called each time a unit is attacked and dies.
 * The agent class now has a `hiredOn` memeber that should record the date they are hired.
 * Agent how has `getKills`, `getMissions` and getDaysInService in order to populate the values in the new history form
 * AgentSheet now has two different constructors depending on where it is called from and what data is required. Not all screens allow for displaying history.
 * A handler has been added to the toggle stats checkbox in order to toggle history hiding and showing.
 * Event handlers have been updated to now handle both forms.
 * AgentSheet display logic has been broken up into multiple methods to support the two different modes the screen can now display.

image

### DIFF
--- a/data/forms/aequipscreen.form
+++ b/data/forms/aequipscreen.form
@@ -104,7 +104,7 @@
         <image/>
         <imagedepressed>BUTTON_SCROLL_DOWN_DEPRESSED</imagedepressed>
       </graphicbutton>
-      <checkbox ID="BUTTON_TOGGLE_STATS">
+      <checkbox id="BUTTON_TOGGLE_STATS">
         <position x="167" y="56"/>
         <size width="22" height="21"/>
         <image/>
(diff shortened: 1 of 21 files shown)


makus82:
Heya Stewart! great work)

FilmBoy84:
Hi Stewart, thanks for the PR
LGTM and a much wanted feature on the roadmap

I'm happy to merge if you are able to fix the linting errors

We're currently using Clang Format 14 / LLVM 14

IIRC Ctrl-K then Ctrl-D will run these checks in visual studio once Clang is installed

FilmBoy84:
Hiya, saw your comment about how medals are awarded

Yes, as far as I know this is the correct system - however, i would add that any OpenApoc implementation may want to have the values for each award put out to an XML form for modding. However, that could always be a future PR - getting a system implemented to match OG is priority

PR #1241, opened 2023-08-16
Remove boost::filesytem support

JonnyH:
We don't support pre-c++17 compilers anyway, so we only need std::filesystem.


FilmBoy84:
Apologies in taking a while to review this
Merging

Issue #1244, opened 2023-08-27
Versions since 20230731 crash to desktop on recruiting new agents


FilmBoy84:
- OS: Windows 10 x64
 - OpenApoc Version [20230812 From GitHub Releases]
 - No Mods in use
 
**Describe the bug**
When entering the recruit agent screen, purchasing equipment screen or moving an agent into/out of a vehicle the game will crash with the following warning

`E class std::shared_ptr<class OpenApoc::Form> __cdecl OpenApoc::Control::findControlTyped<class OpenApoc::Form>(const class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &) const Failed to find control "AGENT_PROFILE_VIEW" within form "FORM_TRANSACTIONSCREEN"`

**To Reproduce**
1. Start a new game in 20230812
2. Head to the base screen
3. Click on the button to recruit agents

**Expected behavior**
Previous version 20230730 does not have this issue, clicking on any of the functions mentioned above does not crash the game

**Additional context**

This may be related to either the new agent history screen #1236 @stewartmatheson or changes made to LUA script loading (which is linked with the economy) #1238

FilmBoy84:
Fixed in #1247

PR #1245, opened 2023-08-28
Added medals to agent sheet.

stewartmatheson:
* Broke out the agent history sheet into its own class.
 * Created a new agent history form that displays the medals.
 * Added a medal tier to agent which allows for the agent to report its current medal tier. Note this is currently hard coded to zero as we don't have a system to track XP as yet.
 * Added display logic to render the medals based on the agents current tier.

image


stewartmatheson:
lets hold off on this till #1244 has been fixed.

PR #1246, opened 2023-08-30
Added fixes for issue 1244

stewartmatheson:
* recuritscreen now has correct subform
 * recruitscreen and transactionscreen will now reference their forms correctly
 * recruitscreen and transactionscreen will now hide and show forms correctly


FilmBoy84:
Many thanks for this

I can now move between screens without issue again
image

PR #1247, opened 2023-09-01
Multifix

stewartmatheson:
Fixed a few issues in the transaction screen forms where handles to form elements and subfroms had not been initialised correctly.
Fixed an issue in event processing where iterators were becoming invalid. 
Added guard conditions as not all battleunits will have an agent.


FilmBoy84:
Tested the following under build `20230830-2-g5df81e84`

Hiring agents
Making a market purchase
Transferring Agents
Making a market sale
Research and Engineering Tasks

All seems good, thanks for the fixes

PR #1248, opened 2023-09-11
Added medal rendering

stewartmatheson:
* Added medals to the agent history form
 * Added code that will hide the medals according to agents current medal tier
 * Added stub method to agent that will always return a tier of zero.

A screen shot with medal tier hard coded to 5

Note: This PR should not change anything that the player can see as the medal tier will always be set to zero. In the above screenshot, the medal tier has been hardcoded to 5 for illustrative reasons. This is intended as a precursor for a bigger set of changes around agent experience.

### DIFF
--- a/data/forms/sheet_agent_history.form
+++ b/data/forms/sheet_agent_history.form
@@ -47,6 +47,45 @@
         <font>smalfont</font>
       </label>
 
+      <graphic id="MEDAL_1">
+        <position x="11" y="155"/>
+        <size width="33" height="47"/>
+        <imageposition>fit</imageposition>
+        <alignment horizontal="center" vertical="centre"/>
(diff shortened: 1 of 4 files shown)


FilmBoy84:
Many thanks for this PR

I'll aim to review this weekend if someone doesn't beat me to it 🍻

makus82:
image
just for info there was also impruvement  line

FilmBoy84:
Finally got round to testing this

Started new game
Played two or three naturally occuring missions in the cityscape observing agent promotion
Medals appeared in their history
No crashes in hire/fire or the marketplace were observed
No form errors observed

Great work!
Many thanks for this PR and another merge to master 🍻 

@makus82 thankyou for the heads up on the missing string
I will create another issue on the tracker for this as agent training logic relates to another issue and we will likely need to fix agent training before another PR adds this missing data to the screen

PR #1250, opened 2023-09-27
Bug Fixes

stewartmatheson:
* Kills and missions where not getting saved correctly. They should be now.
 * Fixed a bug that was causing crashing when a unit is killed however there is no attacker set.


FilmBoy84:
Many thanks for these @stewartmatheson 

Thanks to a fix from JonnyH for the appveyor build I have now been able to test and merge

PR #1253, opened 2023-11-24
Add an option to unfreeze vehicles in game.

Kurtsley:
I have finally started to narrow down the stuck vehicle bug. While I continue to sort it out, I am going to add this debug option to unfreeze vehicles that become stuck during combat in the cityscape. Almost every large battle that breaks out has at least one stuck vehicle that prevents the player from using ultra-fast speed. This makes sure the the vehicles tickstoturn value is not 0, which fixes the issue. Anyway, shift + left click on a vehicle in debug mode will accomplish this.

### DIFF
--- a/data/forms/city/debugoverlay_city.form
+++ b/data/forms/city/debugoverlay_city.form
@@ -1,154 +1,160 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <openapoc>
-  <!-- Debug Overlay -->
-  <form id="DEBUGOVERLAY_CITY">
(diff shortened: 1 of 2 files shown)


Kurtsley:
Save I used to test.

[save_vehiclestuck.zip](https://github.com/OpenApoc/OpenApoc/files/13454969/save_vehiclestuck.zip)

FilmBoy84:
Brilliant, much appreciated for your efforts in this
It's a much needed workaround
Tested and merged

PR #1254, opened 2023-11-25
Add vehicle target debug view plus fix debug overlay cutoff.

Kurtsley:
This adds a debug option to view what each vehicle is currently targeting. I used this while trying to fix the stuck vehicle problem. It turned out that targeting wasn't the problem, but this is pretty useful. The last addition to the debug overlay was being cut off so that was fixed as well.

Green line = XCOM vehicle is targeting an enemy.
Red line = Enemy is targeting an XCOM craft.
Yellow line = The two are targeting each other!
2023-11-25 14_59_52-Greenshot


FilmBoy84:
Oh this is very useful
Tested and Merging 🍻

PR #1255, opened 2023-11-25
Fix vehicles getting stuck in cityscape

Kurtsley:
I think I finally narrowed this problem down. The vehicle was attempting to sidestep to another location when it had 0 tickstoturn. Since it was trying to turn but couldn't, this froze the vehicle in place. This checks if tickstoturn is over 0 before sidestepping. I will admit that I can't test every situation that could happen so this could have unforeseen consequences. I have included the save I was using to test, it is a gigantic firefight. One or more vehicles would get stuck every other time I played on average before the fix. So far the fix has stopped any vehicles from getting stuck for 30 plus playthroughs of the fight.

[save_stucktest.zip](https://github.com/OpenApoc/OpenApoc/files/13465294/save_stucktest.zip)


FilmBoy84:
Generated a significantly large UFO incursion with debug and no stuck vehicles there either
This seems to work
Many thanks for your efforts in resolving this

PR #1257, opened 2023-11-27
Add simple credits menu

Kurtsley:
Here is the first draft. Some thoughts:
I have not figured out how to do two columns and still have text alignment on both. I can do a split list, but the names are all over the place. That is why it is a single column down the middle. The github contributors will probably add some length to the menu. I noticed that a couple people on the discord list seem to be using their real names, I left them in for now but I don't know the rules. The separators between sections are crude underlines because I can't figure out how to get an image to show in the listbox yet. If someone appears on the developer and programmer list, I left them off the programmer list. I kept duplicates on the testers and translators list because that seems to be a seperate job.

### DIFF
--- a/data/forms/creditsmenu.form
+++ b/data/forms/creditsmenu.form
@@ -17,31 +17,31 @@
         <font>bigfont</font>
       </label>
       <!-- List of contributors -->
-      <scroll id="LISTBOX_SCROLL">
-        <position x="553" y="82"/>
+      <scroll id="LISTBOX_SCROLL" scrollpercent = "5">
+        <position x="555" y="84"/>
         <size width="26" height="356"/>
       </scroll>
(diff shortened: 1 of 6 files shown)


Kurtsley:
Fixed the centering and alignment issues. I have commented out the full names for now, will send the list.

FilmBoy84:
As you will have seen on Discord, I'm currently collating a list of names via posts/DM for what everyone wants to be accredited as

I'll then update the wiki list and DM you what we have so this can be merged

I'll aim to close the period for informing me on 9th/12/23 - we can always update later :)

Kurtsley:
Here it is with the completed list. I tried to set it up aligned left like in discord, but some names were too long and it didn't look right. I stayed with the centered layout and I think it's pretty good. Let me know if anything needs changing.

FilmBoy84:
Looks great as is, thanks for doing this 🍻

PR #1261, opened 2023-12-03
Fix list not appearing in ufoaedia.

Kurtsley:
Clicking the info button in the ufopaedia will now actually show the list of items.


FilmBoy84:
All working and correct, thanks for the fix 🍻 

image

PR #1265, opened 2023-12-05
Fix agent shields now show current health

Kurtsley:
Per the stream, agent shields now show current health until the shield is gone. I messed around with vehicles but they work differently because of the recharge. Should they be included?

### DIFF
--- a/game/ui/components/controlgenerator.cpp
+++ b/game/ui/components/controlgenerator.cpp
@@ -459,15 +459,16 @@ AgentInfo ControlGenerator::createAgentInfo(GameState &state, sp<Agent> a,
 	i.shield = a->getMaxShield(state) > 0;
 	float maxHealth;
 	float currentHealth;
+
+	currentHealth = a->getHealth();
+	maxHealth = a->getMaxHealth();
 	if (i.shield)
 	{
-		currentHealth = a->getShield(state);
-		maxHealth = a->getMaxShield(state);
-	}
-	else
-	{
-		currentHealth = a->getHealth();
-		maxHealth = a->getMaxHealth();
+		if (currentHealth == maxHealth || state.current_battle)
+		{
+			currentHealth = a->getShield(state);
+			maxHealth = a->getMaxShield(state);
+		}
 	}
 	i.healthProportion = maxHealth == 0.0f ? 0.0f : currentHealth / maxHealth;
 	// Set state that is used in battle


FilmBoy84:
For agents in battlescape, shield depletes from full down the whole bar, then HP depletes from the top down too (effectively a two layer approach of depletion and gain)

For agents in the city, if the HP is lower than maximum, but they have a shield, the bar is blue (as per shield level) but only as high as the hp (although the shield still functions at it's actual level before HP is eroded)

This allows you to see agents with shields, but lower than maximum HP in the vehicle and building screens... which is where I was struggling in the stream - it was showing a full blue bar because the shield is full, but I couldn't tell if the agent was injured because of that covering

So the OG workaround for this UI feedback issue was to show the blue bar only to the level of the agent HP in cityscape.

Vehicles i need to double check feedback for. It's difficult as with no OG debug controls I need to play the OG to the point I have craft with shields to replicate 😅 

Thanks for looking at this, I shall try to test the new implementation this evening 🍻

Kurtsley:
I see, I can already tell you this is incorrect then. I will need to seperate the city from battle before this is correct.

Kurtsley:
This seems to work as intended. As for vehicles, the cityscape is like their battlescape, so i'm guessing it is fine as is.

FilmBoy84:
Looking great, seems to work

Only issue I have found is the time to update the agent assignment screen

1) I select the agent to add the shield
image

2) I add the shield - here it is applied instantly as it should be :)
image

3) I go back to the "Vehicle Location" screen and he has not been updated - still showing no shield
image


What I have tested - that seems to work as intended - thanks :)
1) Injured Agent equips a shield - shield covers HP but HP shown above agent in battlescape as full bar
image
image

2) When I remove, it shows HP at correct level instantly on portrait
image
image

In battlescape, the injured agent shows himself as shielded, but at the HP level because he will always be fully charged at next mission - this is perfect

image

FilmBoy84:
Save for testing

(Zipped for Github)
[save_Shield Test 1.zip](https://github.com/OpenApoc/OpenApoc/files/13592371/save_Shield.Test.1.zip)

Kurtsley:
I saw that, I've been working on that problem for a bit. This game has an issue with forms not updating and it is never the same problem twice.

FilmBoy84:
What I'll do then, is merge this PR so the UI feedback issue is resolved as form updates are really another issue

Issue #1264, opened 2023-12-05
Message screen showing battle messages in city, clicking brings to random location on map


Kurtsley:
I noticed that in the OG game, the message screen seems to be split between city and battle. The city messages don't show up in the battle mode and vice versa. In OpenApoc, the message screen contains all messages regardless of where they originate. It also looks like the battle messages start empty every time the battle mode starts. I think this would be a cleaner look and would bring the game closer to the OG. If you click on the location of a battle message in the cityscape, it takes you to a random place on the map, seems useless.

I propose splitting the message screen, battle and city. The city screen would keep all messages while the battle screen would reset every battle.

PR #1266, opened 2023-12-07
Split messages between city and battle

Kurtsley:
#1264.

I just went ahead with this. The messages from the city get saved to a separate list when a battle begins. The battle starts with a blank message log. The city messages get loaded back to the messages list after the battle. It works in my limited testing, but I could see this causing an unforeseen issue. I would run this through a few different saves.


FilmBoy84:
Looks great

Im just diving into the original game to check if the battlescape doesn't append the cityscape on mission end though

I have a nagging feeling that whilst the battlescape list is reset at the start of each mission, the end mission list is added to the city list and, if you click on say where an agent died, it just moves the viewport to the building or dimension gate

FilmBoy84:
Will try and break it with the current implementation, and if all is well, i'll merge in a bit :)

PR #1268, opened 2023-12-08
Fix vehicle now travels smoothly in z axis

Kurtsley:
I noticed the vehicles traveled in a jerky motion when changing altitude without a mission. This seems to correct that although I'm not sure how often this even happens.

See video - https://youtu.be/gsdKiVrhCrM


FilmBoy84:
This works brilliantly, have tried with a few vehicles and no problems. Merging

Issue #300, opened 2017-12-04
Flying Vehicles do not immediately raise/lower their altitude when specifying the height.


FilmBoy84:
Currently, vehicles do not raise/lower altitude until you have them enter and exit a building (give them their next command)

Instead they stay at the same height across the city regardless of altitude setting until they enter a building.

They do raise themselves to go over obstacles, but return to their last altitude afterwards.

This should be changed to make the vehicle attempt to move to the specified altitude on the next movement.

FilmBoy84:
As part of #1269 testing, i ran some tests with altitude on Original Game Apoc this evening

Craft should start to alter their altitude incrementally according to engine spec and the next obstacle they face - maintaining or gaining/losing altitude to pass it then continuing to strive for the user specified altitude

https://youtu.be/qbTl1LbLTHQ

PR #1269, opened 2023-12-08
Add vehicles don't dodge when returning home

Kurtsley:
This has bothered me for a while and I'm not sure it's an OG function. When I'm trying to return a vehicle back to the base and it's in a firefight, it keeps dodging which prevents it from getting home as fast as possible. I think there should be some option that a vehicle can simply run away from a fight without dodging. 
In this PR, vehicles will run straight home if:
- The vehicle is returning to base
- The attack mode is set to either defensive or evasive

I would like some feedback on this, maybe they should run home if aggressive? Maybe this should just stay the way it is?

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -179,6 +179,13 @@ class FlyingVehicleMover : public VehicleMover
 				dodge = 100;
 				break;
 		}
+		// Don't dodge if running away
+		if (vehicle.attackMode == Vehicle::AttackMode::Defensive ||
+		    vehicle.attackMode == Vehicle::AttackMode::Evasive &&
+		        vehicle.missions.front().targetBuilding == vehicle.homeBuilding)
+		{
+			return;
+		}
 		if (vehicle.type->aggressiveness > 0 && randBoundsExclusive(state.rng, 0, 100) < dodge)
 		{
 			for (auto &p : state.current_city->projectiles)
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -180,9 +180,7 @@ class FlyingVehicleMover : public VehicleMover
 				dodge = 100;
 				break;
 		}
-		// Don't dodge if running away
-		if (vehicle.attackMode == Vehicle::AttackMode::Aggressive &&
-		    vehicle.missions.front().targetBuilding == vehicle.homeBuilding)
+		if (vehicle.missions.front().targetBuilding == vehicle.homeBuilding)
 		{
 			return;
 		}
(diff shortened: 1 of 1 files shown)


Kurtsley:
I changed it to aggressive stance only, I was wondering if they should even dodge at all if aggressive mode is on. Let me know what you find when you check OG.

FilmBoy84:
I tested this tonight for you (just an in-game test)

It appears the following is the case using the Valkyrie as a subject and attacking organisation buildings for them to attack me

1) Regardless of stance, the vehicle immediately tries to return to base
2) Depending on stance, the vehicle may make small turns to fire back at any vehicles pursuing it (evasive, it just goes home, ignoring all, aggressive, it will turn to launch missiles or fire back as soon as it can, but otherwise continue heading home immediately after firing. Other stances alter the frequency this is done)
3) The vehicle makes no attempt to dodge enemy attacks, regardless of it's stance. I must have remembered this from BETA, as it appears in Final Release, they will happily fly away to their deaths, only firing back if their stance allows it.

So, appears I was wrong, I shall try and do this in BETA at another point to see where the memory of craft avoiding fire on the way home comes from. But for Final Release, certainly, we need to just make it go home and only return fire if on one of the stances that permit it (some reversing may be required for how the intermediate stances work but full aggressive appears to allow turning to fire as soon as a weapon is (re)loaded - presumably giving the missile and air defence modules a purpose if equipped)

https://youtu.be/qbTl1LbLTHQ

Kurtsley:
See the comment I left on the video regarding firing back when going home. I see it this way so far:

- Always go straight home without dodging regardless of stance.
- If stance is not evasive, have the ability to shoot back if the enemy craft gets within range but never stop retreating.
- Should change height while moving, taking obstacles into account. The height should be reached before attacking anything.

Right now, the stance setting looks like it just affects how close the craft gets in combat and the dodge chance. The aggressive retreat mechanic will have to be tested a bit more. I may just use this PR to deal with the first two issues.

Kurtsley:
This should satisfy the first two points. Vehicles should now never dodge when going home and the evasive setting already disables firing. #300 should probably be done separately.

FilmBoy84:
Agreed, the third point is most definitely something for a future decision and PR as dodging does not appear to be final release OG behaviour, i need to check BETA OG and it is most likely going to have to be an "Optional" feature anyway. Moreover, it is something that should be controlled by the as-yet unimplemented "UnitAI / Decision Matrix / Learning AI" functionality linked to stance behaviours - so it could easily scale up in complexity massively and go many places that there are currently no solid design plans or extensive reversing

I have in playtesting identified another issue - when attacking buildings my Valkyrie only used it's secondary missile weapon, it never used it's primary laser weapon and indeed, when out of missiles, just sat above the targeted building doing nothing despite still having a full charge for it's laser. Before i make a tracker issue for this, I will aim to do some testing of craft with multiple weapons systems (for example the Hawk has three) and see if there is indeed a problem with how weapons are prioritised and used

Issue #1262, opened 2023-12-03
Cityscape Tabs Need Unique Selections


Quickmind01:
In OG Apoc, each cityscape tab had its own set of selected items. For instance, in your Craft tab, if you selected one of your craft, it would have a yellow border around it, indicating it was the main thing selected. If you went to the Hostile Vehicles tab, a vehicle selected there would have a yellow border as well, and your craft wouldn't lose its yellow border. This is useful for when you have the camera follow option enabled, and you want to follow your craft and an enemy craft. Also, this is very useful for tracking progress in your two research tabs and the workshop tab. If you select one unit from each tab, you can quickly monitor the progress in each tab. Currently in OA, selecting one researcher/worker deselects the others in the other tabs. OG saves the person selected in each tab separately. Either that, or switching tabs deselects them.

FilmBoy84:
Agreed, this is a useful OG feature

In effect, we need the game to remember previous selection for each tab and, in instances where that is not available select no item for each tab (as opposed to the first, which can be confusing or problematic)

It's worth noting this issue seems to apply across the board, from bases, through vehicles, agents, scientists and organisations

PR #1271, opened 2023-12-11
Add multi select for agents and vehicles

Kurtsley:
Addresses #1262 

This separates the tabs in the cityview into separate lists, allowing them to function individually. The follow vehicle is based on which tab is open, so it only works when the vehicle or hostile vehicle tabs are open. The Scientists can now be selected and monitored as in OG.

I may leave this as a draft for a bit while I continue to think about how to optimize this. Everything is basically being done multiple times now where it was just once before. I'm sure some of the vehicle logic could be condensed, we'll see. The most important thing is if this works just like the OG. On the surface, it seems to work, but there is room for things to break as usual. Please throw everything you have at this PR to try to break something.

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -2539,7 +2539,7 @@ void Battle::enterBattle(GameState &state)
 	state.updateBeforeBattle();
 
 	// Clear selected units in case they die
-	state.current_city->cityViewSelectedAgents.clear();
+	state.current_city->cityViewSelectedSoldiers.clear();
 }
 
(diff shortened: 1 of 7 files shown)


Kurtsley:
This appears to be working, I had many problems with list crap. The usual disclaimer applies, this hasn't been tested to the extreme. Perhaps QM could test this on the next stream and try to break something. I decided to go with functionality over optimization for now. Even if I was capable of optimization, this probably isn't the time.

FilmBoy84:
All appeared well on QMs stream, thanks for this 🍻

PR #1272, opened 2023-12-15
Fix scientists able to be assigned when not at base

Kurtsley:
If you recruit a scientist, they immediately  show up in the unassigned agents list and can be assigned to a project even though they aren't physically at the base.

As you can see, I have assigned Steve Green but he is in another building.
bio

I checked OG, it seems they don't show up in the list until they arrive at the base. Let me know if there is some reason for this that I can't think of.

### DIFF
--- a/game/ui/base/researchscreen.cpp
+++ b/game/ui/base/researchscreen.cpp
@@ -377,6 +377,9 @@ void ResearchScreen::setCurrentLabInfo()
 		if (agent.second->homeBuilding->base != this->state->current_base)
 			continue;
 
+		if (agent.second->currentBuilding != agent.second->homeBuilding)
+			continue;
+
 		if (agent.second->type->role != listedAgentType)
 			continue;
 


FilmBoy84:
Agreed, matching the OG is the best thing to do at present

Whilst it could be useful to "pre-assign" scientists/engineers as seen on QMs stream, we know that this has, at times in development also presented issues (for example it was interlinked with destructors for the long-running AgentID bugs when agents were assigned and then left the base for new duties or were fired)

Tested in game and works nicely

The only issue i noticed was the following - which may need a separate PR

1) Hire a new agent
image
2) The agent does not appear "out of base" (grey grid icon next to portrait) instead it appears with a "at base" icon (green X)
3) Hire the agent anyway
image
4) On returning to the cityscape, find the agent in the tabs - he will correctly appear with the grey-grid icon as he is away from base in a building. This then correctly changes to an Orange pair of wings when in transit, and finally a green X when at their new, assigned, base
image

BeornTB:
Issues notwithstanding, I **loved** the QoL improvement of pre-assigning scientists.

PR #1277, opened 2023-12-18
Attempt to fix missing agent message

Kurtsley:
Attempting to fix the recent missing agent message, just a guess on my part. The civilian list was not being updated when agents were fired, this could cause issues. The soldier list doesn't need this for some reason, they will drop off when killed or fired. I left a logwarning message I used in debugging in the code, this removes it as well.


Kurtsley:
I'm starting to think that this change might be breaking save games, not breaking the game when played from the beginning. QM started an old save when testing this.

Kurtsley:
I'll just set this as ready since I know there is at least one problem this will solve. If this is breaking saves, this will not fix that problem but I don't know what happened with QMs game. That game was messed up in many ways.

PR #1280, opened 2023-12-20
Fix unarmed vehicles patrolling when attacked

Kurtsley:
I noticed in the QM stream that when transtellar was attacked, every single space liner and taxi came out to patrol. I checked OG and found that those vehicles don't come out and try to attack as they have no weapons. This PR checks each vehicle in the building when attacked and only sends out vehicles with weapon slots.

### DIFF
--- a/game/state/city/building.cpp
+++ b/game/state/city/building.cpp
@@ -985,11 +985,14 @@ void Building::underAttack(GameState &state, StateRef<Organisation> attacker)
 	if (owner->isRelatedTo(attacker) == Organisation::Relation::Hostile)
 	{
 		std::list<StateRef<Vehicle>> toLaunch;
-		for (auto v : currentVehicles)
+		for (auto &v : currentVehicles)
 		{
-			toLaunch.push_back(v);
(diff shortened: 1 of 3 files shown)


FilmBoy84:
Im not sure if this quite replicates the OG

In OG, if there are org agents in buildings, they tend to be dispatched as soon as a building comes under fire in vehicles with no weapons - for example, as soon as the attack starts at 9:35 in this video, a bike is deployed to move an agent away

It's only when the org relations are put to negative do they respond with weapon carrying vehicles to the attack

I will try to capture some more accurate video - but i do recall attacking emergency, construction and transtellar buildings could result in a flurry of unarmed vehicles in the OG often before any armed response

https://youtu.be/qbTl1LbLTHQ?t=575

Kurtsley:
This just deals with what happens if the org is hostile and attacked, not sending unarmed vehicles to patrol around the building. I'm not sure if the agents fleeing in unarmed vehicles is implemented yet, but this shouldn't affect that.

PR #1304, opened 2023-12-28
Limit space liners to prevent late game slowdown

Kurtsley:
I figured this should probably be added because we are seeing massive numbers of space liners in late game saves. I picked the number 16 randomly, obviously it could be changed.


FilmBoy84:
Thanks for doing this, seems to work nicely with a quick test 🍻

PR #1305, opened 2023-12-30
Fix lab choice when no scientist selected

Kurtsley:
The current system doesn't seem to account for multiple bases and multiple scientist selections. This is how it works in this PR:

This is dependent on the tab selected eg. BioChem, Physics, Engineering.
- If a scientist is selected, it works the same. The lab that the selected scientist is working in will be shown.
- If no scientist is selected, the matching lab at the current base will be shown.

If a scientist is selected in another tab, they are ignored. Not sure this is OG.

### DIFF
--- a/game/ui/tileview/cityview.cpp
+++ b/game/ui/tileview/cityview.cpp
@@ -135,54 +135,70 @@ sp<Facility> findCurrentResearchFacility(sp<GameState> state, AgentType::Role ro
                                          FacilityType::Capacity capacity)
 {
 	sp<Facility> lab;
-	for (auto &a : state->current_city->cityViewSelectedCivilians)
+	if (!state->current_city->cityViewSelectedCivilians.empty())
 	{
-		if (a && a->type->role == role)
+		for (auto &a : state->current_city->cityViewSelectedCivilians)
 		{
-			state->current_base = a->homeBuilding->base;
-			if (a->assigned_to_lab)
+			if (a && a->type->role == role)
 			{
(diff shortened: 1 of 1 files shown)


FilmBoy84:
I'll do a video this afternoon of OG behaviour and check that this matches or improves on it for you

Seems to work well as is and a fix that is definitely needed <3

FilmBoy84:
Finally got round to testing in the original game - due to lack of debug controls i had to do some save editing but have ascertained that this PR largely follows the original game as seen here

https://youtu.be/Z3OOCXugqFg

A few observations that we need to double check all appear in OpenApoc

* All labs and workshops appear in Orange-Red tiles on the relevant base map at all times when in the lab/workshop screen
* The currently selected lab or workshop appears in blue on the relevant base map at all times when in the lab/workshop screen
* Selecting a different lab or workshop immediately moves the blue tile to cover that facility on the base map, making the prior lab/workshop Orange-Red
* When agents are assigned to a task and are selected from the cityscape lists, selecting the research and manufacture button with an agent selected will take you directly to the lab/workshop in which they are located at the base it is located
image
* If an agent is NOT assigned to a task, but the user selects them and then clicks on the research and manufacture button, the last lab/workshop the user selected at the base they selected it is loaded and shown - this is a very strange behaviour i'm not entirely sure if it is an OG bug or oversight - it seems strange that the game remembers the last selected lab and the base it was at but fails to show you the base the agent you selected is currently at when not assigned to a task. For this reason, the current PR actually fixes this - an improvement on the original game :+1:

PR #1306, opened 2023-12-30
Add option to clear space liners

Kurtsley:
In old saves, there seems to be a build up of space liners that causes the game to run slowly on the cityscape. This option clears all space liners from the city so more can be sent. Since the max is 16 liners per spaceport, none would be sent unless the existing ones were cleared. Maybe this should be automatically check when loading a save?

### DIFF
--- a/data/forms/city/debugoverlay_city.form
+++ b/data/forms/city/debugoverlay_city.form
@@ -149,14 +149,20 @@
 				<alignment horizontal="left" vertical="top"/>
 				<font>smalfont</font>
 			</label>
-			<label id="Base" text="Base Screen - F10 = Finish All Facilities">
+			<label id="MINUS" text="MINUS = Clear Space Liners (If game is running slowly)">
 				<position x="left" y="336"/>
+				<size width="350" height="14"/>
+				<alignment horizontal="left" vertical="top"/>
+				<font>smalfont</font>
+			</label>
+			<label id="Base" text="Base Screen - F10 = Finish All Facilities">
+				<position x="left" y="350"/>
 				<size width="250" height="14"/>
 				<alignment horizontal="left" vertical="top"/>
 				<font>smalfont</font>
 			</label>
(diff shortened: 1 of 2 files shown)


Quickmind01:
Could this be what I'm seeing on stream? It seems like my game is running more slowly and has been for sometime.

Kurtsley:
Yes, your save has almost 7 thousand space liners parked.

Quickmind01:
Well, I would agree with having the check for this automatically set for saves, then. I can't see willingly allowing something like this to occur as being a positive experience. Being able to turn it off can be for those who want to mess around with something silly like what we had show up on stream with dozens of liners trying to escape X-COM's wrath.

Kurtsley:
Luckily this problem has already been fixed, so this won't be a problem with any games moving forward. After thinking about it, I don't think an automatic fix should be put in for a problem that won't exist after this point.

FilmBoy84:
It certainly shouldn't be automatic on save load, though this option is useful for helping people with old saves fix them and therefore should be implemented in the debug form that it is

I've done some rudimentary tests and all seems well

Merging

Quickmind01:
Looks like there may be an issue with transferring a save from this build to the previous release build I've been using. I took my save over to this test build 20231228-2-gf6a78f7c to clear out the extra space liners, and then moved my save back to my regular build 20231228. Within a minute of in-game time, I get this error 
image

The game then CTD. I am using the EWM and music mods.
[save_Superhuman 1.zip](https://github.com/OpenApoc/OpenApoc/files/13801269/save_Superhuman.1.zip)

Quickmind01:
Just tested the test build, and the same thing happens.

Quickmind01:
FilmBoy and I discovered that this fix is only partial. Once I made a save using this build, I had to go into the save to remove the references to these vehicles in the .save\gamestate\vehicles.xml file. Once removed from there, the crash didn't occur and the game ran fine.

Issue #1340, opened 2024-02-13
Option: Automatically reload ammo only of the same type


dethstryke01:
This situation demonstrates why I think this feature would be nice. 

https://youtu.be/mA_C0VO3cpo?t=10196

steveschnepp:
What's the original game behavior?

dethstryke01:
I just ran a quick test, and it looks like the original game automatically reloads with the same type of ammo as its first preference, and then checks for any available ammo and automatically reloads that as well. I think it would be a nice quality of life feature if OpenApoc had an option that would specify that the gun should only be reloaded with the original ammo.  As a player I prefer AP ammo quite a lot over high explosive or incendiary ammo, and I also tend to prefer to rush to close range rather than sitting back and shooting. With that setting turned on I could bring all three ammo types as options without needing to worry about setting my own people on fire accidentally at close range.

FilmBoy84:
This should be resolved in #1356

Please open new issues regarding any problems with multiple ammo types in the late game referencing both this issue ( #1340 ) and #1356

PR #1356, opened 2024-02-27
Add option to only auto reload same ammo type.

Kurtsley:
Addresses #1340.

Weapons with multiple ammo types should only reload the previously loaded type or the weapon is considered out of ammo and must be manually loaded.

Tested lightly, did not seem to mind weapons with one ammo type or none such as the mind bender. Probably should be run through a late game scenario with many different weapon types. The problem was the "lastLoadedAmmoType" didn't exist unless the weapon was manually loaded in battle. This forced me to assign a value to it if the weapon didn't need to be reloaded at the start of the battle. This could in theory cause issues with different ammo types.

### DIFF
--- a/framework/options.cpp
+++ b/framework/options.cpp
@@ -138,6 +138,7 @@ void dumpOptionsToLog()
 	dumpOption(optionATVUFOMission);
 	dumpOption(optionMaxTileRepair);
 	dumpOption(optionSceneryRepairCostFactor);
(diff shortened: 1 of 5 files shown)


FilmBoy84:
Had some plays with this and it seems to work

I'll merge and report any further issues as separates, referencing this PR should they appear

Thanks again Kurtsley :beers:

PR #1357, opened 2024-02-27
Adds building collapse mechanic after a certain percentage of damage.

Kurtsley:
Adds a feature where building will collapse when enough tiles are destroyed.

This is the second attempt, first was outdated. I have the number set at 50 percent at the moment, obviously that can be adjusted.

This still needs to be moddable, I still haven't figured out how to get this value into the serialized gamestate.

### DIFF
--- a/game/state/city/building.cpp
+++ b/game/state/city/building.cpp
@@ -1102,4 +1102,13 @@ bool Building::occupied() const
 
 bool Building::isAlive() const { return countActiveTiles() > 0; }
 
+bool Building::isStable() const
+{
+	if ((float)this->buildingParts.size() / this->initialParts * 100.0 < this->integrityPercentage)
+	{
+		return false;
+	}
+	return true;
+}
+
 } // namespace OpenApoc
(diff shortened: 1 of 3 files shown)


Kurtsley:
Ah, I thought I this was part of the OG. I will just drop this then, but will check out why the garages aren't destroyed.

FilmBoy84:
No worries

I actually saw some potential in it as a new feature (a mid ground difficulty increase between OG and having "raid collapses buildings" on 😄 )

I'll add to the list of potential future things to make proper design proposals for as i think there's something in there that you've hit upon as a future optional feature

Yeah, the heart of the building and garage issue has been strange for a while - you may have seen @quickmind01 stumped by it on his stream. Garages and Hangars are heavily armoured in the OG, making them harder to destroy - but the current debug keys do not destroy them as part of building collapse and it's extremely hard to destroy them in normal play or regular fall damage from collapsing buildings.

This may relate to known current issues relating to damage calcs and ticks, but I have been unable to get clear data on it

That said, in debug terms having the existing function to collapse a building enough to "disable" it but not destroy garages is useful in some ways for certain tests. But, the function really needs to do exactly what it says and collapse it all

PR #1358, opened 2024-02-28
Fix civilians appearing as hostile on battlescape.

Kurtsley:
Addresses #849.

Removes hostile unit spotted messages from civilians and prevents the selection box from targeting civilians as hostile.

The previous attempt to fix didn't work because the check for civilian AI didn't work correctly when they were panicking, which they do quite often. This doesn't touch the negative effects of killing civilians, it will still hurt your status with the civilian org.


FilmBoy84:
Brillaint, thanks for this fix Kurtsley!
LGTM, merging :beers:

Issue #1350, opened 2024-02-21
Removing the engine of a vehicle at a second base without storage causes a crash


dethstryke01:
https://youtu.be/AFcoQnEF-Yc?t=302

[save_save 136.zip](https://github.com/OpenApoc/OpenApoc/files/14359641/save_save.136.zip)

Window 10 x64 22H2
OpenApoc version 0.0.930
no mods in use
[settings.zip](https://github.com/OpenApoc/OpenApoc/files/14359648/settings.zip)

Kurtsley:
I found that this isn't a storage issue per se, it's a fuel issue. This can be replicated at your own base right at the start of a game. Steps from start of game:

- Sell all your fuel immediately
- Send a craft on a trip to burn fuel
- When it gets back, remove engine and re-attach
- Error is thrown

It looks like the engine has no owner, thus the failed to reload message fails.

FilmBoy84:
This should be partially resolved by #1359 so closing

However, there remains an issue with no message being presented to the player for a lack of fuel at base we need further testing and information on as it presents - please open or append existing dedicated issues regarding this as incidents occur

PR #1359, opened 2024-02-28
Fix crash when trying to refuel at base with no fuel.

Kurtsley:
Addresses #1350.

Interesting bug, the game was trying to push a NotEnoughFuel event but was unable to because the engine had no owner. This only happened at a base with 0 fuel and the vehicle's engine was removed and replaced. It no longer crashes, but the NotEnoughFuel message still does not show up. There has been instances of messages not appearing in the past so this may be a separate issue.


FilmBoy84:
Definitely an odd one, thanks for the fix

With regards to the lack of a message regarding new fuel - agreed, there's likely something else at play

Hopefully, further playthroughs from our testers will give some leads :)

Issue #1296, opened 2023-12-26
Items Disappearing When Dropped on Equipment


Quickmind01:
Certain items can "eat" other items dropped on them while in the backpack. It appears to mainly be 2x3 items. Some items, the bottom right corner would trigger the dropping item to disappear, while some of the other squares could do it as well. Please see this video: https://youtu.be/JbxlEiftqcs

Prior to recording the video, I was getting grenades to disappear on the Devastator Cannon, but the video didn't show this so I'm not sure what the difference is. Weapons that use clips seem to be resistant to this issue from what I can remember. This also may lead into another issue related to inventory handling. When attempting to place an item over another in the backpack, sometimes the item will be placed in its previous location, sometimes it'll be placed next to the item you tried to place it over, and sometimes, it just goes on the ground. It appears to be based on where the first item is placed, specifically, how far to the right. This is somewhat shown in the video as well.

This is with build 20231214B along with EWM mod music mod.

Quickmind01:
Follow-up: the Marsec Body Unit also causes this issue with the medi-kit in the bottom right corner.

FilmBoy84:
Oh that's super weird, thanks for report

If one of us gets a save of this, please post here :

BeornTB:
Replicated: IIRC I had just lost a cloaking device by misplacing it atop a stun grapple.

[save_Lost Cloak Item Bug.zip](https://github.com/OpenApoc/OpenApoc/files/13796888/save_Lost.Cloak.Item.Bug.zip)
[log.txt](https://github.com/OpenApoc/OpenApoc/files/13796889/log.txt)
[settings.zip](https://github.com/OpenApoc/OpenApoc/files/13796890/settings.zip)

dethstryke01:
Medkits also disappear when the attempt to equip them misses slightly. 

https://youtu.be/gAMibszWYPQ?t=3408

[save_medkit disappears.zip](https://github.com/OpenApoc/OpenApoc/files/14412032/save_medkit.disappears.zip)
[save_save 179.zip](https://github.com/OpenApoc/OpenApoc/files/14412033/save_save.179.zip)

FilmBoy84:
Should be resolved by #1360

PR #1360, opened 2024-03-01
Fix various equipment placement issues.

Kurtsley:
Addresses #1296.

I was poking around at the agent equip screen code and I may have accidentally fixed the disappearing item problem. I noticed that grenades could be loaded into weapons, so I fixed that by only allowing ammo type equipment to be loaded (sounds obvious).

https://github.com/OpenApoc/OpenApoc/blob/0864ff5ccc0d88969f234dfaf7df3a239144c9e3/game/ui/general/aequipscreen.cpp#L1445

I believe this line in the adding ammo code was causing some items to disappear. Items like the medkit have 0 ammo while grenades have 1 ammo. Any item that got to this point in the code with 0 ammo would be nullified, causing it to disappear forever. The grenades, having 1 ammo, were loaded into the weapon, causing the logerror message.

While this needs more testing with many different item combinations, it seems to fix any issues I was seeing on my end.

### DIFF
--- a/game/ui/general/aequipscreen.cpp
+++ b/game/ui/general/aequipscreen.cpp
@@ -1414,6 +1414,8 @@ bool AEquipScreen::tryPlaceItem(sp<Agent> agent, Vec2<int> slotPos, bool *insuff
 			{
 				canAdd = true;
 				slotPos = offsetPosition;
+				equipmentUnderCursor =
+				    std::dynamic_pointer_cast<AEquipment>(agent->getEquipmentAt(slotPos));
 				break;
 			}
 		}
@@ -1442,6 +1444,7 @@ bool AEquipScreen::tryPlaceItem(sp<Agent> agent, Vec2<int> slotPos, bool *insuff
 		else
 		{
 			equipmentUnderCursor->loadAmmo(*state, draggedEquipment);
+
 			if (draggedEquipment->ammo > 0)
 			{
 				if (draggedEquipmentOrigin.x != -1 && draggedEquipmentOrigin.y != -1 &&


game/ui/general/aequipscreen.cpp
```diff
@@ -1434,14 +1434,15 @@ bool AEquipScreen::tryPlaceItem(sp<Agent> agent, Vec2<int> slotPos, bool *insuff
 	if (canAdd)
```
JonnyH, line 1434: If the theory is true about disappearing equipment - how can canAdd be true? Surely that's only true if the slot is empty (so not reloading), or if it's in the ammo_types array (as checked at line 1375)?

Kurtsley: I'm going to post the tryPlaceItem method here easy reference, I think I have the logic mostly figured out for these errors:

https://github.com/OpenApoc/OpenApoc/blob/63ef6567a5eb49c006ac93859c3e6c21027039e9/game/ui/general/aequipscreen.cpp#L1369-L1466

Here is what I think is happening for two different issues. First, if a grenade or medkit is placed over a weapon, it will try to load it like a clip.

- 1372 - agent->canAddEquipment returns false, finds equipment under the dragged equipment at this position.
- 1379 - dragged equipment is not the same type as the ammo type of the weapon, continues to 1399.
- 1399 - checks for a nearby offset position to place the item, finds one. Sets canAdd to true and sets slotPos to new offset position.
- 1434 - since canAdd is true, checks for equipmentUnderCursor. Importantly, this was set *before* the offset was found. Since that is the case, it thinks there is still an item in the slot and moves to 1444.
- 1444 - Immediately loads the equipment into the weapon,  error is thrown.

In this case, it seems the equipmentUnderCursor variable should be checked again if an offset is found.

Second, if an item is dragged close to a slot but misses, and the item dissapears.

- 1372 - agent->canAddEquipment returns false, finds equipment under the dragged equipment at this position.
- equipmentUnderCursor is not a weapon, continues to 1399.
- 1399 - checks for a nearby offset position to place the item, finds one. Sets canAdd to true and sets slotPos to new offset position.
- 1434 - Since canAdd is true, checks for equipmentUnderCursor. Finds the left arm armor (in my testing in one specific slot). Moves to 1444 since equipmentUnderCursor was found.
- 1444 - Loads equipment as ammo, nothing happens since no weapon in slot.
- 1445 - checks the dragged equipment for ammo greater than 0, since a medkit has 0 ammo, continues to 1464.
- 1464 - Nullifies the dragged equipment, removing the item forever.

Again, it seems the equipmentUnderCursor needs to be rechecked. I did not look into this as deeply when I made my initial PR. I now see that my fix worked because it ignored the equipmentUnderCursor check that was using the original position.

Thoughts on this assessment?

FilmBoy84: Spent some time going through this morning, and your assessment does appear to be correct
The fix does resolve a number of issues

PR #1365, opened 2024-03-03
Fix battleunit pathfinding with unset memory

JonnyH:
The wrapper  BattleUnitTileHelper::canEntierTile function never initialized some values, which could be used by the child function, so could cause valid paths to be rejected due to junk memory values.


FilmBoy84:
Thanks for this Jonny

I will aim to do some thorough testing over next few days and see if anything breaks

FilmBoy84:
Have played an hour or so this morning and it appears to work as intended
Thanks Jonny
Merging

Issue #1316, opened 2024-01-21
Add a Debug Option to destroy selected vehicle on map or tab


FilmBoy84:
It's not always easy to remove vehicles with the current debug tool to destroy them, this should probably be extended to destroy the selected vehicle in the tab or map too

PR #1366, opened 2024-03-03
Add debug option to destroy selected vehicles, clean up debug options

Kurtsley:
Addresses #1316.

Adds an option (minus key in debug mode) to destroy selected vehicles on the map or in the vehicle tabs. This replaces the clear space liners debug option as it shouldn't be an issue moving forward. The unfreeze vehicle in cityscape option has also been removed for the same reason.


FilmBoy84:
LGTM and appears to function as intended
Thanks for this <3

Issue #1287, opened 2023-12-23
The Optional OpenApoc Feature to Crash Vehicles instead of destroy them needs to explain that modules may be lost in recovered vehicles


BeornTB:
- OS: Windows 10 x64 2020H2
 - OpenApoc Version : 20231222
 - List of Mod files in use : none

Vehicles crash and lose items to the evil tow company mafia. Even if you tow them yourself with a valkyrie. QM talked about it on stream, pretty sure this is already reported and everything. This example has a lost accuracy module on a Griffon.

Save
[save_Crashed Lost Weapon.zip](https://github.com/OpenApoc/OpenApoc/files/13760400/save_Crashed.Lost.Weapon.zip)

Log
https://www.dropbox.com/scl/fi/42pekg5fyp0yl8k9hohg5/log.txt?rlkey=rxfrlce3ikdbe1exkg57ny0xf&dl=0

Settings
[settings.zip](https://github.com/OpenApoc/OpenApoc/files/13760401/settings.zip)

FilmBoy84:
This was a "feature" of the Crash recovery system - to ensure that X-COM doesn't unnecessarily benefit from turning the option on, and to encourage marketplace investment in new modules crashed vehicles will lose various bits of equipment as "destroyed"

This probably needs explaining in the tooltip for the option but otherwise is NOT a bug or issue per-se - simply a balancing function for allowing a feature cut from the original game

Kurtsley:
Any suggestions for wording of the tooltip?

FilmBoy84:
I'm not sure how much text space we have - there are FOUR options affecting crashes, one for unsupported ground vehicles, one for vehicles not capable of dimension jump, one for low fuel and one for low HP

I propose appending existing tooltips with `- Weapons and Modules may be lost in crash`

Unsupported Ground Vehicles Crash - Weapons and Modules may be lost in crash
![image](https://github.com/OpenApoc/OpenApoc/assets/33913088/97d017b7-8a19-4eb1-b0ce-dfef549d8e0a)

Uncapable Vehicles Crash when entering Gates - Weapons and Modules may be lost in crash
![image](https://github.com/OpenApoc/OpenApoc/assets/33913088/5bfeb7f3-751d-4e96-87db-29402c3f8870)

Vehicles Crash on Low HP - Weapons and Modules may be lost in crash
[...]

PR #1367, opened 2024-03-03
Add message when module lost on recovery.

Kurtsley:
Address #1287.

Adds a SomethingDiedEvent message when a module is scrapped after a vehicle is recovered. Adds a disclaimer to the tooltips for crashing vehicles explaining the risks.

One example, all four have the same message added.
Screenshot 2024-03-03 160500


Quickmind01:
Great addition!

FilmBoy84:
This works really nicely
No issues found in testing
Merging

Issue #1299, opened 2023-12-27
Day Number Did Not Increment


Quickmind01:
The save file shows I am on Day 27. When I complete the day and save again, it still shows Day 27. The next day then shows Day 28. I don't believe I used any time advancing debug commands during this play through and then saving some time later. This is with build 20231214B along with EWM mod music mod. The Battle save is where you need to start. Complete the mission, and then complete the day to see this bug. The Date Bug save is the next day where the bug takes place.
[Day After 27 is 27.zip](https://github.com/OpenApoc/OpenApoc/files/13774482/Day.After.27.is.27.zip)

FilmBoy84:
I cannot replicate this

Starting with the mission save, i finish the mission, send the griffons in the city home and use debug to crash all the space liners that are slowing the game (I believe this save was made before a fix was implemented)

The game proceeds past 2nd April and moves to 3rd April via the end of week as expected

![image](https://github.com/OpenApoc/OpenApoc/assets/33913088/d19064e1-b5fb-48f7-b420-550e04bc5785)
![image](https://github.com/OpenApoc/OpenApoc/assets/33913088/0af0e065-1b0b-4bf8-9c32-0bcd04aa1bf5)
![image](https://github.com/OpenApoc/OpenApoc/assets/33913088/aecec085-1b66-40c6-b38c-bba59b73bcc3)

Save before midnight of 2nd - 3rd April attached

[save_Before Day End 1.zip](https://github.com/OpenApoc/OpenApoc/files/14446553/save_Before.Day.End.1.zip)

Quickmind01:
I can check this later (getting ready to head to bed), but what I'm referring to is the in the save name information where it shows what Day you're on. It appears the number doesn't increment if you save before allowing any time to pass. I thought it was still occurring in the current version, but I could be wrong.

Quickmind01:
Here is what I was referring to. Day 67 Test should show Day 68, but because time hasn't been allowed to pass, it still shows Day 67. This was from loading Superhuman 1.
![image](https://github.com/OpenApoc/OpenApoc/assets/51022165/f5fa8e4e-9214-4887-b0e1-d3bc363ed569)

Kurtsley:
I see what you mean, the score screen shows the following day but the next day doesn't happen until at least one second ticks by. In OG, the new day actually starts at 00:00.

FilmBoy84:
This appears resolved by #1368

PR #1368, opened 2024-03-05
Fix day starting after 00:00.

Kurtsley:
Addresses #1299.

The new day was starting at 00:01 instead of right at 00:00. This also had an effect on setting the correct hire dates for new hires. If anyone can see a reason this was done originally, let me know.

### DIFF
--- a/game/state/gametime.cpp
+++ b/game/state/gametime.cpp
@@ -174,7 +174,7 @@ unsigned int GameTime::getLastDayOfCurrentMonth() const
 	return getPtime(this->ticks).date().end_of_month().year_month_day().day;
 }
 
-unsigned int GameTime::getDay() const { return (this->ticks + TICKS_PER_DAY - 1) / TICKS_PER_DAY; }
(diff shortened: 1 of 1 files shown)


FilmBoy84:
IIRC The reason for this change was to remedy a host of problems with the End of Day calculations which only update on the first tick of the new day - but it was only ever a temporary solution, so it would be good to rectify it

(You may have seen the text "TICK TOCK" appear on the clock when switching to the alien dimension in play too which was introduced to allow something to be processed - in the OG this is covered with a dedicated loading screen for travelling to the dimension)

I'll aim to do some thorough testing to see what this change does

It may also be interesting to check if this has any effect on the economy bug and associated LUA script

Kurtsley:
It seems like this getDay method is used for checking hire dates for agent and for save games. Actions like updateEndofDay in gamestate use the dayPassed flag, which checks ticks directly. The original PR for this fixed the save game days starting at 7, not 1.

Issue #1070, opened 2021-10-27
Items in the marketplace do not replenish each week and do not adhere to min/max rules specified in economy.xml including if cheats are used to change time


FilmBoy84:
**Describe the bug**
There appears to be an issue restocking the marketplace with items each week - it does not happen - it is most noticeable when playing with mods as there are naturally more items to not be replenished - but it happens in unmodified games too

The economy.xml specifices a minimum stock level and a maximum stock level for each week - RNG influenced by supply/demand from the organisations in the city as well as relations should affect how many are produced (check OG logic) and then the results of that calculation given as the number in stock for that week

This is currently a serious issue for players as they will find, unless they have researched "recharging" weapons, they soon run out of ammunition

Craft also suffer from ammunition shortages and shortages of fuel and elerium that can stop game progression if the player has not got the research items for making their own craft (which have infinite fuel) or alien craft weapons (that recharge)

For that reason, this economy bug is GAME-BREAKING

**To Reproduce**
1. Buy some agent equipment from the marketplace - if using mods, ensure some of this equipment is from those mods - they seem worse afflicted by this bug
![image](https://user-images.githubusercontent.com/33913088/139066865-2fc8f9d8-de4a-466b-b8f6-e47845e371eb.png)
2. Play forward one week (or cheat a week forward) - playing elimates any cheat issues as the cause
3. Go to the marketplace and try to buy the same items - none will be in stock and only any remaining stock at X-COM will be available for sale. New Items appearing that week will also not be in stock
![image](https://user-images.githubusercontent.com/33913088/139067694-017b7e74-795e-4c22-bda7-c4583fbfe982.png)

**Expected behavior**
The marketplace should restock each week obeying the min/max limits and according to the logic used by the OG for economical and relationship effects on RNG between the min/max values

**Desktop (please complete the following information):**
 - OS: Windows 10 x64 2020H2
 - OpenApoc Version : 0.386-52 - both with and without the Extended Weapons Mod (i.e. both vanilla and modded openapoc)

Istrebitel:
Cannot reproduce on latest build:
1. Start new game
2. Do nothing, skip on fastest time for a week
3. Use F1+X to crash UFOs that spawn
4. After week elapses, economy updates appropriately

[...]

PR #1369, opened 2024-03-06
Fix weekly stock not updating under certain conditions

Kurtsley:
Addresses #1070.

I think I have tracked down the issue. I believe it is an error in the update_ecomony.lua script that deals with other orgs buying Xcom produced goods each week.

Based on the lua error I was seeing plus some trial and error, this only happened when there were Xcom produced goods on the market in small quantities. It could be reproduced by:

- At game start, quickly produce some bio transport modules.
- Sell at least one.
- Skip to the end of the week.
- Error is thrown (60 percent of the time), some items don't restock.

In order to get this to happen every time, line 23 can be changed to if r < 100.

When the currentStock is set to lastStock * 80 / 100, or * 66 / 100, the resulting number is a float if the currentStock is small enough. I believe this is the root cause of the error. By rounding down to an integer, the script seems to function as intended.

If you sell an even nice even number like 10, there is no error as 10 * 80 / 100 = 8. Any whole number will work just fine.

Depending on how far the script gets, it looks like some items still restock. There is a 40% chance that these calculations will not happen, leading to this issue being infrequent and hard to track down.

This save that was provided to me helped in testing:
[save_MarketExchange_d6_into_d7.zip](https://github.com/OpenApoc/OpenApoc/files/14504113/save_MarketExchange_d6_into_d7.zip)

### DIFF
--- a/data/scripts/update_economy.lua
+++ b/data/scripts/update_economy.lua
@@ -21,9 +21,9 @@ function updateEconomy()
 			--update stock
 			local r = GS.rng:randBoundsInclusive(0, 100)
 			if r < 30 then
-				eco.currentStock = eco.lastStock * 80 / 100
+				eco.currentStock = math.floor(eco.lastStock * 80 / 100)
 			elseif r < 60 then
-				eco.currentStock = eco.lastStock * 66 / 100
+				eco.currentStock = math.floor(eco.lastStock * 66 / 100)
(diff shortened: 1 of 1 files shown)


FilmBoy84:
Fantastic work! This does indeed appear to fix this hard to nail down issue
Merging

PR #1371, opened 2024-03-13
Ask if uses want to overwrite save file when creating save with existing name

### LINKED ISSUE #1347: When making a save file with the same name as one that already exists, rather than asking "overwrite save file yes/no?" the game sometimes creates a new file with random extra numbers
(Note: large parts of this mystery are resolved in the next comment)

With 146 save files in the folder (184 total files including github zips), I've started noticing that saves I've made have failed to appear in the save file folder, or they do appear with random numbers added to the end. A save called "before week end 2" disappeared, and a save that should have been called "after week end 2" ended up being called "after week end 21181" somehow. 

Footage of myself making these saves and the buggy save file:

https://youtu.be/3F5is1O6n0U?list=PLdWM6woGnOQfDw-LC_E_CtQs7I12fSUrr&t=2997
[save_after week end 21181.zip](https://github.com/OpenApoc/OpenApoc/files/14323496/save_after.week.end.21181.zip)

Here's the last save file I made:
[save_104.zip](https://github.com/OpenApoc/OpenApoc/files/14323604/save_104.zip)

Window 10 x64 22H2
OpenApoc version 0.0.930
no mods in use
[settings.zip](https://github.com/OpenApoc/OpenApoc/files/14323529/settings.zip)

ayrtondenner:
Fix #1347

After discussion in issue thread, I implemented the following logic:

- If writing name in first slot
  - If no savefile exists with same name, then create savefile
  - If savefile found, open overwrite messagebox
- If writing in existing slot, open overwrite messagebox

I also encapsulated overwrite messagebox call in `SaveMenu::askUserIfWantToOverrideSavedGame`

### DIFF
--- a/game/state/savemanager.cpp
+++ b/game/state/savemanager.cpp
@@ -171,9 +171,9 @@ bool SaveManager::findFreePath(UString &path, const UString &name) const
 
 sp<SaveMetadata> SaveManager::getSaveGameIfExists(const UString &name) const
 {
-	auto saveList = getSaveList();
+	const auto saveList = getSaveList();
 	auto it = std::find_if(saveList.begin(), saveList.end(),
-	                       [&name](SaveMetadata &obj) { return obj.getName() == name; });
+	                       [&name](const SaveMetadata &obj) { return obj.getName() == name; });
 
 	if (it != saveList.end())
 	{
--- a/game/state/savemanager.cpp
+++ b/game/state/savemanager.cpp
@@ -163,23 +163,27 @@ bool writeArchiveWithBackup(SerializationArchive *archive, const UString &path, 
 
 bool SaveManager::findFreePath(UString &path, const UString &name) const
 {
-	path = createSavePath("save_" + name);
-	if (fs::exists(path))
-	{
-		for (int retries = 5; retries > 0; retries--)
-		{
-			path = createSavePath("save_" + name + std::to_string(rand()));
-			if (!fs::exists(path))
-			{
-				return true;
-			}
-		}
+	path = createSavePath(name);
+	bool pathExists = fs::exists(path);
 
-		LogError("Unable to generate filename for save %s", name);
-		return false;
+	return !pathExists;
+}
+
+SaveMetadata SaveManager::getSaveGameIfExists(const UString &name) const
+{
+	auto saveList = getSaveList();
+	auto it = std::find_if(saveList.begin(), saveList.end(),
+	                       [&name](SaveMetadata *obj) { return obj->getName() == name; });
+
+	if (it != saveList.end())
+	{
+		return *it;
 	}
 
-	return true;
+	//UString path;
+	//return !findFreePath(path, name);
+
+	return {};
 }
 
 bool SaveManager::newSaveGame(const UString &name, const sp<GameState> gameState) const
--- a/game/state/savemanager.cpp
+++ b/game/state/savemanager.cpp
@@ -169,21 +169,19 @@ bool SaveManager::findFreePath(UString &path, const UString &name) const
 	return !pathExists;
 }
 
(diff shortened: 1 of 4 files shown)


game/ui/general/savemenu.cpp
```diff
@@ -284,42 +284,66 @@ void SaveMenu::tryToLoadGame(sp<Control> slotControl)
...
-
-		fw().stageQueueCommand({StageCmd::Command::PUSH, messageBox});
+		auto slot = parent->getData<SaveMetadata>();
+		SaveMenu::askUserIfWantToOverrideSavedGame(slot);
 	}
 }
 
+void SaveMenu::askUserIfWantToOverrideSavedGame(sp<SaveMetadata> saveMetadata)
+{
+	auto saveName = saveMetadata->getName();
```
JonnyH, line 321: Generally, if something *can* be const, I think it should be declared as such.

game/state/savemanager.cpp
```diff
@@ -163,23 +163,25 @@ bool writeArchiveWithBackup(SerializationArchive *archive, const UString &path,
...
-		}
+	path = createSavePath(name);
+	bool pathExists = fs::exists(path);
 
-		LogError("Unable to generate filename for save %s", name);
-		return false;
+	return !pathExists;
+}
+
+sp<SaveMetadata> SaveManager::getSaveGameIfExists(const UString &name) const
```
JonnyH, line 172: Not related to your change, as it's just following what's already there, but I really need to remove the sp<> wrapper here - it's not *really* a thing that needs shared ownership and arguably a bit unnecessary.

game/ui/general/savemenu.cpp
```diff
@@ -284,42 +284,66 @@ void SaveMenu::tryToLoadGame(sp<Control> slotControl)
 
 void SaveMenu::tryToSaveGame(const UString &saveName, sp<Control> parent)
 {
+	// If saving new item from first row
 	if (parent->Name == newSaveItemId)
 	{
-		if (saveManager.newSaveGame(saveName, currentState))
+		auto saveGameMetadata = saveManager.getSaveGameIfExists(saveName);
```
JonnyH, line 290: can be const?

game/ui/general/savemenu.cpp
```diff
@@ -284,42 +284,66 @@ void SaveMenu::tryToLoadGame(sp<Control> slotControl)
...
+		    {
+			    fw().stageQueueCommand({StageCmd::Command::POP});
+		    }
+		    else
+		    {
+			    clearTextEdit(activeTextEdit);
+		    }
+	    });
+
+	std::function<void()> onNo = std::function<void()>([this] { clearTextEdit(activeTextEdit); });
```
JonnyH, line 338: Can be auto? The type is already defined in the rhs

game/state/savemanager.cpp
```diff
@@ -163,23 +163,25 @@ bool writeArchiveWithBackup(SerializationArchive *archive, const UString &path,
...
+{
+	auto saveList = getSaveList();
```
JonnyH, line 174: these can be const?

game/ui/general/savemenu.cpp
ayrtondenner: You mean in the parameter, like this?

```c++
void SaveMenu::askUserIfWantToOverrideSavedGame(const sp<SaveMetadata> saveMetadata)
```

ayrtondenner:
@JonnyH I added some more const values, let me know if I should missed some.

JonnyH:
Thanks, looks good.

I think c++ really should be "const-by-default", and force you to state if variables are mutable. I just find it makes it *much* easier to reason about code if you don't have to worry about variables changing :)

Issue #1302, opened 2023-12-27
Alien Containment accounting: 2x each item and Brainsucker Pod = 80%


BeornTB:
- OS: Windows 10 x64 2020H2
 - OpenApoc Version 20231225-3
 - No mods

**Describe the bug**
Alien containment sometimes counts items in 2 categories that each take their own separate % of space. Also, Brainsucker Pods currently count as 80% of containment space _each_.

![screenshot002](https://github.com/OpenApoc/OpenApoc/assets/46300745/c22b6f35-0d37-46a4-b22a-966f9e060234)

FilmBoy84:
For anyone working on this issue, please see also issue:

#369 
#608 

And a potential way to improve handling overflows when implementing the OG "Alien Containment is full" screen with extras

#1203

BeornTB:
Oh and the containment wasn't tared at 0%, once everything in it was incinerated, the balance was at -20% (or was it -19%...). So 6 brainsucker pods at 80% each (!) would be 480% space utilization, but the empty containment was below 0%, and the numbers don't add up to 480 either.

Issue #608, opened 2019-06-15
Alien Containment List


Quickmind01:
In my alien containment, there are two entries for Dead Skeletoids and Dead Micronoid Aggregates. Brainsucker pods are also listed and take 80% space.
[save_Stream Test.zip](https://github.com/OpenApoc/OpenApoc/files/3293375/save_Stream.Test.zip)

Quickmind01:
Perhaps the pods are bugged in the Sell menu, but they are taking 80% containment space. Also, the Dead Skeletoid and Dead Micronoid are still listed in the alien containment, but are grayed out since 0 are left.

makus82:
i get this )) 999%
![image](https://user-images.githubusercontent.com/11581207/59554608-2dfabf80-8fae-11e9-846e-c221e167d3e4.png)

FilmBoy84:
Confirmed, brainsucker pods should NOT be in the alien containment and take a stupid amount of space there

RELATED ISSUES #369 AND #370

Here's just one example of how empty i have to keep my alien containment
![image](https://user-images.githubusercontent.com/33913088/63095429-1ebed080-bf63-11e9-9f68-404392e07ba0.png)

PR #1372, opened 2024-03-15
Remove brainsucker pod from alien containment

ayrtondenner:
This partially solves issues #1302 and #608, since brainsucker pod is not being considered in either alien containment usage calculation or aliens listing, but we still have duplicated aliens.

I also tried to simplify usage percentage math.


FilmBoy84:
Tested and everything appears where it should

Thanks for the fix 🍻 
screenshot002
screenshot004
screenshot005
screenshot006

PR #1374, opened 2024-03-16
Removed duplicated items from alien containment list

### LINKED ISSUE #1302: Alien Containment accounting: 2x each item and Brainsucker Pod = 80%

ayrtondenner:
Fixes #1302 and #608

Aliens in containment list were appearing duplicated whenever an alien type exists in more than one base.

Before:
image

After:
image

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -3366,7 +3366,7 @@ void Battle::exitBattle(GameState &state)
 	}
 
 	// Give player vehicle a null cargo just so it comes back to base once
-	for (auto v : playerVehicles)
+	for (auto v : returningVehicles)
 	{
 		v->cargo.emplace_front(
 		    state, StateRef<AEquipmentType>(&state, state.agent_equipment.begin()->first), 0, 0,
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -3366,7 +3366,7 @@ void Battle::exitBattle(GameState &state)
 	}
 
 	// Give player vehicle a null cargo just so it comes back to base once
-	for (auto v : returningVehicles)
+	for (auto v : playerVehicles)
 	{
 		v->cargo.emplace_front(
 		    state, StateRef<AEquipmentType>(&state, state.agent_equipment.begin()->first), 0, 0,
--- a/game/ui/base/transactionscreen.cpp
+++ b/game/ui/base/transactionscreen.cpp
@@ -386,8 +386,12 @@ void TransactionScreen::populateControlsAlien()
 {
 	int leftIndex = getLeftIndex();
 	int rightIndex = getRightIndex();
+	auto &alienTransactionControls = transactionControls[type];
+
 	for (auto &ae : state->agent_equipment)
 	{
+		auto &alienTypeName = ae.first;
+
 		if (!ae.second->bioStorage)
 		{
 			continue;
@@ -395,7 +399,16 @@ void TransactionScreen::populateControlsAlien()
 		// Add alien
 		for (auto &b : state->player_bases)
 		{
-			if (b.second->inventoryBioEquipment[ae.first] > 0)
+
+			auto it =
+			    std::find_if(alienTransactionControls.begin(), alienTransactionControls.end(),
+			                 [&alienTypeName](const sp<TransactionControl> &transactionControl)
+			                 { return transactionControl->itemId == alienTypeName; });
+
+			// Removing brainsucker pod from alien containment list
+			if (b.second->inventoryBioEquipment[ae.first] > 0 &&
+			    alienTypeName != "AEQUIPMENTTYPE_BRAINSUCKER_POD" &&
+			    it == alienTransactionControls.end())
 			{
 				auto control = TransactionControl::createControl(
 				    *state, StateRef<AEquipmentType>{state.get(), ae.first}, leftIndex, rightIndex);
@@ -403,7 +416,7 @@ void TransactionScreen::populateControlsAlien()
 				{
 					control->addCallback(FormEventType::ScrollBarChange, onScrollChange);
 					control->addCallback(FormEventType::MouseMove, onHover);
-					transactionControls[type].push_back(control);
+					alienTransactionControls.push_back(control);
 				}
 			}
 		}
--- a/game/ui/base/transactionscreen.cpp
+++ b/game/ui/base/transactionscreen.cpp
@@ -398,7 +398,7 @@ void TransactionScreen::populateControlsAlien()
 		// Add alien
 		for (auto &b : state->player_bases)
 		{
-			auto alienTypeExistingControl = findControlById(alienTypeName);
+			auto alienTypeExistingControl = findControlById(type, alienTypeName);
 
 			if (b.second->inventoryBioEquipment[ae.first] > 0 &&
 
--- a/game/ui/base/transactionscreen.cpp
+++ b/game/ui/base/transactionscreen.cpp
@@ -398,15 +398,10 @@ void TransactionScreen::populateControlsAlien()
 		// Add alien
 		for (auto &b : state->player_bases)
 		{
+			// Don't add an alien type if it's already added in transaction controls
 			auto alienTypeExistingControl = findControlById(type, alienTypeName);
 
-			if (b.second->inventoryBioEquipment[ae.first] > 0 &&
-
-			    // Removing brainsucker pod from alien containment list
-			    alienTypeName != "AEQUIPMENTTYPE_BRAINSUCKER_POD" &&
-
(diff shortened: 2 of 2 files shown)


FilmBoy84:
I have tested this today
Appears to work
Though i am having trouble finding where brainsucker pods now are (since removed from alien containment list)
They no longer appear to be in Agent Equipment list either (either on agent equip or agent equipment market screens); despite two launchers being recovered (which do appear)

Going to do further testing before i merge as im not sure if pods now disappear into the void

FilmBoy84:
Okay, a little delay in getting saves to you, but here is what I have discovered

There is an agent in this mission of the ORIGINAL GAME holding a brainsucker pod, keep him alive and finish the mission... On returning to base, the brainsucker pod is placed in ALIEN CONTAINMENT and is available for research

[SAVEGAME 1.zip](https://github.com/OpenApoc/OpenApoc/files/14711133/SAVEGAME.1.zip)
image

In OpenApoc, under the build for PR, again, please find attached a savegame with an agent that has a brainsucker pod in their inventory, please finish the mission and you will see on return from that mission the pod neither appears in Alien Containment OR Agent inventory - IT SHOULD DO but this PR sadly makes it disappear completely
[Test.zip](https://github.com/OpenApoc/OpenApoc/files/14711377/Test.zip)
Win64 build: https://ci.appveyor.com/project/OpenApoc/openapoc/builds/49418718/artifacts

ayrtondenner:
Wait, so when a brainsucker pod should appear in the Alien Containment list? Only when it's alive?

FilmBoy84:
There is no pod autopsy, so yes, pods appear in the Alien Containment as they can only be "alive".

The problem referred to in #1302 and several linked issues is that for some reason Brainsucker Pod (Alive) was taking up too much space in Alien Containment rather than it should not be there

I need to check what happens when the pod is researched though - I have a nagging feeling research then adds a "Brainsucker (Alive)" to the Alien Containment on completion research of which should then return a "Brainsucker (Dead)" to the Alien Containment

But that sounds too easy for getting live brainsuckers, so i really need to investigate it as it may be the case that researching the live pod just results in it's destruction (unless you want to test with the Original Game save i provided above first?)

ayrtondenner:
Ok, so should I reverse changes made in https://github.com/OpenApoc/OpenApoc/pull/1372?
Because then Brainsucker Pods will re-appear.

And then I will have to check the values being used for Brainsucker Pods when calculating storage usage for Alien Containment.

FilmBoy84:
Agreed, I'm starting to think we may have to revert #1372 too

Although before anything we need to check the OG logic to save further increasing the workload with anything else missed regarding pods (they are fairly unique in being ammo, equipment and a live alien)

Apologies again, it was my bad in testing #1372 - I assumed removing them from Alien Containment was needed, tested that it removed them, and failed to realise that they are mean to be in there just not consume all the capacity they do 😓

ayrtondenner:
image
Each Brainsucker Pod is taking 11 units of storage space, 11 * 16 = 176 units of storage space - a Alien Containment cell has 20 units of storage space.

ayrtondenner:
> Although before anything we need to check the OG logic to save further increasing the workload with anything else missed regarding pods (they are fairly unique in being ammo, equipment and a live alien)


Just to confirm, I used the save file you provided and the Brainsucker life cycle is like you described: Brainsucker Pod -> Brainsucker (Alive) -> Brainsucker (Dead)

> Apologies again, it was my bad in testing https://github.com/OpenApoc/OpenApoc/pull/1372 - I assumed removing them from Alien Containment was needed, tested that it removed them, and failed to realise that they are mean to be in there just not consume all the capacity they do 😓

Checking in OG game and all Brainsucker stages take 1 slot of Alien Containment.

game/state/battle/battle.cpp
```diff
@@ -3366,7 +3366,7 @@ void Battle::exitBattle(GameState &state)
 	}
 
 	// Give player vehicle a null cargo just so it comes back to base once
-	for (auto v : returningVehicles)
+	for (auto v : playerVehicles)
```
JonnyH, line 3369: If this is a different issue maybe a different PR is useful?

Keeping one-logical-chance-per-PR often helps later for reading logs or bisecting

game/ui/base/transactionscreen.cpp
```diff
@@ -386,18 +386,27 @@ void TransactionScreen::populateControlsAlien()
...
+
 		if (!ae.second->bioStorage)
 		{
 			continue;
 		}
 		// Add alien
 		for (auto &b : state->player_bases)
 		{
-			// Removing brainsucker pod from alien containment list
+			auto alienTypeExistingControl = findControlById(type, alienTypeName);
```
JonnyH, line 401: Can be const?

game/ui/base/transactionscreen.cpp
```diff
@@ -386,18 +386,27 @@ void TransactionScreen::populateControlsAlien()
 {
 	int leftIndex = getLeftIndex();
 	int rightIndex = getRightIndex();
+
 	for (auto &ae : state->agent_equipment)
 	{
+		auto &alienTypeName = ae.first;
```
JonnyH, line 392: Can be const?

game/ui/base/transactionscreen.cpp
```diff
@@ -386,18 +386,27 @@ void TransactionScreen::populateControlsAlien()
...
 		// Add alien
 		for (auto &b : state->player_bases)
 		{
-			// Removing brainsucker pod from alien containment list
+			auto alienTypeExistingControl = findControlById(type, alienTypeName);
+
 			if (b.second->inventoryBioEquipment[ae.first] > 0 &&
-			    ae.first != "AEQUIPMENTTYPE_BRAINSUCKER_POD")
+
+			    // Removing brainsucker pod from alien containment list
```
JonnyH, line 405: I don't really like having a complex enough conditional so that comments are needed in the middle - can make a little hard to read.

Maybe this can be separated out so each is evaluated separately? Might need to reverse the comparisons?



e.g. something like:
```
if (b.second->inventoryBioEquipment[ae.first] == 0)
 continue;

//Skip brainsuckers
if (alienTypeName == "AEQUIPMENTYPE_BRAINSUCKER_POD")
  continue;

// Skip if it's already added
if (alienTypeExistingControl)
  continue;

// Create the control
auto control = ... etc etc etc
```

Does that make the logic clearer?

JonnyH (changes requested):

game/ui/base/transactionscreen.cpp
JonnyH: Though some things on this list may be removed by #1381 anyway :P

game/state/battle/battle.cpp
ayrtondenner: This was wrongly copied from https://github.com/OpenApoc/OpenApoc/pull/1379

game/ui/base/transactionscreen.cpp
ayrtondenner: Even if the variable will have a new value for every loop run, is still advised to create it as const?

JonnyH: If it's not changed after declaration, I'd mark it "const".

I generally see that const *should* be the default, as it means you know at a glance it won't change value later in that block.

game/state/battle/battle.cpp
ayrtondenner: Done, this now is reverted.

game/ui/base/transactionscreen.cpp
ayrtondenner: Updated to const:

https://github.com/OpenApoc/OpenApoc/blob/12f960b12bf8b7d40fd104f027fedb298ccdcd30/game/ui/base/transactionscreen.cpp#L401

game/ui/base/transactionscreen.cpp
ayrtondenner: Updated to const:

https://github.com/OpenApoc/OpenApoc/blob/12f960b12bf8b7d40fd104f027fedb298ccdcd30/game/ui/base/transactionscreen.cpp#L390-L392

game/ui/base/transactionscreen.cpp
ayrtondenner: I reworked the condition to avoid having so many nested conditions. I also removed the outdated "Brainsucker Pod" reference:

https://github.com/OpenApoc/OpenApoc/blob/12f960b12bf8b7d40fd104f027fedb298ccdcd30/game/ui/base/transactionscreen.cpp#L399-L416

PR #1375, opened 2024-03-16
Checks if trade state is active, hides it from rendered list if not.

ayrtondenner:
Fixes https://github.com/OpenApoc/OpenApoc/issues/608#issuecomment-502384799

Before rendering transaction control list, set if they are visible based on index and stock conditions. Also isolated condition checks at `Trade.isActive()`

Before:
image

After:
image

Obs.: as you can see in the second image, the scrollbar is rendered even if the aliens list is smaller than the transaction view.

### DIFF
--- a/game/ui/general/transactioncontrol.h
+++ b/game/ui/general/transactioncontrol.h
@@ -94,6 +94,18 @@ class TransactionControl : public Control
 		int getBalance() const { return getRightStock(true); }
 		// ScrollBar support. Set current value.
 		int setBalance(const int balance);
+
+		/// <summary>
+		/// Checks if trade is active or not. Can be used to determine to draw shades, remove from
+		/// item list, etc.
+		/// </summary>
+		/// <returns>If trade is active or not.</returns>
+		bool isActive() const
+		{
+			auto isActive = (getLeftIndex() == getRightIndex() ||
+			                 (getLeftStock() == 0 && getRightStock() == 0)) == false;
+			return isActive;
+		}
 	};
 
   private:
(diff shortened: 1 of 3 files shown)


FilmBoy84:
This appears to work, but i cannot test fully until #1374 is resolved

game/ui/general/transactioncontrol.h
```diff
@@ -94,6 +94,18 @@ class TransactionControl : public Control
...
 		int setBalance(const int balance);
+
+		/// <summary>
+		/// Checks if trade is active or not. Can be used to determine to draw shades, remove from
+		/// item list, etc.
+		/// </summary>
+		/// <returns>If trade is active or not.</returns>
+		bool isActive() const
+		{
+			auto isActive = (getLeftIndex() == getRightIndex() ||
```
JonnyH, line 105: Maybe this logic can be split out?

if (getLeftIndex() == getRightIndex())
  return false;
if (getLeftStock == 0 && getRightStock() == 0)
  return false;
return true;

ayrtondenner: Logic improved

https://github.com/OpenApoc/OpenApoc/blob/f8c0609d18e468c19a91894ab58f33132f74d01c/game/ui/general/transactioncontrol.h#L103-L112

XCOM-HUB:
This is a good idea for the Alien Containment, but not so good when it comes to the items you can Buy/Sell. Just look at what happened when I bought up everything in the Ground vehicle equipment section: 

Skärmbild 2024-04-18 031942
I would prefer the OG way when it comes to items, not with aliens though.

PR #1381, opened 2024-03-22
Removing filters that removed Brainsucker Pod references.

ayrtondenner:
After discussion in https://github.com/OpenApoc/OpenApoc/pull/1374, it was decide to revert changes made in https://github.com/OpenApoc/OpenApoc/pull/1372

Now Brainsucker Pod is being listed again at Alien Containment screen:

image


FilmBoy84:
Thanks for this, tested and merging :beers:

PR #1382, opened 2024-03-22
Fix crash on "Vehicle manufacturing halted" message

### LINKED ISSUE #1378: Workshop Crash on Not Enough Funds
- OpenApoc Version: _20240315_0_g3273d585
 - List of Mod files: Extended Weapons Mod 8.1

**Describe the bug**
I have a Workshop at Base 5, that is about to finish their work on `Retaliator`. There are more `Retaliator`s in production queue. There is not enough money to start the next production cycle.

**To Reproduce (Scenario A)**
1. Load the enclosed save game;
2. Unpause the game;
3. At 05:00 game time, the game freezes, the sound stutters;
4. In several seconds (real time), the game crashes with no error message and nothing useful dropped in log.

**To Reproduce (Scenario B)**
1. Load the enclosed save game;
2. Go to Base 5, Large Workshop 1. Cancel the production order;
3. Unpause the game;
4. The game proceeds as expected.

**To Reproduce (Scenario C)**
1. Load the enclosed save game;
2. Go to Base 5, Large Workshop 1. Reduce the ordered amount to `1`;
3. Unpause the game;
4. There is a 'Reassign the Workshop' notification. The game proceeds as expected.

**Expected behavior**
In Scenario A, the production should halt and the warning message should appear when there is not enough funds to continue; just as it happens in Scenarios B and C. 

**Savegame**
The compressed save is included, 5 minutes (game time) before the crash. I'm also including the log file, but it seems of little use.[log.zip](https://github.com/OpenApoc/OpenApoc/files/14672971/log.zip)
[653.zip](https://github.com/OpenApoc/OpenApoc/files/14672975/653.zip)

**Additional context**
Note that this only happens to the `Retaliator` order: the other manufacturing orders fail gracefully.

JonnyH:
The event contained a VehicleType ID, but was indexing the Vehicle list to get the name, which wasn't correct.

Should fix #1378

### DIFF
--- a/game/ui/tileview/cityview.cpp
+++ b/game/ui/tileview/cityview.cpp
@@ -4341,7 +4341,7 @@ bool CityView::handleGameStateEvent(Event *e)
 					item_name = game_state->agent_equipment[ev->topic->itemId]->name;
 					break;
 				case ResearchTopic::ItemType::Craft:
-					item_name = game_state->vehicles[ev->topic->itemId]->name;
+					item_name = game_state->vehicle_types[ev->topic->itemId]->name;
 					break;
 			}
 			setUpdateSpeed(CityUpdateSpeed::Pause);


JonnyH:
Another reason for IDs to be more strongly typed, so using a VEHICLETYPE ID in a VEHICLE array wouldn't build :P

Issue #1319, opened 2024-01-28
Add the City Repair Options to the More Options setting screen


FilmBoy84:
As reported in #1188 there are a number of issues getting the recently merged Incremental City Repair features on to the More Options Screen / Form

In addition to this, the relevant options and their values are not added to the configuration file along with the other options on first run / change. This needs to be implemented.

ayrtondenner:
Checking the code, I see that the problem starts with both new options being read as bool, which is not the case:

https://github.com/OpenApoc/OpenApoc/blob/319676dc5ed5391731df2f72384a2ac0a902823c/framework/options.cpp#L437-L444

I tried to programmatically find a way to determine each options type based on their declaration, but since I couldn't do it, I created a list with int and float options to check when the default bool approach is not the right one:

```c++
	// By default, city options list treat all options as boolean values
	// But we have some exceptions with different value types that needs to be checked
	const std::list<UString> cityNotificationsIntList = {"MaxTileRepair"};
	const std::list<UString> cityNotificationsFloatList = {"SceneryRepairCostFactor"};

	// FIXME: Can this be optimized?
	for (auto &p : *citynotificationList)
	{
		const UString full_name = p.first + "." + p.second;

		const auto isOptionInt =
		    std::find(cityNotificationsIntList.begin(), cityNotificationsIntList.end(), p.second) !=
		    cityNotificationsIntList.end();

		if (isOptionInt)
		{
			continue;
		}

		const auto isOptionFloat =
		    std::find(cityNotificationsFloatList.begin(), cityNotificationsFloatList.end(),
		              p.second) != cityNotificationsFloatList.end();

		if (isOptionFloat)
		{
			continue;
		}

		auto checkBox = mksp<CheckBox>(fw().data->loadImage("BUTTON_CHECKBOX_TRUE"),
		                               fw().data->loadImage("BUTTON_CHECKBOX_FALSE"));
		checkBox->Size = {240, citylistControl->ItemSize};
		checkBox->setData(mksp<UString>(full_name));
		checkBox->setChecked(config().getBool(full_name));
		auto label = checkBox->createChild<Label>(tr(config().describe(p.first, p.second)), font);
		label->Size = {216, citylistControl->ItemSize};
		label->Location = {24, 0};
		label->ToolTipText = tr(config().describe(p.first, p.second));
		label->ToolTipFont = font;
		citylistControl->addItem(checkBox);
	}
```
[...]

PR #1386, opened 2024-03-24
Add City Repair settings to More Options screen

ayrtondenner:
This is a partial fix addressing #1319

1. Add MaxTileRepair and SceneryRepairCostFactor settings and its values to Cityscape Options list
2. Unifying identic logic instructions related to Battlescape Options and Cityscape Options

image

As you can see, this PR is just a partial fix since it does NOT allow the update of these settings values. But since this already adds a lot of updates, and since it seems that to complete this task it will be necessary to create a new form for numeric values - int and float - I decided to push this PR before doing any further work on this.

Obs.: I created lists to store settings name that are not bool, but either int or float. Please let me know if there's any better way to get these settings value instead.


FilmBoy84:
Tested and changed values for visual feedback confirmation

All appears good, will merge this and leave #1319 open until a means to modify values is provided

image

Issue #1387, opened 2024-03-28
Renaming and overwriting old save files in game will not rename the save file


XCOM-HUB:
- OS: Windows 11 Pro x64 23H2
 - OpenApoc Version: 0.0.991
 - No Mods

I can no longer overwrite new text on my old save files in game. You can create a new save file and the text will be saved, but it will not allow any more name changes after that. This wasn't a problem before.

PS: The save files you overwrite work as usual, it's just that you can't change the text/rename anymore.

ayrtondenner:
Gonna test with an older version, maybe this is something that changed after my PR https://github.com/OpenApoc/OpenApoc/pull/1371 that updated the save system regarding to file name

ayrtondenner:
Before any update on this matter, I was wondering when exactly we want to rename savefiles instead of overwriting them?

Update https://github.com/OpenApoc/OpenApoc/pull/1371 was done because before it, whenever the user tried to create a new savefile with an already existing name, it ended up creating a new save file with an existing name + random suffix. Right now, the game will check if the name already exists and asks if the user wants to overwrite the existing savefile in the following situations:

1. If the user selects the existing savefile and tries to save without changing its name
2. When the user types the name for a new savefile but ends up typing the same name as an already existing one

I want to have more info on that before writing any update.

ayrtondenner:
Just tested with an older version to be sure that it was renaming a save file before, it was indeed the case. I'm thinking about updating it with the following logic:

1. If the user selects an older row from an existing savefile, ask if it should overwrite if; if yes, then delete the savefile with the older name and create a new one with the new name - effectively renaming it.
2. If the user clicks in the "new" row but writes the name of an existing savefile, ask if it can overwrite it; if yes, then save new file with the same name, overwriting it.

I guess this is the logic that makes more sense.

PR #1393, opened 2024-04-02
Issue 1387: restored renaming savefile feature

ayrtondenner:
Fixes https://github.com/OpenApoc/OpenApoc/issues/1387

PR https://github.com/OpenApoc/OpenApoc/pull/1371 removed feature of renaming existing save files. In this PR it is being added again with the following behavior:
- When selecting an existing savefile and saving with same name, the savefile will be overwritten
- When selecting an existing savefile and saving with a different name, the savefile will be overwritten and renamed with new name
- When creating a new savefile with same name as existing one, the existing savefile will be overwritten

Minor refactorings:
- Updated `getSaveGameIfExists` to return an `optional` instead of `shared_ptr`, and updated references properly
- Updated some parameters and variable declarations with `const` and `auto` references whenever possible

### DIFF
--- a/game/ui/general/savemenu.cpp
+++ b/game/ui/general/savemenu.cpp
@@ -304,7 +304,7 @@ void SaveMenu::tryToSaveGame(const UString &saveName, const sp<Control> parent)
 		// If the user use the first row but wrote the name of an already existing save game
 		else
 		{
-			SaveMenu::askUserIfWantToOverrideSavedGame(saveGameMetadata);
+			SaveMenu::askUserIfWantToOverrideSavedGame(saveGameMetadata, saveName);
 		}
 	}
 
@@ -312,23 +312,25 @@ void SaveMenu::tryToSaveGame(const UString &saveName, const sp<Control> parent)
 	else
 	{
 		const auto slot = parent->getData<SaveMetadata>();
-		SaveMenu::askUserIfWantToOverrideSavedGame(slot, true);
+		SaveMenu::askUserIfWantToOverrideSavedGame(slot, saveName);
 	}
 }
 
 void SaveMenu::askUserIfWantToOverrideSavedGame(const sp<SaveMetadata> saveMetadata,
-                                                const bool deleteOlderSaveGame)
+                                                const UString &saveName)
 {
-	const auto &saveName = saveMetadata->getName();
+	const auto &metadataSaveName = saveMetadata->getName();
 	const auto messageBoxTitle = "Override saved game";
-	const auto messageBoxContent = "Do you really want to override " + saveName + "?";
+	const auto messageBoxContent = "Do you really want to override " + metadataSaveName + "?";
 
 	auto onYes = std::function<void()>(
-	    [this, saveMetadata, saveName, deleteOlderSaveGame]
+	    [this, saveMetadata, saveName]
 	    {
-		    const auto savingResult = saveManager.saveGameViaView(saveMetadata, currentState, deleteOlderSaveGame);
+		    const auto savingResult =
+		        saveManager.overrideGame(*saveMetadata, saveName, currentState);
 
-		    savingResult ? fw().stageQueueCommand({StageCmd::Command::POP}) : clearTextEdit(activeTextEdit);
+		    savingResult ? fw().stageQueueCommand({StageCmd::Command::POP})
+		                 : clearTextEdit(activeTextEdit);
 	    });
 
 	auto onNo = std::function<void()>([this] { clearTextEdit(activeTextEdit); });
--- a/game/ui/general/savemenu.cpp
+++ b/game/ui/general/savemenu.cpp
@@ -304,7 +304,7 @@ void SaveMenu::tryToSaveGame(const UString &saveName, const sp<Control> parent)
 		// If the user use the first row but wrote the name of an already existing save game
 		else
 		{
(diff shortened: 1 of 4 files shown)


game/ui/general/savemenu.cpp
```diff
@@ -304,23 +304,24 @@ void SaveMenu::tryToSaveGame(const UString &saveName, const sp<Control> parent)
...
+		SaveMenu::askUserIfWantToOverrideSavedGame(slot, saveName);
 	}
 }
 
-void SaveMenu::askUserIfWantToOverrideSavedGame(const sp<SaveMetadata> saveMetadata)
+void SaveMenu::askUserIfWantToOverrideSavedGame(const sp<SaveMetadata> saveMetadata,
+                                                const UString &saveName)
 {
-	const auto &saveName = saveMetadata->getName();
+	const auto &metadataSaveName = saveMetadata->getName();
```
JonnyH, line 322: While we're here, this sort of string should be wrapped in tr() to allow translate.

Probably best to have a separate PR, but don't want to miss stuff like this :)

game/ui/general/savemenu.h
```diff
@@ -36,11 +36,12 @@ class SaveMenu : public Stage
 	void loadWithWarning(sp<Control> parent);
 	// these functions will display prompt to execute action
 	void tryToLoadGame(sp<Control> slotControl);
-	void tryToSaveGame(const UString &textEdit, sp<Control> parent);
+	void tryToSaveGame(const UString &textEdit, const sp<Control> parent);
 	void tryToDeleteSavedGame(sp<Control> &control);
 
 	// Opens pop-up asking the user if wants to override existing saved game
-	void askUserIfWantToOverrideSavedGame(sp<SaveMetadata> saveMetadata);
```
JonnyH, line 43: I'm not sure if SaveMetadata needs to be wrapped in an sp<> - it's just a stack of strings, copying it around would be easier and simpler?

JonnyH: Making getSaveGameIfExists() et al return an optional<SaveMetadata> might be more obvious in it's intent?

ayrtondenner: Could you please show a suggestion regarding how to return this as an optional?

JonnyH: Again it's probably not an issue for this PR, as it's already using an sp<>, but it doesn't seem like it's necessary.

For things that *may* return a SaveMetadata, instead of using an sp<SaveMetadata> and comparing it against nullptr, a std::optional<SaveMetadata> is probably a more fitting abstraction to how it's actually used.

e.g. you can do something like:
```
std::optional<SaveMetadata> get_save()
{
  if (save_exists)
    return SaveMetadata{save, metadata,. arguments};
  else
   return {}; // Returns an "empty" optional
}

void do_other_things_with_save(const SaveMetadata &metadata) // You *must* have a valid SaveMetadata already, so no need for this to be std::optional<SaveMetadata>.
{
....whatever
}

void doThingWithSave()
{
  auto optional_metadata= getSave();
  if (optional_metadata)
  {
     // get_save() returned an optional<> containing a SaveMetadata
     // operator* returns the underlying object of an optional<>
     const auto &metadata = *metadata;
     do_other_things_with_save(metadata);
  }
  else
  {
    // get_save() returned  an "empty" optional
  }
}
```

an sp<> is a reference counted pointer to a heap allocation that then contains the SaveMetadata (or a nullptr) - but that seems unnecessary if the SaveMetadata doesn't actually *need* shared ownership reference counting....

game/ui/general/savemenu.cpp
ayrtondenner: Added `tr()` call to both message box title and content:

```c++
	const auto messageBoxTitle = tr("Override saved game");
	const auto messageBoxContent = tr("Do you really want to override " + metadataSaveName + "?");
```

game/ui/general/savemenu.h
ayrtondenner: Updated `SaveManager::getSaveGameIfExists` return type to `std::optional<SaveMetadata>`:

```c++
std::optional<SaveMetadata> SaveManager::getSaveGameIfExists(const UString &name) const
{
	const auto saveList = getSaveList();
	const auto it =
	    std::find_if(saveList.begin(), saveList.end(),
	                 [&name](const SaveMetadata &obj) { return obj.getName() == name; });

	if (it != saveList.end())
	{
		return *it;
	}

	return {};
}
```

Also updated `SaveMenu::askUserIfWantToOverrideSavedGame` parameters to use `const SaveMetadata &saveMetadata` instead of shared pointer.

Thanks for the help, really appreciate it!

game/ui/general/savemenu.cpp
JonnyH: tr() uses the passed in string as a key - so using "metadataSaveName" as part of that likely makes key it's looking for include the save name, so probably won't work.

I think the "correct" way would be to use a format string replacement instead of string concatination - IE something like "messageBoxContent = format(tr("Do you really want to override %s?"), metadataSaveName)" 

so the translation replacement string is "Do you really want to override %s?", and the translated string also has a replacement tag so still works as expected.

ayrtondenner: Updated:

```c++
	const auto &metadataSaveName = saveMetadata.getName();
	const auto messageBoxTitle = tr("Override saved game");
	const auto messageBoxContent =
	    format(tr("Do you really want to override %s?"), metadataSaveName);
```

Issue #1312, opened 2024-01-01
[QoL] Optional Feature to Prefix None-X-COM vehicle names with a universal symbol to set them apart from player units


FilmBoy84:
It became apparent on @quickmind01 's Stream of the 31st December 2023 that it is not always easy to differentiate X-COM Vehicles from Organisation Vehicles by name alone as the vehicles are assigned name by `<Vehicle Type> <Number>`

For this reason, a QoL option to allow NONE-X-COM craft to be prefixed with a universal character supported by the game fonts should be implemented

I propose that X-COM Craft continue to be named as normal, though none-XCOM craft are prefixed by a `*`

This will break immersion in a minimal manner and should only be presented as an optional feature under the `More Options-->Cityscape` list in game

Thus in an example where 324 is the vehicle number, the X-COM Hawk Air Warrior would appear as `Hawk Air Warrior 324` but a none-X-COM Hawk would appear as `* Hawk Air Warrior 324`

Quickmind01:
Simply having an asterisk is a welcome improvement and would be acceptable for a first step. Unless epic battles are anticipated where you'll want to know who owns each vehicle, having more than one symbol may be unnecessary, but an option for more would be nice. Thank you for adding this request.

BeornTB:
How about only renaming the XCom craft with the prefix : "X-COM Hawk Warrior 1" and that way none of the immersion is affected, and your ownership is emphasized?

FilmBoy84:
> How about only renaming the XCom craft with the prefix : "X-COM Hawk Warrior 1" and that way none of the immersion is affected, and your ownership is emphasized?

The main issue is string length - text often goes out of bounds with more than a single character addition

The Original Game just numbered the craft for X-COM in the range 1-99 after the vehicle type counting from the first craft that X-COM acquired from that type and counting up

Three digit numbers were almost exclusively other organisations

So we should probably try and replicate that too

SirPlusse:
Potential compromise between owner prefixes and string length: Triple digits don't seem to be an issue, so maybe make X-COM craft have a default designation along the lines of (craft name) X25? Like, preface the number with an X so you know it's yours? You'd probably need to strike an extremely fine balance between performance and attrition to go above 50 or so, so triple-digit X-COM craft are unlikely to be an issue here.

Though yes, just putting an asterisk (or an X, to abbreviate Beorn's alternative) at the start would probably be more eye-catching...

FilmBoy84:
[...]

PR #1405, opened 2024-04-13
Added "VehiclesPrefix" option

ayrtondenner:
Added "Vehicle prefix" option picker at More Options view, like discussed in #1312
It's just the visual part, there's no functionality implemented yet.

image

### DIFF
--- a/framework/options.cpp
+++ b/framework/options.cpp
@@ -140,6 +140,7 @@ void dumpOptionsToLog()
 	dumpOption(optionSceneryRepairCostFactor);
 	dumpOption(optionLoadSameAmmo);
 	dumpOption(optionShowCurrentDimensionVehicles);
+	dumpOption(optionVehiclesPrefix);
 
 	dumpOption(optionStunHostileAction);
 	dumpOption(optionRaidHostileAction);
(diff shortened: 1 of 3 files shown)


FilmBoy84:
Works nicely, i'll hold off from merging this until the feature is implemented though

The prefix probably needs to be translatable using characters already supported by the game
For example
`tr("*")`

ayrtondenner:
@FilmBoy84 I was thinking about the following: what if I update the option string to inform that this is only a visual display and not a fully implemented feature as of now, and then you merge this into main?

My worry is that someone starts implementing this without checking this PR and ends up re-implementing this visual part.

FilmBoy84:
Understandable - I have commented on #1312 accordingly

My only hangup is merging incomplete stuff could cause problems - we all know that end users tend not to read and post errors first (if at all)

@JonnyH your thoughts on merging this?
I can see arguments either way

JonnyH:
I'm normally against merging things that aren't used - it's an easy way for things to get forgotten and cruft accumulate with nothing actually testing it.

And it looks pretty separable, there isn't really much downside to holding it in a branch and possibly re-merging it when the thing is wired up?

Kurtsley:
Maybe you should be assigned to the issue so people know you are working on it.

FilmBoy84:
> Maybe you should be assigned to the issue so people know you are working on it.

Done, good call :)

PR #1408, opened 2024-04-14
Refactored TransactionControl::setScrollbarValues()

### LINKED ISSUE #1398: changing base to expansion causes new shipment when changing base with pending transaction on main base
when changing base from first to second with pending transaction from first to market that exceeds the total number in storage in second base

setMaximum is called that clamps present order qty to new max since it was greater

https://github.com/OpenApoc/OpenApoc/blob/37206e25f6b9a2b129efc9919960deca0be51a45/game/ui/general/transactioncontrol.cpp#L75

setValue does not return false and instead sets a new value to the transaction control which flags it to trigger the form event change callback
https://github.com/OpenApoc/OpenApoc/blob/37206e25f6b9a2b129efc9919960deca0be51a45/forms/scrollbar.cpp#L36
https://github.com/OpenApoc/OpenApoc/blob/37206e25f6b9a2b129efc9919960deca0be51a45/forms/scrollbar.cpp#L67

in the callback lambda function updateValues() calls setBalance()
https://github.com/OpenApoc/OpenApoc/blob/37206e25f6b9a2b129efc9919960deca0be51a45/game/ui/general/transactioncontrol.cpp#L697
https://github.com/OpenApoc/OpenApoc/blob/37206e25f6b9a2b129efc9919960deca0be51a45/game/ui/general/transactioncontrol.cpp#L122
https://github.com/OpenApoc/OpenApoc/blob/37206e25f6b9a2b129efc9919960deca0be51a45/game/ui/general/transactioncontrol.cpp#L912

the shipments map is modified, creating a new transaction.
https://github.com/OpenApoc/OpenApoc/blob/37206e25f6b9a2b129efc9919960deca0be51a45/game/ui/general/transactioncontrol.cpp#L921

to reproduce 
in base 1 create transaction to destroy a large number of brainsucker pods that is larger than the amount in storage at second base.
change base to second base.
control will show a new transaction to destroy all brainsucker pods in storage

can be fixed by not setting the value within setMaximum, since the value is set by the calling function immediately afterwards.

ayrtondenner:
Fixes #1398 

While I was debugging this issue, I found out that only reworking value calculation for `scrollBar->setMaximum()` with variable declaration already fixed the error. To be honest, I don't know exactly how this fixed the problem, since it doesn't seem to have any difference at all, but the reality is that now this function seems to be working properly.

If anyone has any feedback about how this variable declaration changed the outcome of it, I would be happy to hear. 

Video examples:
- Before fix: https://youtu.be/RPfieRXkVD8
- After fix: https://youtu.be/dyMGoDuLHE8

### DIFF
--- a/game/ui/general/transactioncontrol.cpp
+++ b/game/ui/general/transactioncontrol.cpp
@@ -63,18 +63,24 @@ void TransactionControl::initResources()
 
 void TransactionControl::setScrollbarValues()
 {
-	if (tradeState.getLeftIndex() == tradeState.getRightIndex())
+	auto leftStock = 0;
+	auto rightStock = 0;
+	auto balance = 0;
+
+	if (tradeState.getLeftIndex() != tradeState.getRightIndex())
 	{
-		scrollBar->setMinimum(0);
-		scrollBar->setMaximum(0);
-		scrollBar->setValue(0);
+		leftStock = tradeState.getLeftStock();
+		rightStock = tradeState.getRightStock();
+		balance = tradeState.getBalance();
 	}
-	else
-	{
-		scrollBar->setMinimum(0);
-		scrollBar->setMaximum(tradeState.getLeftStock() + tradeState.getRightStock());
-		scrollBar->setValue(tradeState.getBalance());
-	}
+
+	const auto minimum = 0;
+	const auto maximum = leftStock + rightStock;
+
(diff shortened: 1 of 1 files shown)


FilmBoy84:
Thanks for this, built and tested, LGTM :+1:

PR #1409, opened 2024-04-14
Refactoring vehicle list at CityView::update() for better readability

ayrtondenner:
I was taking a look at PR https://github.com/OpenApoc/OpenApoc/pull/1402 and I decided to refactor updates a little bit for better readability. It shouldn't have any behavior change otherwise.

- Added `const` modifier at declarations whenever possible
- Broke `if` condition for ignoring vehicles into different boolean declarations
- Updated `redo` condition to use an early exit instead of having long list of code inside `if` structure


FilmBoy84:
Built and tested, LGTM so merging now :+1: 

Thanks again for these recent PRs

Issue #1323, opened 2024-02-02
Lawpistol sometimes says "out of range" when it is out of ammo


dethstryke01:
Lawpistol says "out of range" at zero range, becomes broken. 

https://youtu.be/9lN0vJou2l0?t=3080

Window 10 x64 22H2
OpenApoc version 0.0.930
no mods in use

[settings.zip](https://github.com/OpenApoc/OpenApoc/files/14140894/settings.zip)

[save_test 2.zip](https://github.com/OpenApoc/OpenApoc/files/14140896/save_test.2.zip)

Kurtsley:
It looks like the pistol was out of ammo. I believe the hitbox shouldn't be any different when out of ammo, just the ticker on the bottom when you initially run out.

dethstryke01:
Totally agree, it's just out of ammo. I was confused because it said "out of range" and I forgot to check whether ammo might be a problem. Maybe the error in this situation should say "out of ammo" instead?

ayrtondenner:
Do you have any save in the battle where I can check this?

dethstryke01:
Unfortunately not at the moment, but I can find one easily the next time I'm debugging OpenApoc. It's not only the lawpistol but most weapons that say "out of range" when they're out of ammo, so I imagine it'd be pretty easy to spot just by emptying weapons of ammo. I've seen that one tons of times.

ayrtondenner:
I tried to replicate it with a save here but I got a "Out of ammo" message. Maybe it's something relate to time-based instead of real-time gameplay?

![image](https://github.com/OpenApoc/OpenApoc/assets/13112588/e90952a4-bf23-4aad-a0a4-1c690a2bd902)

dethstryke01:
It might be a turn based specific issue because I was playing exclusively turn based when I was looking for bugs in OpenApoc a while ago. I do remember that I saw the issue quite a few times with quite a few different weapons, but there may have been some other factor involved. Once I'm back in the world of OpenApoc I'll try to find a better example of the problem with a save file.

ayrtondenner:
I just tried with turn-based and also got only the Out of ammo message at the bottom. Maybe there's some other factors related to it.

dethstryke01:
Seems likely. I think it happened frequently, but not necessarily every time a weapon ran out of ammo.

PR #1417, opened 2024-04-17
Fix out of range message when out of ammo.

Kurtsley:
Addresses #1323 

This *should* fix this issue. The logic is:

If the weapon can't fire,
If the weapon isn't empty,
if the weapon is a launcher, no arc
else out of range.

This means the weapon will never display these other messages if out of ammo. Is this OG?

I find this situation difficult to test, so I can't tell if this will break other things.


FilmBoy84:
This was indeed a sod to replicate :rofl: 

But have achieved in modding by making a weapon that had a burst fire mode with the "wrong" multiple of ammo in the clip and an arcing projectile

Appears to work so merging

PR #1420, opened 2024-04-18
Prevent scrapping used modules.

Kurtsley:
Prevents passenger modules from being scrapped if the number of passengers would drop below what the vehicle could support. Prevents cargo modules from being scrapped if any cargo is on board.

Should this destroy extra cargo or passengers and add to the risk of recovery? I don't think so but it is something to think about for the future.


FilmBoy84:
> Should this destroy extra cargo or passengers and add to the risk of recovery? I don't think so but it is something to think about for the future.

For now,  probably not, I agree it's a potential option for the future - but probably one only for hardcore players and a proper working out of a system that enhances the user experience rather than simply frustrates would need some careful planning beforehand

Thanks for this fix, was a little tricky to test as requires a very specific crash situation, but LGTM and merging :beers:

PR #1427, opened 2024-04-28
Updating brainsucker launcher reference to use constant

ayrtondenner:
Quick update to use constant reference for brainsucker launcher instead of number.

### DIFF
--- a/tools/extractors/extract_agent_equipment.cpp
+++ b/tools/extractors/extract_agent_equipment.cpp
@@ -487,7 +487,7 @@ void InitialGameStateExtractor::extractAgentEquipment(GameState &state) const
 		}
 
 		// Mark brainsucker launcher
-		if (edata.sprite_idx == 44)
+		if (edata.sprite_idx == IT_BRAINSUCKERLAUNCHER)
 		{
 			e->launcher = true;
 		}


FilmBoy84:
Interesting, this seems to have stopped the BattleImagePack 44 errors i was getting in a recent run

Out of interest, does the dimension missile launcher and entropy launcher have any erroneous references where they use a number instead of const?

ayrtondenner:
> Interesting, this seems to have stopped the BattleImagePack 44 errors i was getting in a recent run
> 
> Out of interest, does the dimension missile launcher and entropy launcher have any erroneous references where they use a number instead of const?

For both launchers, I couldn't find a number reference in `extract_agent_equipment.cpp`, it seems that all references to them it's already using the defined constant:

https://github.com/OpenApoc/OpenApoc/blob/d139bf8c7fcbbe317309b1a63f16b805efd7a56c/tools/extractors/extract_agent_equipment.cpp#L72-L73

https://github.com/OpenApoc/OpenApoc/blob/d139bf8c7fcbbe317309b1a63f16b805efd7a56c/tools/extractors/extract_agent_equipment.cpp#L457-L460

https://github.com/OpenApoc/OpenApoc/blob/d139bf8c7fcbbe317309b1a63f16b805efd7a56c/tools/extractors/extract_agent_equipment.cpp#L425-L427

https://github.com/OpenApoc/OpenApoc/blob/d139bf8c7fcbbe317309b1a63f16b805efd7a56c/tools/extractors/extract_agent_equipment.cpp#L461-L465

I couldn't find any reference in this file to numbers 45 or 46. Are you having any similar error, but with another number?

FilmBoy84:
Thanks for response

Yes, the BattleImagePack error references 40, 44, 45, 46 and 48 IDs

I had narrowed this down to Disruptor Gun, Brainsucker Launcher, Entropy Launcher, Dimension Missile Launcher and either Vortex or Boomeroid Grenades

Curiously, the imagepack errors are only existent in the alien dimension maps and seem somewhat arbitrary as to the conditions they are thrown in

It's not easy to replicate - especially with the late game being a pre-requisite

ayrtondenner:
> Yes, the BattleImagePack error references 40, 44, 45, 46 and 48 IDs

I searched for these numbers and didn't find any reference to items with these IDs that could be replaced like I did in this PR with the brainsucker launcher. Maybe looking for the error message instead of these IDs would be more successfully.

PR #1428, opened 2024-04-29
Fix Brainsucker Pod store space

### LINKED ISSUE #1383: Brainsucker Pod taking 16 of storage space instead of 1
Brainsucker Pod is taking 16 of storage space at Alien Containment instead of 1, like in OG game.
Problem mentioned at issue https://github.com/OpenApoc/OpenApoc/issues/608 and also discussed at https://github.com/OpenApoc/OpenApoc/pull/1374, but I decided to create this issue to centralize discussion and information here.

image

I was trying to find where in the code this value is being set, but no success yet.

Save: [save_Stream Test.zip](https://github.com/OpenApoc/OpenApoc/files/14729096/save_Stream.Test.zip)

ayrtondenner:
Fixes #1383 (also mentioned in #608 and #1374)

- `tools/extractors/extract_agent_equipment.cpp`: agent equipment extractor is setting Brainsucker Pod store space as 16 via `tacp` / `data_t`, so here we set the correct store space of 1.
- `game/state/gamestate.cpp`: running DataExtractor updates `base_gamestate` values but it seems that this change is only reflected in new games, and older saved games have their own gamestate, is that correct? Because of that problem, it was also needed to update store space value from gamestate when reading an older savefile.

Now 1 Brainsucker Pod will account for only 5% of Alien Containment space (1/20 = 0.05)

image

### DIFF
--- a/game/state/gamestate.cpp
+++ b/game/state/gamestate.cpp
@@ -209,7 +209,14 @@ void GameState::initState()
 		{
 			w->ammo_types.emplace(this, t.first);
 		}
+
+		// Fixing brainsucker pod store space for older saves
+		if (t.first == "AEQUIPMENTTYPE_BRAINSUCKER_POD")
+		{
+			t.second->store_space = 1;
+		}
 	}
+
 	for (auto &a : this->agent_types)
 	{
 		a.second->gravLiftSfx = battle_common_sample_list->gravlift;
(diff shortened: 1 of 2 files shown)


FilmBoy84:
Yes saves carry their own gamestate for a host of reasons - not least, wider capability for save editing (you can modify data loaded in gamestate directly on a per-save basis) and compatibility with various mods (that use gamestate_patch then a comparison on load with base_gamestate and any data in an associated save)

PR #1431, opened 2024-05-06
Checking if vehicle missions list is empty before calling front()

### LINKED ISSUE #1430: "front() called on empty list" error when sending vehicle to point in map
1. Select a X-Com vehicle
2. Click "Go to map point"
3. Click somewhere in the map
4. Wait for the vehicle to reach selected point

Error raised when calling `.front()` in the following line:
https://github.com/OpenApoc/OpenApoc/blob/d139bf8c7fcbbe317309b1a63f16b805efd7a56c/game/state/city/vehicle.cpp#L183

Save file: [test-vehicle.zip](https://github.com/OpenApoc/OpenApoc/files/15227220/test-vehicle.zip)
Just load it and unpause, the error will be raised when selected vehicle reach its destination.

ayrtondenner:
Fixes #1430

Be aware that `list.front()` raises this error only when running the game in Debug mode.

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -180,7 +180,8 @@ class FlyingVehicleMover : public VehicleMover
 				dodge = 100;
 				break;
 		}
-		if (vehicle.missions.front().targetBuilding == vehicle.homeBuilding)
+		if (not vehicle.missions.empty() &&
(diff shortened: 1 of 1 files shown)


FilmBoy84:
Nice catch, confirmed and merging

Thanks for the fix :beers:

PR #1433, opened 2024-05-07
Added optional feature to show prefix for non-XCOM vehicles

### LINKED ISSUE #1312: [QoL] Optional Feature to Prefix None-X-COM vehicle names with a universal symbol to set them apart from player units

ayrtondenner:
Resolves #1312 and finishes #1405

Added optional feature that, when enabled, will add "*" prefix when displaying name of non-X-COM vehicles in alert message. Default option value is true.

Function `Vehicle::getFormattedVehicleNameForEventMessage(GameState &state)` will add prefix when option is true and when vehicle owner is not the player.

More Options view:
image

X-COM vehicle attacking construction vehicle:
destroyed-by

X-COM vehicle being attacked by enemy vehicle:
destroyed-by-2
image

### DIFF
--- a/framework/options.cpp
+++ b/framework/options.cpp
@@ -140,7 +140,7 @@ void dumpOptionsToLog()
 	dumpOption(optionSceneryRepairCostFactor);
 	dumpOption(optionLoadSameAmmo);
 	dumpOption(optionShowCurrentDimensionVehicles);
-	dumpOption(optionVehiclesPrefix);
+	dumpOption(optionShowNonXCOMVehiclesPrefix);
 
 	dumpOption(optionStunHostileAction);
 	dumpOption(optionRaidHostileAction);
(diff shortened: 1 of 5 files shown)


FilmBoy84:
Works nicely, appears in message ticker and message list as intended

image

Only noticeable thing is that it displays the prefix for UFOs that are destroyed but not for crashes or spot
image

I guess this could be explained as it's an "Unidentified Flying Object" until it's downed where the message then displays the craft type and prefixes it

Nonetheless, something probably for future debate and maybe a PR as this works and is a requested feature

Merging

PR #1436, opened 2024-05-08
Update getUsage type to float

ayrtondenner:
Since getUsage is returning the result of a division operation, it makes more sense the function return be float instead of int.

- Updated texts format and number comparisons to proper float type
- Added `const` and `auto` modifiers whenever possible
- Removed unnecessary `std::round` at getUsage

Obs.: as indicated by the branch name, my original idea was to convert it to double, but then I checked that all existing references to getUsage function are already using float.

### DIFF
--- a/game/state/city/base.cpp
+++ b/game/state/city/base.cpp
@@ -569,39 +569,37 @@ int Base::getCapacityTotal(FacilityType::Capacity type) const
 	return total;
 }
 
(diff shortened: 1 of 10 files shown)


FilmBoy84:
Very nice and good reasoning for the change
As with the other PR i'll aim to test this over next few days and see if anything breaks
LGTM on the changes though :)

FilmBoy84:
Couldn't see anything wrong in testing
Merging

Thanks for the PR 🍻

PR #1437, opened 2024-05-08
Add store space item info

### LINKED ISSUE #1429: Add store space value per item at Alien Containment stats view
Right now, Alien Containment view doesn't show in stats the store space used by an item. Example: Brainsucker Pod takes 1.

image

ayrtondenner:
Resolves #1429

Right now it only adds absolute value for store space. To display this information as percentage value, I will have to think a way to let the sheet view knows which screen is calling it, since the available space can come from item storage or alien containment. I guess it makes more sense to leave this for another PR in the future.

Files `aequipmentsheet.cpp` and `vehiclesheet.cpp`:
- Added store space as "Storage" in second line
- Refactored code to remove duplicated function that displayed info of unsearched items
- Updated label references for lines below Storage label

Alien Containment:
image
image

Vehicles:
image
image
image

Agent:
image
image

### DIFF
--- a/game/ui/general/aequipmentsheet.cpp
+++ b/game/ui/general/aequipmentsheet.cpp
@@ -42,22 +42,25 @@ void AEquipmentSheet::clear()
 }
 
 void AEquipmentSheet::displayImplementation(sp<AEquipment> item, const AEquipmentType &itemType,
-                                            bool researched)
+                                            const bool researched)
 {
-	if (!researched)
-	{
-		displayAlien(item, itemType);
-		return;
-	}
+	const auto itemName = researched
+	                          ? itemType.name
+	                          : (itemType.bioStorage ? tr("Alien Organism") : tr("Alien Artifact"));
 
-	form->findControlTyped<Label>("ITEM_NAME")->setText(itemType.name);
-	form->findControlTyped<Graphic>("SELECTED_IMAGE")
-	    ->setImage(item ? item->getEquipmentImage() : itemType.equipscreen_sprite);
+	const auto selectedImage =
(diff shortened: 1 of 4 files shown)


FilmBoy84:
Just started the review process on this

A couple of things i notice at a glance

`Base Storage Used` or a synonymous short string may be better than `Storage` as the current implementation reads as the item has that Storage capacity within it rather than it takes up that space in Base Storage. Suggestions welcome for something short enough to fit that's clear in showing what we are referring to

I notice the changes to the manner in which items are labelled as Alien Artefacts
Going to do some thorough testing with this that may take a while as I recall the mechanism has been problematic before - hopefully this resolves it, so exciting to see :)

ayrtondenner:
> `Base Storage Used` or a synonymous short string may be better than `Storage` as the current implementation reads as the item has that Storage capacity within it rather than it takes up that space in Base Storage.

I agree. I used "Storage" because that was the only word that came up to my mind as informative enough to explain the info while being short enough to fit in the label space.

> I notice the changes to the manner in which items are labelled as Alien Artefacts

While I refactored code, I tried to have the same logic as previously implemented:

- Equipment:
  - If item is not researched and is bioStorage: "Alien Organism"
  - If item is not researched and is NOT bioStorage: "Alien Artifact"
  - If item is researched: `itemType.name`
- Vehicle:
  - If item is not researched: "Alien Artifact"
  - If item is researched and item variable has value: `item->type->name`
  - If item is researched and item variable is null: `type->name`

FilmBoy84:
This works nicely

As commented above, any "better" string to describe the function can be done in a future PR as and when found

Similarly with % based representation of space used

Thanks for your work as always 🍻

Issue #1445, opened 2024-05-18
Ammo clip/mag with only 1 bullet will not be added to Ammo mag/clip in weapon that only has a few bullets left


XCOM-HUB:
- OS: Windows 10 Pro x64 22H2
 - OpenApoc Version: [0.0.1062](https://ci.appveyor.com/project/OpenApoc/openapoc/builds/49841322)
 - No Mods

If you have only one bullet left and have another ammo clip/mag with few bullets and try to load those bullets to your weapons mag/clip, then that will just swap out the clip/mag. It should be added to your weapons already existing mag/clip because that is the only logical solution.

PR #1446, opened 2024-05-19
Add partial clip refilling in battlescape

Kurtsley:
Addresses #1445

If the ammo amount in the clip plus the ammo amount in the gun is less than the max ammo amount, the clip will be loaded in the gun instead of swapped as before and the clip will disappear. This appears to be what OG does.


XCOM-HUB:
I can confirm that your reload-fix works great. I like that you added the maximum amount of each clip/mag. It is now much easier to see when your mag/clip is no longer full 🔫👍

Issue #1333, opened 2024-02-05
During base defense mission no equipment or aliens are recovered


muton-commander:
Windows 10

During base defense mission no equipment or aliens are recovered. No Craft present at base. 

One Spitter and one egg in hangar. After mission no corpses or live aliens captured. Removed agent equipment is lost.

Base has store space and Alien containment

[save_Aliens and equipment not recover.zip](https://github.com/OpenApoc/OpenApoc/files/14172621/save_Aliens.and.equipment.not.recover.zip)
[build-id.zip](https://github.com/OpenApoc/OpenApoc/files/14172630/build-id.zip)

[settings.zip](https://github.com/OpenApoc/OpenApoc/files/14172622/settings.zip)

Kurtsley:
Base defense missions break the existing logic for post-battle looting and alien capture. If you have no vehicle in the building where the battle takes place, you can't recover aliens. It does not check if this is a base and has alien containment or storage capacity. The same would happen if you DID have a vehicle but no bio transport module. The game would ignore the live aliens because it believes you can't successfully transport them home. 

There is another issue where if your base has multiple vehicles, they will be loaded with loot at the end of a base defense mission and create multiple "Items have arrived" messages. This needs to be fixed as well. Nothing should be loaded onto vehicles at the end of a base defense mission.

I believe a check in the exitBattle function could fix most of these issues as that is where the list of player vehicles is created to haul the cargo (around line 3157 in battle.cpp).

XCOM-HUB:
@Kurtsley Funny you mentioned this just two days ago. I first ran into this yesterday and wrote about it today #1422

PR #1447, opened 2024-05-19
Live alien recovery after base defense battle

### LINKED ISSUE #1422: Capturing Live Aliens in a Base Attack Mission will come up as 0 in the [DEBRIEFING]
- OS: Windows 11 Pro x64 23H2
 - OpenApoc Version: [0.0.1038](https://ci.appveyor.com/project/OpenApoc/openapoc/builds/49642396)
 - No Mods

When my base was under attack I successfully captured one living Chrysalis, but the [DEBRIEFING] states that I haven't captured any live aliens. When un-pausing in Cityscape I get four messages that I have retrieved items from tactical combat zone and I can clearly see that I have a live Chrysalis in my Alien Containment.

Skärmbild 2024-04-18 201643
Skärmbild 2024-04-18 172021

When completing a mission elsewhere you only get one message about items arriving at your base:

Skärmbild 2024-04-18 171528

SAVE-FILE in Battlescape (Just end the turn and let my Laser Turrets kill one Spitter and one Hyperworm): 
[Base Under Attack Battlescape 3.zip](https://github.com/OpenApoc/OpenApoc/files/15029387/Base.Under.Attack.Battlescape.3.zip)

ayrtondenner:
Fixes #1422

Now, after finishing a Base Defense battle, live aliens will be considered for score calculation, and they will NOT be send to any vehicle, but will be added to base alien storage instead.

Updates list:
- For score math, live aliens will be considered if mission is a base defense and base has alien storage **OR** if mission has player craft with available bio storage.
- Similarly, live aliens will be set as "leftover" only when **BOTH** conditions above are false.
- At `battle.cpp`, instructions inside `if (!playerVehicles.empty())` and `for (auto &v : playerVehicles)` will only be executed when mission is **NOT** a base defense one. Since both instruction groups deal with adding stuff at vehicles to be send to base, it seems that they shouldn't execute when defending a base. Let me know if that's correct or needs to be changed.
- Minor refactors merging nested if conditions that could be represented at same if statement

I didn't add any alien storage limit check when recovering live aliens, since it seems we don't have any limit enforcing right now.

Also, I guess that to fix equipment issue reported in #1333, similar changes must be done but regarding item loot instead.

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -46,6 +46,7 @@
 #include <algorithm>
 #include <glm/glm.hpp>
 #include <limits>
+#include <optional>
 
 namespace OpenApoc
 {
@@ -2684,28 +2685,7 @@ void Battle::finishBattle(GameState &state)
 	// - give him alien remains
 	if (state.current_battle->playerWon && !state.current_battle->winnerHasRetreated)
 	{
-
-		// Check if mission is base defense, and defended base has alien containment facility to
-		// store live aliens from battle
-		auto playerHasBaseAlienStorage = false;
-
-		if (state.current_battle->mission_type == Battle::MissionType::BaseDefense)
-		{
-			for (const auto &base : state.player_bases)
-			{
-				if (base.first != state.current_base.id)
-					continue;
-
-				for (const auto &facility : base.second->facilities)
-				{
-					if (facility->type->capacityType == FacilityType::Capacity::Aliens)
-					{
-						playerHasBaseAlienStorage = true;
-						break;
-					}
-				}
-			}
-		}
+		auto playerHasBaseAlienStorage = getIfPlayerHasBaseAlienStorage(state);
 
 		const auto playerHasCraftBioStorage = state.current_battle->player_craft &&
 		                                      state.current_battle->player_craft->getMaxBio() > 0;
@@ -3131,8 +3111,9 @@ void Battle::exitBattle(GameState &state)
 	// If player has vehicle with bio capacity then all bio loot goes to leftover loot
 	// This is regardless of "enforce limits" which only makes us enforce it
 	// on vehicles that have capacity in the first place
-	bool bioCarrierPresent = false;
-	bool cargoCarrierPresent = false;
+	auto cargoCarrierPresent = false;
+	auto bioCarrierPresent = false;
+	auto playerHasBaseAlienStorage = getIfPlayerHasBaseAlienStorage(state);
 	for (auto &v : playerVehicles)
 	{
 		if (v->getMaxCargo() > 0)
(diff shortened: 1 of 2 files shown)


XCOM-HUB:
I have just tried your latest fix, and it works 🎊🎉👍

ayrtondenner:
Thanks @XCOM-HUB , were you able to check if it didn't mess up any logic for other mission types? I want to be sure that older logic of shipping living aliens via vehicles bio storage is still properly working.

XCOM-HUB:
I just did a UFO crash mission and everything seemed to work. Live Aliens were captured and the score for capturing Live Aliens seemed to be accurate. I couldn't see anything wrong with your fix ( [0.0.1065](https://ci.appveyor.com/project/OpenApoc/openapoc/builds/49844216) )

FilmBoy84:
> I just did a UFO crash mission and everything seemed to work. Live Aliens were captured and the score for capturing Live Aliens seemed to be accurate. I couldn't see anything wrong with your fix ( [0.0.1065](https://ci.appveyor.com/project/OpenApoc/openapoc/builds/49844216) )

Were aliens still added to vehicles before being added to base on return?

XCOM-HUB:
The Live Aliens were retrieved from the mission, and put in Alien Containment because I saw them once my vehicle arrived at my base. I don't know if there is a way to see if there is any Live Aliens inside of my vehicles biological storage directly after the mission. I have to check that.

FilmBoy84:
> The Live Aliens were retrieved from the mission, and put in Alien Containment because I saw them once my vehicle arrived at my base. I don't know if there is a way to see if there is any Live Aliens inside of my vehicles biological storage directly after the mission. I have to check that.

Easiest way to check would be to do a UFO recovery with a bio transport module
On mission end BEFORE vehicle returns to base, check alien containment - captured aliens should not be there yet
When the vehicle returns they should be there

This PR aims to fix Base Defence missions - on completion the aliens should automatically go to the alien containment regardless of vehicles / bio transport modules being present

What we also need to check is that if there is NO alien containment they will be sent to a vehicle with a bio transport module

If no alien containment is present and no vehicle with a bio transport module is present then the aliens should be destroyed with no points for recovery

XCOM-HUB:
The Live Aliens did not go straight to my base after doing a UFO crash site mission. Both dead and alive aliens were added to my Alien Containment in base after the vehicle returned to base. I have double checked it.

FilmBoy84:
All checks passed, LGTM, merging

Thanks as always for working on this 🍻

XCOM-HUB:
Aliens won't be added to the Bio-Transport Module if you don't have any Alien Containment in your bases (regular mission in a building not owned by X-Com).

SAVE-FILE: [At least one Live Alien.zip](https://github.com/user-attachments/files/15805495/At.least.one.Live.Alien.zip)

Issue #1455, opened 2024-05-22
Brainsucker Launcher still being labeled as "Alien Artifact" after having researched it 100%


XCOM-HUB:
- OS: Windows 11 Pro x64 23H2
 - OpenApoc Version: [0.0.1076](https://ci.appveyor.com/project/OpenApoc/openapoc/builds/49863777)
 - No Mods

![Brainsucker Launcher researched but not the pods](https://github.com/OpenApoc/OpenApoc/assets/55395993/8ea68847-2ca2-42ec-bf18-0b23ce94f82f)
![Fires Brainsucker pods](https://github.com/OpenApoc/OpenApoc/assets/55395993/bc67fd80-aa5a-49b0-be43-a73fcf38ae47)
(Maybe the name of the Brainsucker Launcher will be revealed after finishing researching the Brainsucker pods!?)

There are also two Alien craft weapons that has their names already revealed despite no research (Light and Medium Disruptor Beam.)
![Not yet researched](https://github.com/OpenApoc/OpenApoc/assets/55395993/51ee2ca1-eef7-4873-bcc5-a2e33d7796c0)
![Alien craft weapon name showing despite no research](https://github.com/OpenApoc/OpenApoc/assets/55395993/a590a560-789b-4d4b-b4b7-22d451657d1c)

ayrtondenner:
Could you please provide a save file for these cases?

XCOM-HUB:
> Could you please provide a save file for these cases?

Here you go!
[Unresearched Alien Artifacts.zip](https://github.com/OpenApoc/OpenApoc/files/15421891/Unresearched.Alien.Artifacts.zip)

XCOM-HUB:
The name of the Brainsucker Launcher is revealed after you have researched the Brainsucker Pods as well as the Brainsucker Launcher. I think the name should first be revealed after you have only researched the Launcher, and a prompt/pop up window should warn you that you have to research the Brainsucker Pods to be able to fully use it (only when trying to equip a Brainsucker Launcher.)

![After researching Brainsucker Pods](https://github.com/OpenApoc/OpenApoc/assets/55395993/fbe94f2c-98a9-4fd3-9a40-f97d6be0b4ac)

PS: I know that the Brainsucker Launcher is useless since the Brainsuckers will only attack your agents, maybe hostile agents like the Cult of Sirius for example (I haven't tested that yet.)

XCOM-HUB:
[...]

PR #1463, opened 2024-05-25
Warning all weapon ammo types that needs to be researched before using

ayrtondenner:
Following with #1455, game will now inform user all the ammo types that needs to be researched together with weapon before you can be able to use it.

Also updated `tryPickUpItem` to not only return pickup result but also which item was selected.

Example where Brainsucker Launcher is researched, but Brainsucker Pod is not:
image

Let's say that a weapon has multiple ammo types. Warn message will properly display list of ammo types that need to be researched:
image

Items that are not weapon, or weapon with no ammo type will display default message. Example with Disruptor Gun:
image

### DIFF
--- a/game/ui/general/aequipscreen.cpp
+++ b/game/ui/general/aequipscreen.cpp
@@ -32,6 +32,7 @@
 #include "game/ui/general/aequipmentsheet.h"
 #include "game/ui/general/agentsheet.h"
 #include "game/ui/general/messagebox.h"
+#include <boost/algorithm/string/join.hpp>
 
 namespace OpenApoc
 {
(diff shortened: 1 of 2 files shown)


FilmBoy84:
Thanks for this 🍻 

Reading though - LGTM

I'm compiling now and will get on testing tomorrow

XCOM-HUB:
The name of the Brainsucker Launcher should be revealed after you have only researched the Brainsucker Launcher, but the prompt that says that you need to also reaserch the Brainsucker Pods should still be in. I also think you should be able to sell the Brainsucker Launcher after only having researched the Brainsucker Launcher.

ayrtondenner:
> The name of the Brainsucker Launcher should be revealed after you have only researched the Brainsucker Launcher, but the prompt that says that you need to also research the Brainsucker Pods should still be in.

Addressed in https://github.com/OpenApoc/OpenApoc/pull/1472

ayrtondenner:
> I also think you should be able to sell the Brainsucker Launcher after only having researched the Brainsucker Launcher.

@XCOM-HUB I just tested it and I was able to sell Brainsucker Launcher after having it researched but not Brainsucker. Can you test it again with current version?

image
image

XCOM-HUB:
Yes! I can confirm that you can sell the Brainsucker Launchers now without also having to wait for the Brainsucker Pods to be fully researched. Thank you for your work and dedication to the OpenApoc project 👍 @ayrtondenner

PR #1468, opened 2024-06-07
Adding currency formatting to all label texts starting with dollar sign char

ayrtondenner:
Following #1467, now adding credit label formatting to all text fields where its content starts with '$' char

image
image
image
image
image
image
image


FilmBoy84:
Tested and LGTM, merging 🍻

PR #1472, opened 2024-06-18
Showing weapon name when weapon is researched but not ammo type

ayrtondenner:
When a weapon is researched, but not its ammo type, weapon name will be displayed, but game will comply about need to research ammo before using it.

image

If weapon is not researched, it's still displayed as Alien Artifact

image

Same for Buy and Sell screen:

image
image

Research topic will be searched by name, to check if it matches with weapon name. If no topic match is found, then code will use previously `item->type->canBeUsed` function - which is not the best case, since this will consider not only weapon research status but also ammo research status.


FilmBoy84:
Thanks for this, tested and appears to be working as intended
Merging 🍻

PR #1473, opened 2024-06-22
Renaming property to "Max Samples" and adding "Storage" property at ufopaedia

### LINKED ISSUE #1470: Rename "Aliens Held" to "Alien Storage Capacity" for clarity
I think it is misleading to have it say "Aliens Held" because that indicates that there are 20 Aliens currently held inside the Bio-Transport Module.

PS: If "Alien Storage Capacity" is too long, replace it with "Capacity" or "Alien Capacity".
Aliens Held

ayrtondenner:
Fixes #1470

1. Renamed label "Aliens Held" to "Max Samples" in Bio-Transport Module.

image

2. Added label "Storage" at ufopaedia for agent and vehicle equipments

image
image
image

Obs.: besides "Max Samples", the label "Max Lifeforms" was also suggested, but its length is greater than what view can properly display:

image

### DIFF
--- a/data/languages/ufo_string.pot
+++ b/data/languages/ufo_string.pot
@@ -4448,7 +4448,7 @@ msgstr ""
 msgid "Cargo"
 msgstr ""
 
-msgid "Max Lifeforms"
+msgid "Max Samples"
 msgstr ""
 
 msgid "Jamming"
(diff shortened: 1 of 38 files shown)


FilmBoy84:
Max Samples makes more sense anyway as future mods may use biological objects that are not aliens

Tested and LGTM, merging 🍻

PR #1479, opened 2024-06-30
Changing log type at buildFacility and destroyFacility

ayrtondenner:
As described [here](https://www.reddit.com/r/OpenApoc/comments/1drmirh/game_crash_on_starting_a_new_campaign/), the game will throw a bunch of errors when starting a new campaign. That's because both `buildFacility` and `destroyFacility` are called during new campaign creation at a time where their validation fails - which seems to be the normal process during new campaign.

Changing log type from `LogError` to `LogWarning` so we can avoid blocking when starting a new campaign, while still warning that these checks didn't work when needed to follow these errors.

### DIFF
--- a/game/state/city/base.cpp
+++ b/game/state/city/base.cpp
@@ -284,7 +284,7 @@ void Base::buildFacility(GameState &state, StateRef<FacilityType> type, Vec2<int
 {
 	if (canBuildFacility(type, pos, free) != BuildError::NoError)
 	{
-		LogError("Error when trying to build facility!");
+		LogWarning("Error when trying to build facility!");
 		return;
 	}
 
@@ -407,7 +407,7 @@ void Base::destroyFacility(GameState &state, Vec2<int> pos)
 {
 	if (canDestroyFacility(state, pos) != BuildError::NoError)
 	{
-		LogError("Error when trying to destroy facility!");
+		LogWarning("Error when trying to destroy facility!");
 		return;
 	}
 


XCOM-HUB:
Wouldn't it be nice to have the option to start a new Campaign where you were able to choose which base you would like to begin with and build the basic facilities in that base just the way you want instead of having to use the F10 command and the cheat option to give you extra money to be able to afford to choose your base layout?

XCOM-HUB:
You could add additional buttons beside each difficulty so you could choose to start a standard game or an optional game mode where you get to choose any base you want and then add the basic free facilities like 1 Alien Containment, 2 Stores, 3 Living Quarters, 1 Workshop, 1 Biochemistry Lab, 1 Quantum Physics Lab, 1 Training Area, 1 Psi-gym and 1 Medical Bay. These free facilities would be completed instantly when you place them where you want them. Maybe even give the option to choose where to put the Access Lift!?

FilmBoy84:
> Wouldn't it be nice to have the option to start a new Campaign where you were able to choose which base you would like to begin with and build the basic facilities in that base just the way you want

This has been discussed MANY times before in OA development

The main issue (other than UI and coding a solution) is that is VASTLY changes the balance at each of the five difficulty levels as both the starting buildings for each map and the starting money assume that you will have to lose a few days of game time and cash adjusting the basement to your liking

If you start being able to customise your *inherited* basement from the start it changes that dynamic drastically and we'd have to alter UFO spawn, starting cash, tech level progression and a host of other things to balance the boost to the player it offers

However, i do like the idea of being able to specify which starting base you use in each city - though again this would need to incur a financial penalty to starting money to stop everyone picking the "optimal" base instead of leaving it to random chance

XCOM-HUB:
> However, i do like the idea of being able to specify which starting base you use in each city - though again this would need to incur a financial penalty to starting money to stop everyone picking the "optimal" base instead of leaving it to random chance

Just subtract the cost of the base you choose from your starting capital/funds. Also, add the right amount of money before you choose your base. You should have a little more money from the beginning when you start with no base compared to a standard game where the game chooses a base for you. I realize this would be a difficult thing to implement. A lot of things need to be in place for this to work. I do hope it will be implemented eventually.

PR #1483, opened 2024-07-01
Stop adding empty cargo at vehicle when exiting battle

### LINKED ISSUE #1482: Vehicles carrying an empty Alien Artifact after leaving a mission
- OS: Windows 10
- OpenApoc Version: [0.0.1117](https://ci.appveyor.com/project/OpenApoc/openapoc/builds/50118887)
- No Mods
- Savefiles:
  - Battle: [Just kill the Anthropod.zip](https://github.com/user-attachments/files/16044988/Just.kill.the.Anthropod.zip)
  - After battle: [Alien Containment in Base 2.zip](https://github.com/user-attachments/files/15806919/Alien.Containment.in.Base.2.zip)

When a vehicle leaves a battle, it carries an empty Alien Artifact item with count = 0.

image
image

ayrtondenner:
Fixes #1482

Vehicles leaving a mission used to have an empty "alien artifact" item, with count zeroed.
This item is not added anymore, which means that vehicles with no loot are now empty for good.

As a consequence, the dialog "Items from tactical combat zone have arrived" now will only appear when landed vehicle really have something to unload at base, since it used to appear for every single vehicle landing.

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -3361,12 +3361,9 @@ void Battle::exitBattle(GameState &state)
 		}
 	}
 
-	// Give player vehicle a null cargo just so it comes back to base once
+	// Sending vehicles back to base
 	for (auto &v : playerVehicles)
 	{
-		v->cargo.emplace_front(
-		    state, StateRef<AEquipmentType>(&state, state.agent_equipment.begin()->first), 0, 0,
-		    nullptr, v->homeBuilding);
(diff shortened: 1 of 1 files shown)


XCOM-HUB:
It works 🥳 My Phoenix Hovercar 2 no longer gets a message about "Items from tactical combat zone have arrived". Hurray 🥳

Killing all the X-COM bugs one at a time 👍

PR #1490, opened 2024-08-11
Fixing logic at Base::alienContainmentExists

ayrtondenner:
I realized that the logic at `Base::alienContainmentExists` is wrong and the method name is misleading right now.
Fixing logic so alien containment is True when base has alien capacity.


FilmBoy84:
Good spot and thanks for the fix
Tested and Merging
🍻

Issue #1310, opened 2023-12-30
Ship stuck in slow-motion


BeornTB:
- OS: Windows 10 x64 2020H2
 - OpenApoc Version 20231225-3
 - No mods

During combat in the alien dimension, a retaliator was repeatedly stasis bombed and at one point, it came back but remained very slow. Stays that way upon reloading.

[save_G4 Retaliator Slo-Mo.zip](https://github.com/OpenApoc/OpenApoc/files/13799000/save_G4.Retaliator.Slo-Mo.zip)

[Log Settings.zip](https://github.com/OpenApoc/OpenApoc/files/13799001/Log.Settings.zip)

FilmBoy84:
This does indeed appear to be a slow craft

* Loaded save
* Deselected group of vehicles - selecting only retaliator
* Allowed time to pass - slow craft
* Selected manual control and instructed retaliator to move to a new location, allowed time to pass - slow
* Checked to see the top speed of craft, it's reporting correctly with no missing modules

![image](https://github.com/OpenApoc/OpenApoc/assets/33913088/7757f796-3d88-4f18-b2f7-b1aab8679362)
![image](https://github.com/OpenApoc/OpenApoc/assets/33913088/34590373-2b30-4a39-a624-9a7b02111cdb)

Kurtsley:
Been looking at this, this isn't a stasis bomb issue it is the last example of a vehicle getting "stuck" in the cityscape. While the other version of "stuck" was fixed, ships can get stuck moving very slowly across the map forever.

Fortunately, it seems to be fixable. I added a debug command that sets the vehicle's current position to it's goal position. When it warps to the spot it wants to be, the vehicle snaps out of it and acts normally. Why exactly this works is something I'm still working on, but there are some clues. 

First, from what I see this only happens on the slowest speed. This explains why I never had ships get stuck like this in my testing. I simply never user the slowest speed when testing aerial battles. When I noticed this issue in a couple of Filmboy's videos, it was when the speed was set to slowest.

Second, it looks like it happens in the middle of a turn. I have a hunch that it happens when a vehicle selects a new target when it's current target is destroyed. I say this because I didn't see it happen when damage was disabled.

I'm going to continue testing this, but hopefully it should be fixed soon and the stuck vehicle issue can be eliminated forever.

FilmBoy84:
PARTIAL FIX via debug command in #1489 leaving open until fully resolved

FilmBoy84:
[...]

PR #1492, opened 2024-08-15
Fix minimum velocity when turning

Kurtsley:
Addresses #1310 

I found what is causing the slow vehicle problem, but this is just a band-aid approach to fixing it.

On line 573, if ticksToMove is 1 and ticksToTurn is high (100+), the vehicle velocity will be set to a ridiculously small number. If this happens, the vehicle can be stuck moving too slowly to reach its' goal, especially if the vehicle is facing the correct direction. Since the vehicle is facing its' goal, the velocity never changes as it is set when a new goal is selected. This PR sets the minimum ticksToMove to 2, as this seems to be enough to stop the problem. Since this issue only happened on the slowest speed where ticks are set to 1, increasing the number to 2 should be enough.

Rewriting the turning speed mechanic is above my paygrade, so this should stop this problem until the day someone tackles that section.

https://github.com/OpenApoc/OpenApoc/blob/5f18f2759f0770e8bea46c9fb8752a280d31acfe/game/state/city/vehicle.cpp#L560-L575

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -567,7 +567,7 @@ class FlyingVehicleMover : public VehicleMover
 						    std::max(floorf(glm::length(vectorToGoal) /
 						                    glm::length(vehicle.velocity) * (float)TICK_SCALE) -
 						                 5.0f,
-						             1.0f);
+						             2.0f);
 						if (ticksToMove < vehicle.ticksToTurn)
 						{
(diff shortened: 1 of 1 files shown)


FilmBoy84:
Thankyou as always for the fix
Agreed, looks like there will need to be a rework of the code at a later point
But providing this stopgap solution is greatly appreciated
Tried a couple of saves with the issue and this appears to resolve it

Merging

kgd192:
Should finally have some spare time again this weekend and will have a look into the code for turning and how to improve it 🙂

Issue #1411, opened 2024-04-14
The Hoverbike and Blazer Turbo Bike have lower Top Speeds than in OG


XCOM-HUB:
- OS: Windows 11 Pro x64 23H2
 - OpenApoc Version: [0.0.1025](https://ci.appveyor.com/project/OpenApoc/openapoc/builds/49612209)
 - No Mods

Hoverbike Top Speed in OG: 14 VS. 10 in OpenApoc
![Hoverbike OG](https://github.com/OpenApoc/OpenApoc/assets/55395993/3f0d9392-4924-4f80-82f4-754d414c74a4)
![Hoverbike OpenApoc](https://github.com/OpenApoc/OpenApoc/assets/55395993/fd185ac5-1d6d-409e-b5bc-7a9e5f40b5ba)

Blazer Turbo Bike Top Speed in OG: 7 VS. 3 in OpenApoc
![Skärmbild 2024-04-15 000221](https://github.com/OpenApoc/OpenApoc/assets/55395993/215e1739-3227-4eef-9be1-2e3547d36a9b)
![Skärmbild 2024-04-15 004914](https://github.com/OpenApoc/OpenApoc/assets/55395993/7386aa9f-0aae-48a0-95f6-60db7c52e974)

Edit: I can now confirm that the speeds of a Hoverbike and a Blazer Turbo Bike differ if compared between OG and OpenApoc. I made the Hoverbike and Blazer Turbo Bike race from the far left side of the cityscape map to the far right side of the cityscape map using straight roads in both OG and OpenApoc. The distances I used in both OG and OpenApoc were of identical length.

The Hoverbike took 33 seconds in OG while in OpenApoc the same distance took 1 minute and 10 seconds.

The Blazer Turbo Bike took 1 minute and 7 seconds in OG while in OpenApoc the same distance took 3 minutes and 29 seconds.

![Skärmbild 2024-04-15 025025](https://github.com/OpenApoc/OpenApoc/assets/55395993/67193c44-7ee0-4a94-84be-01318f275b42)

FilmBoy84:
We probably need to check this is a code issue or a #997 issue

Depends how top speed is calculated - as there's the specified power of the engine, then a host of weight calcs depending on what equipment and ammo is loaded IIRC

Will make a note to do some further investigation

Thanks for report

Kurtsley:
Have you looked at their speed in a side by side  comparison with OG?

XCOM-HUB:
[...]

PR #1495, opened 2024-08-30
Fix #1310 Turning Slow Down Mechanic

### LINKED ISSUE #1310: Ship stuck in slow-motion

kgd192:
This PR should solve Issue #1130 by decoupling the slow down during turns mechanic from the game ticks. It now uses the vehicles angular velocity. This unfortunately lead to another correction value, which is purely based on data analysis. Average deviation of the Slowdown compared to the old method is about 2.5%.
@FilmBoy84 I will do the fix for Lint this weekend, I tried to work with just the command line and without IDE out of curiosity, this somehow disregarded the formatting.

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -36,6 +36,8 @@
 #include <limits>
 #include <queue>
 #include <random>
+#include <fstream>
+#include <iostream>
 
 namespace OpenApoc
 {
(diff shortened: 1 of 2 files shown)


FilmBoy84:
Very nice

Did some playing around with manual control (select craft and N key, right click to select destination) and varying altitudes as well as testing on Road Vehicles too, appears to work as intended

There is a typo on the comment for `vehicle.h` line 61 where "conclusion" is put as `conclution` that should probably be amended before merge

Thanks as always for your work on this 🍻

kgd192:
> There is a typo on the comment for `vehicle.h` line 61 where "conclusion" is put as `conclution` that should probably be amended before merge

Good catch, i will amend that tonight.

Otherwise as already mentioned I am pretty sure this will probably be visited again once the ticks vs fps issue has been corrected/streamlined and maybe within the solution of #1411 as well.

Issue #1499, opened 2024-10-13
Crash when buying gear or personnel


Kothyxaan:
I have a crash whenever I order personnel or equipment.
Crashes at 

I have tried versions (both crashed):
OpenApoc-20240829
OpenApoc-20240903

Crashes about 15 seconds (game time) after bought.

Version:
openapoc-20240815
Is working fine for buying equipment and hiring personnel, but it crashed after the first battle (notified of victory, onto splash screen, crash).

Kothyxaan:
Just in case it matters. 
I downloaded the windows install options for each of those builds, ran the install.

FilmBoy84:
Hi Kothyxaan

Sorry for delayed response - I shall try and get on replicating this Thursday to see if you can work through to a solution

Is the error on a new, clean, game or a save?

Kothyxaan:
Hey, thanks for the reply. 
I tried it on a new install with a new save.

FilmBoy84:
Thanks for further info, will help with replication 🍻

Kothyxaan:
Hopefully it is a simple fix and not just a problem I have. Good luck either way. Also thank you for the work you are doing, X-Com Apoc was/is one of my favourite games that I return to every few years, so cheers!

FilmBoy84:
OK managed to replicate on `20240903`

I start a new game (Easy in this case)
I buy some High Explosives
![image](https://github.com/user-attachments/assets/0e546628-f1b5-48c8-a8f5-7e3d01b4e00a)
![image](https://github.com/user-attachments/assets/50171dc2-44ad-4ba9-bdf0-0a1506b49c83)

I locate Airtrans 1 (which is doing the transport mission) on the map
![image](https://github.com/user-attachments/assets/3b1f8f34-77b7-479a-84b0-370431abd172)
![image](https://github.com/user-attachments/assets/0b5fb9bd-45a5-4fa8-99b9-4d91ca277d92)

Now, there are TWO ways to crash this

The first is as the airtrans enters the circled warehouse for pick-up - the game will crash on entry if the vehicle is selected and followed
![image](https://github.com/user-attachments/assets/7bbed6ed-664a-4762-99d4-c0e8740267c6)

Alternatively, DO NOT follow the craft and let it enter and exit the warehouse, the game will then crash on entering the X-COM base (when the usual items delivered message would pop up)

Log, game settings and save attached
[...]

PR #1501, opened 2024-11-03
Fix facilities check at non-base

Kurtsley:
Addresses #1499 

I feel that everything happening regarding aliens at the beginning of getServiceDestination should only be checked at bases. Obviously non-bases won't have alien containment. This should fix this specific issue of checking non-bases for facilities.


FilmBoy84:
Thanks as always Kurtsley 🍻 

Will get appveyor sorted and on to testing for merge

FilmBoy84:
Jonny has kindly resolved the Appveyor issue

Have done the following to test this build

Started a new game
Purchased a host of agent equipment
Selected an Airtrans as soon as it spawned and followed it to and from the factory to the X-COM base
Ran a couple of missions with and without bio transport and live aliens
No crashes

LGTM

Thanks for this fix and apologies for the delay getting it to merge

Issue #940, opened 2020-09-09
No Vehicle object matching ID "VEHICLE_19"


Lukewhale:
Hi, I have been speaking to FilmBoy and he has ask me to log this and send save and log

[save_lukewhale.zip](https://github.com/OpenApoc/OpenApoc/files/5198244/save_lukewhale.zip)

[log.txt](https://github.com/OpenApoc/OpenApoc/files/5198253/log.txt)

All i done in the game was buy/sell and build base.
(Note) Sold ground units and bought hover bikes x4 , hover cars x4 and equip them with lasers and better engines. Change the name of soldiers and recruit new researchers/engineers. Dont this is this may have caused a issues but i moved all the researchers and engineers out of the base and then picked the highest skill level

Anything else just let me know.. Thank you   

OpenApoc error

class std::shared_ptr<class OpenApoc::vehicle> _cdecl OpenApoc::stateObject<class OpenApoc::Vehicle>::get(const class OpenApoc::GameState &,const class std::basic_string<char_traits<char>,class std::allocator<char> > &): No vehicle matching ID "VEHICLE_19"

void _cdecl OpenApoc::StateRef<class OpenApoc::vehicle>::resolve(void) const: No Vehicle object matching ID "VEHICLE_19" found

FilmBoy84:
This is part of #931 So please see there too :)

FilmBoy84:
Closing as referenced and a duplicate

makus82:
Thx for report anyway L =]

Issue #931, opened 2020-08-04
[Cityscape] No Vehicle object matching ID "VEHICLE_19"


BeornTB:
![image](https://user-images.githubusercontent.com/46300745/89300417-2c6adb80-d636-11ea-81c9-1b04d82c9111.png)

Anytime I load a save in the cityscape, vehicle_19 crashes the party - so to speak.

I mentioned it in another thread but it keeps re-occuring every time so here's a fresh one. New save/log on the latest build.

[VehicleID-CTD.zip](https://github.com/OpenApoc/OpenApoc/files/5022461/VehicleID-CTD.zip)

Atrosha:
gdb backtrace

```
Thread 1 "OpenApoc" received signal SIGSEGV, Segmentation fault.
OpenApoc::GameState::updateEndOfSecond (this=0x18250e0) at /home/atrosha/OpenApoc/game/state/gamestate.cpp:1063
1063				for (auto &e : v->equipment)
(gdb) bt
#0  OpenApoc::GameState::updateEndOfSecond (this=0x18250e0)
    at /home/atrosha/OpenApoc/game/state/gamestate.cpp:1063
#1  0x000000000080ebf8 in OpenApoc::GameState::update (this=0x18250e0, ticks=6)
    at /home/atrosha/OpenApoc/game/state/gamestate.cpp:1029
#2  0x000000000080ec5f in OpenApoc::GameState::update (this=<optimized out>, 
    ticks=ticks@entry=6) at /home/atrosha/OpenApoc/game/state/gamestate.cpp:987
#3  0x00000000005ac31f in OpenApoc::CityView::update (this=0x281b300)
    at /usr/include/c++/10/bits/shared_ptr_base.h:1324
#4  0x000000000049655e in OpenApoc::Framework::run (this=this@entry=0xca2270, 
    initialStage=std::shared_ptr<OpenApoc::Stage> (use count 1, weak count 1) = {...}) at /usr/include/c++/10/bits/shared_ptr_base.h:1324
#5  0x0000000000469f03 in main (argc=-9440, argv=0x7fffffffde58)
    at /usr/include/c++/10/bits/shared_ptr_base.h:759
```

BeornTB:
Here's one at the ending of a tactical mission, same deal.

[CTD VehicleID Tact.zip](https://github.com/OpenApoc/OpenApoc/files/5023600/CTD.VehicleID.Tact.zip)

Atrosha:
[...]

PR #1507, opened 2024-11-20
FIX - [Cityscape] No Vehicle object matching ID "VEHICLE_19" #931#940

NagyAlex96:
"No Vehicle object matching ID error" issue resolved.

Tested with these save files. This error was already resolved (I couldn’t reproduce it), but it didn’t work properly with old saves. It should work correctly now.

#931
#940

Facts (tested):
Old saves might contain an agent_id error.
When there are more than 12 passengers, the number is not displayed correctly.

My commit includes my previous fix because it wasn't merged into the main branch."

### DIFF
--- a/game/state/gamestate.cpp
+++ b/game/state/gamestate.cpp
@@ -1101,6 +1101,16 @@ void GameState::update(unsigned int ticks)
 		{
 			if (v.second->city == current_city)
 			{
+				auto vehicleMission = v.second->missions;
+				if (!vehicleMission.empty() &&
+				    vehicleMission.back().type == VehicleMission::MissionType::AttackVehicle &&
+				    (vehicleMission.back().targetVehicle == nullptr ||
+				     vehicleMission.back().targetVehicle->city != current_city))
+				{
+					v.second->clearMissions(*this);
+					v.second->addMission(*this, VehicleMission::gotoBuilding(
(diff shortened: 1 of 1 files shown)


FilmBoy84:
> Facts (tested): Old saves might contain an agent_id error. When there are more than 12 passengers, the number is not displayed correctly.

I cannot seem to replicate a problem with more than 12 passengers not displaying properly? In both the Original Game and OpenApoc anything above 12 is replaced with a "+" symbol (although i would argue a better solution should exist to two-digit number display, although we are limited by a 640x480 legacy UI)

Can you please explain what i need to be looking for?

> My commit includes my previous fix because it wasn't merged into the main branch."

I've just merged the other fix :)
Going forward, we usually keep each change in separate commits to a pull
And each fix in a separate PR

But no worries, thanks for heads up on this

NagyAlex96:
"Really? I've never had more than 12 passengers, so I never noticed that for more than 12 passengers, the number is replaced with a '+' symbol.
I think it would be better if it showed the number instead of the '+' symbol.

Could you contact me on Discord?

FilmBoy84:
No probs, official OpenApoc server is here
(Most dev conversation not relevent to PRs there anyways)

https://discord.gg/WwDaFS9

Issue #1392, opened 2024-04-02
Base selection doesn't update properly when selecting another base via tab view


ayrtondenner:
**Desktop (please complete the following information):**
 - OS: Windows 10 x64
 - OpenApoc Version 0.0.994
 - Mods: None

**Describe the bug**
If you enter Base view, enter another view via tabs where you can change between bases (eg Alien Containment), then you select another base and come back to Base view, the base name and map location doesn't work properly. Selecting base again doesn't update.

**To Reproduce**
1. Go to Bases view with Base 1 selected
2. Select a tab view that allows you to change base selection, eg Alien Containment
3. Select Base 2
4. Click OK
5. Base selection will show Base 2, but with name and location at map (red square) for Base 1
6. Click to select Base 2
7. No change will happen
8. Only if you click in Base 1 and then click in Base 2 that all information will be updated properly

**Expected behavior**
After leaving tab view, Base 2 should be properly selected at base selection, base name and base map location.

**Savegame**
[save_Stream Test.zip](https://github.com/OpenApoc/OpenApoc/files/14841282/save_Stream.Test.zip)

**Video**
https://youtu.be/VeqxuS47yRc

Kurtsley:
In the OG, if you select a different base from a different view like Alien Containment or Transactions screen, the original base will remain selected when you go back to the base screen. So if you have base 1 selected, you will always be returned to base 1 when exiting any other screen. I actually think this is the correct way to do it since what you do in another view doesn't necessarily mean you want to switch bases in all views. If I'm buying things for base 2, I may not actually want to switch away from my main base when exiting the transaction screen. I know this isn't exactly what this issue is about, but I've noticed that this issue has come up before. Now, the incorrect base will be shown when exiting another screen and nothing is highlighted. Should the original functionality be kept?

PR #1508, opened 2024-11-24
Fix BaseScreen not highlighting correct base

### LINKED ISSUE #1298: Buy Sell With Multiple Bases Bug
When you have multiple bases, the initially selected base before you go into the Buy Sell window appears to be the only base that has its inventory displayed. When switching to the other base, it still shows the initial base's inventory. A second bug is combined with this. When you exit back to the Bases screen, the bases in the top left corner have nothing selected, and the base layout and base name do not match. Please see https://www.twitch.tv/videos/2014685651 for an in-depth test. This is with build 20231214B along with EWM mod music mod.
[Buy Sell Bug With Multiple Bases.zip](https://github.com/OpenApoc/OpenApoc/files/13774457/Buy.Sell.Bug.With.Multiple.Bases.zip)

Kurtsley:
Fixes #1298 and #1392

This restores the functionality of the original game. In the transfer, buy and sell, and alien containment screens, you will be returned to the original base that was selected from the base screen. In the research screen, you will be transferred to the base you select regardless of the original selection. Not sure if this is desired but this does fix bases not being highlighted as well. If it would be better to unify the behavior for all screens, it would be an easy fix.

### DIFF
--- a/game/ui/base/basescreen.cpp
+++ b/game/ui/base/basescreen.cpp
@@ -96,16 +96,12 @@ void BaseScreen::begin()
 	                  [](Event *) { fw().stageQueueCommand({StageCmd::Command::POP}); });
 	form->findControlTyped<GraphicButton>("BUTTON_BASE_BUYSELL")
 	    ->addCallback(
-	        FormEventType::ButtonClick,
-	        [this](Event *) {
-		        fw().stageQueueCommand({StageCmd::Command::PUSH, mksp<BuyAndSellScreen>(state)});
-	        });
+	        FormEventType::ButtonClick, [this](Event *)
+	        { fw().stageQueueCommand({StageCmd::Command::PUSH, mksp<BuyAndSellScreen>(state)}); });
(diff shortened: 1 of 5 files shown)


Kurtsley:
Strange, I formatted this in Visual Studio and it still failed.

FilmBoy84:
> Strange, I formatted this in Visual Studio and it still failed.

Are you on LLVM/Clang 14?

Kurtsley:
Not sure, I've never bothered with that before. I'll try in Linux.

Kurtsley:
I just ran clang-format from WSL and copied the file over. Same exact results.

FilmBoy84:
Thanks for this fix Kurtsley, tested and merging 🍻 

I'll have a think about unification of functionality - for now it matches OG which is the priority

What was the resolution to the Linting issue btw?

Kurtsley:
I just copied the entire file from github and made my change without formatting the file on my end.

PR #1511, opened 2024-11-26
Update installer to correct version information

Kurtsley:
The version number has been broken for a while in the Windows installer, we are using a different format now. The installer now uses the commit date, number of commits, and hash. I can't properly test this, but if it fails it will still fall back to 0.0.0

Now:
installerpic

Fix:
installpic2

### DIFF
--- a/install/windows/installer.nsi
+++ b/install/windows/installer.nsi
@@ -28,20 +28,20 @@
 ;--------------------------------
 ;General
 
-	;Get major and minor version numbers from the git tag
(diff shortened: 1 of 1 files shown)


FilmBoy84:
Thanks for these launcher fixes

Regarding Copyright Year, we should probably have the latter year set by current system year so it's always current given the OSS nature of the project ensuring longevity

But that's stuff to consider when we need to change it again 🤣

PR #1512, opened 2024-11-29
Fix UFOpaedia form crash

Kurtsley:
A value was added to the UFOpaedia screen without changing the form, causing a crash in certain situations. This adds an extra label and value for this situation.


FilmBoy84:
Thanks as always for the fix
Merging

PR #1514, opened 2024-12-02
Add moving portals

Kurtsley:
This moves portals at the end of every week by a semi-random amount and makes sure it isn't within 4 blocks of another object. The initial portal creation also checks in the a 4 block radius. Generate portals was being called daily as well as every invasion, these have been removed.

The updateEndOfWeek function was changed as it is called at the start of a game and was throwing an error with the portal creation.

This should be tested thoroughly, I had one crash on exit that may or may not have been related to these changes.

### DIFF
--- a/game/state/city/city.cpp
+++ b/game/state/city/city.cpp
@@ -323,7 +323,35 @@ void City::dailyLoop(GameState &state)
 			b.second->updateWorkforce();
 		}
 	}
-	generatePortals(state);
+}
+
+void City::weeklyLoop(GameState &state) { generatePortals(state); }
+
+bool City::canPlacePortal(Vec3<float> position)
+{
+	if (!map->tileIsValid(position) || !map->getTile(position)->ownedObjects.empty())
+	{
+		return false;
+	}
+
+	for (int i = 1; i < 5; i++)
+	{
+		for (int x = -1; x <= 1; x++)
+		{
+			for (int y = -1; y <= 1; y++)
+			{
+				for (int z = -1; z <= 1; z++)
+				{
+					const auto newTile = position + Vec3<float>(x, y, z) * static_cast<float>(i);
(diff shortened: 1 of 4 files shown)


FilmBoy84:
Much appreciated, I've put out a ping to testers on the discord to try this PR's build

Regarding the crash, were you able to get a trace on it at all? Or an idea if it was an End of Week or End of Day related calc?

Kurtsley:
It happened before I switched the portal generation from end of day to end of week. It happened it the City destructor and looked like it was having trouble removing a tile.

Kurtsley:
This next commit changes the z axis check to 2. It also puts a simple weighting system in place to distribute the portals, these values can be tweaked. The x and y axis's are using a normal distribution where the center of the map is most likely to be chosen. I find it works to push the portals away from the edge of the map, especially on superhuman. The z axis has been changed from 2 -8 to 4 - 8 and is weighted towards the high end. If a new position can't be found when trying to move a portal, it will simply stay in place instead of disappearing.

Kurtsley:
I think this is ready to be added. It isn't perfect, but it can be tweaked later.

I've added a handy debug option "P" to move the portals at any time. Even though I tried to weight them towards the middle of the map, they always seem to drift to the edges and stay there. Since they are looking for an empty space to occupy, they always take the path of least resistance once they reach the map edge and rarely go back into the city.

On a side note, I think a separate issue should be raised to link the portals in the alien dimension to specific portals in the city. I didn't quite get around to that in this PR.

FilmBoy84:
Thanks for your work on this
Just finished testing and no noticeable crashes

I agree though, the fact that portals drift away from the city, despite being weighted, as the game progresses is a little frustrating - there must be something else to the calculation we are missing

Nonetheless, it's a massive improvement on the previously static portals

Merging 🍻

Issue #1109, opened 2022-10-28
setPosition called on vehicle with no tile object crash


FilmBoy84:
**Describe the bug**
A Valkyrie, carrying agents to a partially collapse building in the cityscape causes the following error
![image](https://user-images.githubusercontent.com/33913088/198640405-79e8ae20-1deb-4c66-be41-29fd464e8cd4.png)
The game can be progressed by "holding" the ENTER key to confirm each prompt that appears, each time edging forward the game a few ticks until the craft is at the building and can start the mission

**To Reproduce**
1. Load the attached save game
[save_20221027 Save 7 Vehicle Bug 1.zip](https://github.com/OpenApoc/OpenApoc/files/9889184/save_20221027.Save.7.Vehicle.Bug.1.zip)
2. Allow time to continue
3. The above error prompt will begin to display
4. Holding the ENTER key to confirm the prompt will allow time to move forward a few ticks until the craft arrives at it's destination and land in the hangar (which is still present)

**Expected behavior**
If a building is damaged or partially collapse, as long as the hangar or garage is not destroyed, craft should be able to approach the building and land/park as normal. If the garage/hangar is destroyed - an appropriate prompt should be presented to the player informing them that the building is destroyed/damaged and travel there be refused by the system until the hangar/garage is repaired.

In the original game, partially destroyed buildings also affected battlescape map choice and generation, but the algorithm for this is not implemented and already assigned to another issue, but it should be taken into consideration when working on a solution that deals with partially collapse buildings and hangar/garage status at that building

Video of the bug can be found here on a recent stream at ~1:52:20 onward
https://youtu.be/t47gqvHg2m4?t=6743

FilmBoy84:
Mod Files for the above savegame

[20221027 mods.zip](https://github.com/OpenApoc/OpenApoc/files/9889253/20221027.mods.zip)

Andy51:
This one was not caused by a Valkyrie but by a Rescue transport that grabbed a Construction vehicle parked in a building. Could not figure out how it ended up in a crashed and parked state at the same time (could be something to do with zero fuel in its engine), but this should have been fixed with #1137 as rescue of the parked vehicles is no longer possible.

FilmBoy84:
Should be resolved by #1515 
Thanks @NagyAlex96 🍻

PR #1515, opened 2024-12-06
FIX - setPosition called on vehicle with no tile object crash #1109

NagyAlex96:
The issue turned out to be more complex than it initially seemed. Rescue V. 77 was trying to rescue Construction V. 76, which had crashed and was located inside a building. Since it didn't have a 'tileObject,' fixing that introduced new issues. These were also resolved. However, further testing is needed to ensure that Rescue V. still functions properly.


FilmBoy84:
Thanks for this :beers:

I'm still testing to try and find instances of crash recovery not working. If noone else finds anything with the build here https://ci.appveyor.com/project/OpenApoc/openapoc/builds/51123518/artifacts by Christmas i shall merge

PR #1527, opened 2025-02-28
Add a mouse capture option

Kurtsley:
I saw that OpenXcom had this and thought it would be a good addition.


JonnyH (approved):
Looks good.

Would we want this changeable in game too? But that can be done later if needed.

And the UI xml has a lot of churn, but I guess that's just what the qt editor does.

Kurtsley:
Someday, when the options menu is improved, I hope all the launcher options can be added.

FilmBoy84:
LGTM and functions as intended - merging

Issue #1211, opened 2023-04-26
Organisation Infiltration Graph not functioning as per Original Game


FilmBoy84:
**Describe the bug**
Currently this is only referred to in the roadmap #263

In OpenApoc, organisations only populate the graph when the applicable button at the top right is pressed, and only when there is a record of infiltration to report, for example

![image](https://user-images.githubusercontent.com/33913088/234589553-d2f365b2-59b1-4e3d-905a-965071b7e22e.png)

In the original game, on new game start, the screen appears as follows
![image](https://user-images.githubusercontent.com/33913088/234589972-e509bf10-741a-4f78-9ac4-e3ba2f92a611.png)

By selecting Organisations, you can add their "line" to the graph in an appropriate colour
These are overlayed where applicable with the left-right down most colour being layered at the top from the list of colours (and organisations) at the screen bottom
![image](https://user-images.githubusercontent.com/33913088/234590470-43bcb81f-94f5-4261-9d67-368df6eb101a.png)

As soon as there is an alien presence in a parent company building, an appropriate elevation in their infiltration level is indicated on the graph - this allows X-COM to better decide where to send agents to "investigate" for alien activity manually before an alert - it rises and falls with the number of aliens present in the city on an organisation by organisation basis

![image](https://user-images.githubusercontent.com/33913088/234591154-af55621b-0caa-4319-b069-fe04f68311dc.png)

Clearing the infections at Osiron will immediately drop their infiltration level in the example above

PR #1531, opened 2025-03-02
Alien infiltration screen update

Kurtsley:
Addresses #1211 
Here is the updated alien infiltration screen, changes:

- [x] Organizations added
- [x] Proper org order as per OG
- [x] Individual org select while keeping position
- [x] Fix curve? NO
- [x] Fix 640 x 480 glitch

Org spacing isn't perfect, but pretty close. I don't think the icon size is exact either, but I used the existing cityview org icons.

There is a problem on 640 x 480 where the icons won't appear until moused over. This is an issue that has popped up before.

There is a TODO about making curved lines, that may be above my skills.

Let me know if anything else is broken.

### DIFF
--- a/data/forms/city/infiltration.form
+++ b/data/forms/city/infiltration.form
@@ -45,7 +45,7 @@
 		<listbox id="ORG_SELECT_BOX" scrollbarid="ORG_SELECT_SCROLL">
 			<position x="420" y="60"/>
 			<size width="140" height="298"/>
-			<item size="0" spacing="0"/>
+			<item size="0" spacing="2"/>
 			<selcolour r="255" g="255" b="255" a="0"/>
 		</listbox>
 		<graphicbutton id="ORG_SCROLL_UP" scrollprev="ORG_SELECT_SCROLL">
--- a/data/forms/city/infiltration.form
+++ b/data/forms/city/infiltration.form
@@ -38,6 +38,30 @@
         <image/>
         <imagedepressed>BUTTON_OK_DEPRESSED</imagedepressed>
       </graphicbutton>
+		<scroll id="ORG_SELECT_SCROLL">
+			<position x="566" y="71"/>
+			<size width="26" height="278"/>
+		</scroll>
+		<listbox id="ORG_SELECT_BOX" scrollbarid="ORG_SELECT_SCROLL">
+			<position x="420" y="60"/>
+			<size width="140" height="298"/>
+			<item size="0" spacing="0"/>
+			<selcolour r="255" g="255" b="255" a="0"/>
+		</listbox>
+		<graphicbutton id="ORG_SCROLL_UP" scrollprev="ORG_SELECT_SCROLL">
+			<tooltip text="Scroll Up" font="smallset"/>
+			<position x="568" y="46"/>
+			<size width="22" height="21"/>
+			<image/>
+			<imagedepressed>BUTTON_SCROLL_UP_DEPRESSED</imagedepressed>
+		</graphicbutton>
+		<graphicbutton id="ORG_SCROLL_DOWN" scrollnext="ORG_SELECT_SCROLL">
+			<tooltip text="Scroll Down" font="smallset"/>
+			<position x="568" y="352"/>
+			<size width="22" height="21"/>
+			<image/>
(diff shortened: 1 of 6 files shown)


FilmBoy84:
Many thanks for doing this

As to the curve - if it holds up the feature, defer it as another issue and we can make a "Feature Request" entry accordingly to add it later. Really offering this missing functionality from the OG is a priority - beautifying it to match can always come later :)

I'll compile and check out the implementation over the next 36hours or so as time allows and get back to you on the 640x480 glitch

Is this another forms issue? Or are we looking at renderer do you think?

Kurtsley:
It was a form issue, should be fixed.

Kurtsley:
BTW, in OG the selected orgs are saved when exiting the screen. Right now the screen defaults to blank. Even though they are saved in OG, they don't appear to update when you reopen the screen. I found myself pushing the top ten button again to get the latest infiltration data. Should this be a priority?

FilmBoy84:
Had a play around and it works nicely
image

As to the graph being populated automatically

I have a couple of thoughts on this

* Should it automatically repopulate on opening with the "last" values shown - this would allow the user to then click the top right button and see the change although it would likely confuse new users or users who don't understand that it's a comparison?

* Should it automatically populate with the current level of infiltration (afterall, any history can be seen on the graph)

* Should it be blank until the button is clicked or an organisation is selected?

Overall, im inclined to say that the best QoL solution would be the second - have it automatically populate with all infected orgs shown/selected and the current state of infections. The user can then add/subtract organisations from this but you only need one click (entering the screen) to see current state of affairs at a glance - this is a welcome improvement on the original game

Kurtsley:
Indeed, second one it is. I'm going to say this is good for now, much better than what it was. I think more work needs to be done about how the graph actually samples the infiltration data and how often. The code suggests that each notch on the x-axis is one day, and that seems to be fairly accurate. In OG, if you see an infection starting on a certain day and it continues for multiple days, the notch where the graph starts to climb will be incremented backwards by on as each day passes. I'm not sure if it resets at exactly midnight or it takes the previous 24 hours into account when opening the graph.

In any case, I won't be able to set up a curved line without cheating so I will leave that up to others.

Kurtsley:
By the way, should that OG org order be used in other parts of the game such as the cityview org list?

FilmBoy84:
With regards to graph update intervals - i think it may be hourly or even every 30/15 minutes which may go some way to explaining the illusion of curves (more, tiny straight lines)

Here's the original game at 23:00 on the first day

I started a new game, ignored the alert at lifetree for first mission, ran forward time until the first UFO appeared at 23:00

This is the graph then...

image

Additionally, as the day starts at 12:00 midday, and we have moved ~1.5 blocks on the graph X-axis this suggests that half a day is 2 marks and a full day must be 4. However this doesn't really fit with the "major" demarcations on the axis so it could be fairly arbitrary; additionally, going forward, you will see things stretched and compressed and i cannot see a definite pattern

As to order, the following seems to vary according to level of infiltration - highest is always RED with the lowest always INDIGO
image

If you continue running not responding to anything, other orgs will add to the list and the top one is always the most infiltrated

Here's the same screen at Midnight -  just an hour later - note that we now have a full 2 marks on the X-Axis of data so an additional end of day calc must have been applied
image

At 6:40AM the next day, this is the graph - note the reordering of most infiltrated organisation at the bottom on the colour key despite them remaining in the same position on the list on the right
image

I also note that once again, the X-Axis marking doesn't seem to tally with time exactly however, the updates to the graph do appear to be quite regular

Forwarding to Midnight, Lifetree Aliens have "died off" so that infection is removed leaving just the remaining two orgs
image
They remain ordered by infiltration, this just coincides with right list order)

Forwarding another day to midnight, a few more orgs are infected due to alien growth in the city
Even though government is higher up list on right, it is less infected so is 2nd on the bottom list to Evonet
image

So, to surmise

* Right-hand list of orgs remains consistent throughout
* Lower "key" of colours reorders organisations to have the most infiltrated in RED and the least infiltrated in INDIGO
* Graph update interval needs data from reversing - it appears to be every 15/30/60 minutes but the graph expands and compresses time on the X-Axis so it's hard to ascertain which of those applies
* This PR is in a state good enough to merge - so i shall - any further updates can be done in future PRs and i shall make "issues" accordingly to cover what needs finishing

PR #1533, opened 2025-03-03
Fix mouse grab on Linux

Kurtsley:
So apparently on Linux, SetRelativeMouseMode also needs to be true for the mouse grab feature to work. I can't test on MacOS so no idea if this is working there. It doesn't seem any UI interactions have been affected with this change.


FilmBoy84:
Tried compiling builds for the following and all working as intended

Ubuntu
Mint under QubesOS
SteamOS (older 2.x Debian based) under QubesOS
Fedora under QubesOS

Merging

PR #1534, opened 2025-03-16
Don't follow vehicle when idle

Kurtsley:
Vehicles are not followed when idle, same with the soldiers in vehicles. An option has been added to prevent the following of vehicles on the strategy map as per OG, off by default.


FilmBoy84:
Works as intended
Option to switch for Map Layer works nicely

Tested with a Hovercar on Novice
* Selected starting hovercar
* Clicked "go to location" with follow enabled
* Screen follows and i cannot move away
* Hovercar reaches location, i can instantly pan the viewport
* Select a new location to travel to
* Open Map layer
*Also Follows
* Toggle option to not follow on map layer
* Screen does not follow

Merging

PR #1540, opened 2025-03-25
Fix alien craft having 0 speed

Kurtsley:
I didn't realize alien craft have no engine, this meant they could not move with my change. It is fixed now. The 03162025 release is completely broken and should be destroyed.


FilmBoy84:
Good catch and thanks for quick fix
Merging

PR #1541, opened 2025-04-04
Format with clang-format-18

JonnyH:
Were these missed from the branch? Have I somehow got my local version/settings wrong?


JonnyH:
Looks like a difference between minor clang-format versions - 18.1.3 vs my local 18.1.8 - a pain....

PR #1543, opened 2025-04-05
Fix infiltration screen blank when exceeding 42 days

Kurtsley:
The infiltration screen would not draw lines if the number of days in game exceeded the number of steps in the graph (42). 

Save to test:
[Sickve.-fixed.zip](https://github.com/user-attachments/files/19618687/Sickve.-fixed.zip)


FilmBoy84:
Thanks as always for the fix
Tested and working
🍻

Issue #1438, opened 2024-05-09
Fix Disruptor Multi-Bomb Launcher title display at equipment view 


ayrtondenner:
![image](https://github.com/OpenApoc/OpenApoc/assets/13112588/82461adf-2e49-4fec-8c55-07104febbe88)

Kurtsley:
Is this an EWM weapon? If not, how did the OG handle this?

ayrtondenner:
I started a new game and tried to see display of longest titles available:

![ufo2p_000](https://github.com/OpenApoc/OpenApoc/assets/13112588/2b658abb-d585-47b2-9c0e-7f9965760c47)

FilmBoy84:
> Is this an EWM weapon? If not, how did the OG handle this?

It's in the original game, Original game handles as follows

Vehicle Equip Screen - Single Line, some text out of bounds
![image](https://github.com/OpenApoc/OpenApoc/assets/33913088/20dc6eae-f911-4192-9e7d-e81b442e4b4d)

Marketplace Screen - Single Line, some text out of bounds
![image](https://github.com/OpenApoc/OpenApoc/assets/33913088/b4a44469-90d8-4757-843c-0744bf63a5b6)

So it appears OG forces single line, but still has text go out of the box 😅

Kurtsley:
This may be an issue, those lines that limit the text to one line are part of the background. I was under the impression that we can't edit those files in any way. Maybe we should just go with what the original developers did and cut off the text horizontally.

![Screenshot 2024-05-15 205213](https://github.com/OpenApoc/OpenApoc/assets/73447098/3dcc428d-30ac-408d-aca1-d9976d0ca0e2)

ayrtondenner:
Found another instance of weapon text being bigger than box size

![image](https://github.com/OpenApoc/OpenApoc/assets/13112588/bc7e78fe-ca8a-443c-a2ff-74706c292894)

Kurtsley:
At least that can be fixed because there are no lines surrounding the text.

PR #1550, opened 2025-07-30
Allow word wrapping in item and ammo names

Kurtsley:
Addresses #1438.

This seems like the best way to handle this since we can't edit the background picture.

The name is extended below the line.
longitem1

longitem2

There is no way to add more horizontal room here. This would be the hard limit for characters in an item name.
longitem3

Ammo can also be word wrapped.
longitem4

Obviously this is designed to work with the extended weapons mod since most of the fixes apply to those weapon names. This should probably be extensively tested with every long name, I didn't get that far yet. In theory this should work for armor and everything else.


FilmBoy84:
This is a nice solution
Whilst it only affects mods it's definitely needed so merging

Thanks as always 🍻

Issue #1559, opened 2025-11-02
Agents transferred or fired are not correctly removed from their current research/work projects leading to an error


FilmBoy84:
**Describe the bug**
Discord user Silverhaze provided the following savegame under OpenApoc 20250822

[Pre-fix Save 1.zip](https://github.com/user-attachments/files/23292400/Pre-fix.Save.1.zip)

The save will not load due to an error with AGENT_245

I unpacked the save and removed all entries relating to AGENT_245 from agents.xml , research.xml and buildings.xml before repacking into the save

On running the game and loading the fixed save time appeared to run, however, going immediately to "quit game" after running a few seconds, i received another error with AGENT_241

I repeated the clean process for this agent ID, repacked the save, and loaded the game, ran time on a few seconds and went to quit, whereup I was again faced with an error for AGENT_232

The clean, repack, and load game process was done for a further two cycles afterwards removing AGENT_355 and AGENT_356 respectively before the save would load, run a few seconds and then quit without error

**Expected behavior**
After removing all agent IDs relating to 245, 241, 232, 355 and 356 from the above save, the game will run and quit without issue as attached below:

[post-fix 1.zip](https://github.com/user-attachments/files/23292421/post-fix.1.zip)

However, the agents that were transferred are also removed - losing those agents

Really the game needs to act as the original game - as soon as agents are transferred or fired (i.e. on the confirmation request button click) any work or research assigned to the affected agents is cleared, any craft they may be in they are taken off the roster for, and they are either transferred or returned to the city / destroyed as normal

Settings CONF attached for 20250822

[OpenApoc_settings.zip](https://github.com/user-attachments/files/23292431/OpenApoc_settings.zip)

andremmfaria:
This issue appears to be fixed by #1565

PR #1565, opened 2026-01-08
Fix lab assignment on transfer

JonnyH:
Cleans up agent lab assignment, rather than keeping a separate boolean and StateRef<Lab> just use that ref being null to imply no lab assigned.

Though it changes the serialized format - removing that assigned_to_lab flag, it should actually affect backward compatibility with older saves - as that should already be implied by the non-null lab_assigned, and them being out of sync is a failure anyway. At next load, that unused flag should be ignored and future saves will not write it.

Compatibility of new saves with old game versions, however, will break.


JonnyH:
Fixes #1559

PR #1619, opened 2026-08-08
Fix damage to alien buildings affecting score

Kurtsley:
Addresses #1589

Damage to alien buildings will not negatively affect your score.


FilmBoy84:
Thanks for this Kurtsley 🍻 

Tested and no issues found, merging

PR #1626, opened 2026-08-25
So what do we do about this.

khallmark:


khallmark:
After discussion on discord, this is going to be HEAVILY broken up into something reasonable.

khallmark:
This PR is going to be carefully broken out into bug fixes, improvements, and extension points like a metal renderer that can be maintained independently from the core codebase.

PR #1627, opened 2026-08-26
fix(city): use-after-free in Base::die() when the base still has occupants

khallmark:
## What

`Base::die()` iterates two live `std::set`s while calling `die()` on each element, and those `die()` calls erase from the very sets being iterated:

```cpp
for (auto a : building->currentAgents)   { a->die(state, true); }
for (auto v : building->currentVehicles) { v->die(state, true); }
```

- `Agent::die()` → `currentBuilding->currentAgents.erase(thisRef)` (`game/state/shared/agent.cpp`)
- `Vehicle::die()` → `currentBuilding->currentVehicles.erase(...)` (`game/state/city/vehicle.cpp`)

This copies each container before iterating it. Six lines plus a comment; one file, +8/−2.

## Why

Incrementing a range-for's iterator after its underlying tree node has been freed is undefined behaviour. On libc++ it dereferences a freed red-black-tree node and segfaults.

That is the crash in #915, whose backtrace runs `Base::die` → `Battle::exitBattle` → the `BattleDebriefing` OK-button lambda — returning from a lost base-defence mission. The reporter also noted the `currentVehicles` loop looked like it had the same problem. It does, and both are fixed here.

`Vehicle::die()` already guards its own agent loop exactly this way, with a comment explaining the hazard. These two loops were simply missed.

## Testing

- Compiles clean against current `master` (`cmake --build build --target OpenApoc_GameState`).
- I have a regression test in my fork that reproduces the crash — exits 139 without this change, passes with it — but it leans on test helpers that don't exist here yet. Happy to follow up with it if you want the harness; I kept this PR to the fix alone so it stays easy to read.
- Originally found by running the engine unattended for long stretches. The crash reproduces reliably once a base with several occupants falls.

## Risks

Low. No API or behaviour change beyond not crashing — the same elements are killed in the same order. The only cost is copying two small containers on a path that runs once, when a base is destroyed.

## Links

- Fixes #915

## Checklist

- [x] Single-purpose change, one file
- [x] Builds against current `master`
- [x] Existing behaviour preserved
- [ ] Regression test — available on request, needs test-helper infrastructure not yet upstream

Issue #1617, opened 2026-06-23
[Cityscape] "Incorrect tile coordinates" while crafts fighting


AironT:
`E class OpenApoc::Tile *__cdecl OpenApoc::TileMap::getTile(int,int,int) Incorrect tile coordinates -1,82,12`
`E bool __cdecl OpenApoc::FlyingVehicleTileHelper::canEnterTile(class OpenApoc::Tile *,class OpenApoc::Tile *,bool,bool &,float &,bool &,bool,bool,bool) const No 'to' position supplied`

**Desktop (please complete the following information):**
 - OS: Windows 11 x64 25H2
 - OpenApoc Version: 20260419, portable
 - no mods

**Describe the bug**
Found such bug messages only for Battlescape, but I got this in Cityscape while my crafts were fighting with another craft and were continuously shifting to the end of the map. When they were to close to the border I started to get this error. It seems AI was trying to move crafts in battle as usual and new points were out of map border. And error occurs when craft is there.

**To Reproduce**
1. Get a bunch of air crafts and attack the building
2. Retreat to the outer city zone
3. Command to attack enemy crafts in Aggressive mode.
4. Wait when crafts go out of map border and See error

**Expected behavior**
AI move Crafts during fight only inside map border.

**Additional context**
It's not crashed the game. After clicking OK button game continue.

**Copy of your OpenApoc_settings.conf**
[OpenApoc_settings.zip](https://github.com/user-attachments/files/29236335/OpenApoc_settings.zip)

**Savegame**
[bug_fight_out_of_city_border.zip](https://github.com/user-attachments/files/29236182/bug_fight_out_of_city_border.zip)
Prepared a setup for fast testing. All already set, just set 4x speed and watch for crafts battle.

**Screenshots**
<img width="347" height="226" alt="Image" src="https://github.com/user-attachments/assets/72bdaa83-fbe4-47a8-bb9a-7a48bdc73945" />
<img width="567" height="209" alt="Image" src="https://github.com/user-attachments/assets/7cd35c0d-4e43-405e-a873-3224b7d85ead" />

PR #1634, opened 2026-08-27
fix(city): bounds-check the flying target scan before reading a tile

khallmark:
## What

`VehicleTargetHelper::adjustTargetToClosestFlying` has two candidate scans. The `Sidestep` branch bounds-checks before dereferencing:

```cpp
if (!map.tileIsValid(x, y, z))
{
    continue;
}
auto t = map.getTile(x, y, z);
```

The `PickNearbyPoint` branch, thirty lines above it in the same function, does not. This adds the same guard there. +7 lines, one file.

## Why

`PickNearbyPoint`'s loops walk outward from the midpoint — `x` from `midX - i` to `midX + i`, likewise `y` and `z` — so for a target near the edge of the city map they step off it and call `getTile()` with an out-of-range coordinate.

That's the error in #1617:

```
E class OpenApoc::Tile *__cdecl OpenApoc::TileMap::getTile(int,int,int) Incorrect tile coordinates -1,82,12
```

reported when craft fight near the map border and keep getting re-targeted outward.

I'm not proposing anything new here — the project already made this call in the sibling branch; this site was just missed.

## Testing

- Compiles clean against current `master`.
- No behaviour change for in-bounds candidates: the guard only skips coordinates that would have been an invalid lookup.

I have **not** run the reporter's repro, so I'd treat this as fixing the reported error rather than confirmed-closing the issue. #1617 also quotes a second error:

```
canEnterTile ... No 'to' position supplied
```

Skipping out-of-bounds candidates should remove the cause of that too, since the target never gets nominated — but I can't confirm it without the save, so I've said `Refs #1617` rather than `Fixes`.

## Risks

Very low. Pure guard, no logic change. Worst case is that a candidate which previously produced an error log and undefined behaviour is now skipped, and the scan continues to the next one.

## Links

- Refs #1617
- Same defect class as #1627 (a guard applied at one site and missed at its sibling)

## Checklist

- [x] Single-purpose change, one file
- [x] Builds against current `master`
- [x] Matches the existing idiom in the same function
- [ ] Reporter's save not replayed — see Testing

### DIFF
--- a/game/state/city/vehiclemission.cpp
+++ b/game/state/city/vehiclemission.cpp
@@ -829,6 +829,13 @@ VehicleTargetHelper::adjustTargetToClosestFlying(GameState &state, Vehicle &v, V
 						if (x == midX - i || x == midX + i || y == midY - i || y == midY + i ||
 						    z == midZ - i || z == midZ + i)
 						{
+							// The scan walks outward from the midpoint and can step off the
+							// map, so bounds-check before the lookup - same guard the
+							// Sidestep branch below already applies.
+							if (!map.tileIsValid(x, y, z))
+							{
+								continue;
+							}
 							auto t = map.getTile(x, y, z);
 							if (t->ownedObjects.empty())
 							{


FilmBoy84:
This does seem to resolve further instances of the error and I cannot replicate the original issue

Thanks, merging 🍻

PR #1636, opened 2026-08-27
fix(gl20-renderer): ask for a texture format the context actually has

khallmark:
## What

Chose the palette index texture format from the GL context version instead of hardcoding `GL_RED`: `GL_RED` at 3.0 and above, `GL_LUMINANCE` below. One value now serves as both internal and external format, replacing a hardcoded legacy component count of `1`.

One commit, one file — `framework/render/gl20/ogl_2_0_renderer.cpp`, +34/−2.

## Why

**OpenApoc renders a black window on macOS.** The intro plays, music plays, input works, and the mouse cursor draws correctly over an otherwise empty screen — so nothing looks obviously broken, which is presumably how this survived.

Palette index textures were uploaded with `GL_RED` as the external format:

```cpp
gl20::TexImage2D(gl20::TEXTURE_2D, 0, 1, parent->size.x, parent->size.y, 0, gl20::RED,
                 gl20::UNSIGNED_BYTE, l.getData());
```

`GL_RED` only became a legal external format for `glTexImage2D` in GL 3.0, via `ARB_texture_rg`. The GL2 backend does not get GL 3.0 on macOS: `displayInitialise` requests 3.0, the request fails, it lowers the version, and the resulting context reports `2.1 Metal - 90.5`.

On a 2.1 context `GL_RED` is `GL_INVALID_ENUM`. The upload is rejected, the texture is never populated, and the driver substitutes the zero texture — it says so directly:

```
UNSUPPORTED (log once): POSSIBLE ISSUE: unit 0 GLD_TEXTURE_INDEX_2D is unloadable
and bound to sampler type (Float) - using zero texture because texture unloadable
```

Every paletted sprite then samples index 0 and draws transparent. That is the entire game interface. RGB images take a different path and are unaffected, which is why the cursor still appears.

`GL_LUMINANCE` is the 2.1 spelling and replicates its single channel into `.r`, which is exactly what the palette shaders sample:

```glsl
float idx = texture2D(tex, texcoord, 0.0).r;
gl_FragColor = tint * texture2D(pal, vec2(idx, 0.0), 0.0);
```

Nothing errored at the call site, so this was localised by probing the shader rather than by reading it. Forcing the palette fragment shader to a constant filled the screen, so geometry and texcoords were sound; outputting the sampled index gave zero everywhere; outputting `texcoord` gave a clean 0→1 gradient. That left the texture itself, and its upload was the only call on that path using a 3.0-era enum.

## Testing

- [ ] Unit tests added/updated
- [x] Manual testing completed
- [x] No new tests needed — this is a driver-format branch with no seam the existing harness can reach; the existing suite is unchanged and still passes

Verified on an M2 Max, macOS, where the context is `2.1 Metal - 90.5`: with this commit the MicroProse intro and the Apocalypse main menu render in full colour; without it the window is empty. Existing tests pass and CI is green on both workflows.

I have no way to exercise a Linux or Windows GL 2.1 driver from here, so I have not tested those.

## Risks

Low, and bounded by the version check. Contexts reporting GL 3.0 or above take the same `GL_RED` path they take today, so every configuration that currently works is untouched — the change only affects the case that is already broken. The format is resolved once into a function-local static on first use, after the context is current.

The untested surface is a GL 2.1 context on Linux or Windows. Those move from `GL_RED` (rejected) to `GL_LUMINANCE` (accepted), which should only ever be an improvement, but I state it plainly rather than imply coverage I do not have.

Rollback is reverting the single commit.

## Links

N/A — no tracker issue; found while working on window resizing, not from a filed report.

Context, if useful: this came out of a larger piece of work on making the window arbitrarily resizable with the UI chrome staying put, staged at [khallmark/OpenApoc](https://github.com/khallmark/OpenApoc/pulls). This fix is independent of all of it and stands alone — I would rather send them one at a time than drop it all at once.

## Checklist

- [x] Follows repo conventions
- [x] Self-review completed
- [x] Docs updated (if needed) — N/A, no user-facing behaviour or option changes
- [x] All tests pass

### DIFF
--- a/framework/framework.cpp
+++ b/framework/framework.cpp
@@ -51,6 +51,80 @@ using namespace OpenApoc;
 namespace OpenApoc
 {
 
+namespace
+{
+
+// Relative option defaults from options.cpp.
+constexpr const char *kDefaultDataPath = "./data";
+constexpr const char *kDefaultModPath = "./data/mods";
+
+// True if the path is a directory that actually holds OpenApoc data. A directory that exists
+// but holds none of these markers is a leftover, not a data directory.
+bool dataPathLooksValid(const UString &path)
+{
+	if (path.empty())
+	{
+		return false;
+	}
+	std::error_code ec;
+	fs::path p(path);
+	if (!fs::is_directory(p, ec))
+	{
+		return false;
+	}
+	return fs::exists(p / "mods", ec) || fs::exists(p / "fonts", ec) ||
+	       fs::exists(p / "city", ec) || fs::exists(p / "forms", ec);
+}
+
+// The settings file keeps whatever Framework.Data was last used, and that path can simply stop
+// existing -- a deleted checkout or worktree is the usual way. Nothing validated it, so the game
+// carried on with a dead path: every asset load failed one at a time, and the first visible
+// symptom was an unrelated-looking mod error rather than "your data path is gone".
+void applyDataPathFallback()
+{
+	const UString configured = Options::dataPathOption.get();
+	if (dataPathLooksValid(configured))
+	{
+		return;
+	}
+	if (dataPathLooksValid(kDefaultDataPath))
+	{
+		LogWarning("Data path \"{0}\" is missing; falling back to \"{1}\"", configured,
+		           kDefaultDataPath);
+		Options::dataPathOption.set(kDefaultDataPath);
+		return;
+	}
+	LogWarning("Data path \"{0}\" is missing and no fallback was found", configured);
+}
+
+// Mods live inside the data directory, so the mod path has to follow it. ModPath was resolved
+// independently, against the working directory, so pointing Framework.Data at a data directory
+// elsewhere left mods unfindable -- and unlike every other asset that failed with a hard error.
+//
+// Keyed off the *value* rather than config().optionOverridden(): the settings file routinely
+// carries a persisted "ModPath=./data/mods", which is the default written back out, not a
+// choice the user made. Treating that as an override is what left the mod path pinned.
+void applyDataRelativeModPath()
+{
+	if (Options::modPath.get() != kDefaultModPath)
+	{
+		return;
+	}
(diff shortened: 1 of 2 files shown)


JonnyH:
I don't think this is correct - as the compatiblity modes including GL_EXT_texture_rg (by definition) require GL2.x compatibility - formats included - requring enums added by gl3.x is incorrect. If you limit it to 3.x compatiblity there's a million other things that "should" break too - like shaders and their bindings.

Perhaps this is a platform bug? And somewhat recent introduced? I've certainly run the gl2.x code on MacOS devices previously.

EDIT: It's not entirely impossible that apple *removed* support for GL_EXT_texture_rg in the meantime (having previously determined, over a decade ago, that was supported on functionally every device), but... *why*? I'll admit I've not run an apple platform for a while, but "weird specific platform workarounds" would push them outside the border of what I'm willing to support.

Could this be a functional "workaround" for a platfom issue, that *happens* to be avoided by changing the texel format? It wouldn't be the first time for a macos system :)

khallmark:
Apple Silicon happened.

Let me investigate it a little further.

khallmark:
Thank you for taking the time to review my contribution. I'm moving this to draft status pending further research based on your feedback.

JonnyH:
Probably the "best" thing to do is to do the (relatively) small tweaks to get the GLES3 renderer working on macos - it's targetted at "OpenGLES-3", but that was more on the assumption that was the "lowest common denominator" for that era of hardware, but (for whatever reason.....) Apple decided not to advertise that as supported on their "desktop" devices. That reenderer has clear advantages - specificaly around "batching" similar draw commands, that really make kinda pointless to attempt to "support" the GL2.x renderer unless *absolutely* necessary,

As an aside - as someone who has personally worked on Apple's GPU software stack I hate that this sort of thing is necessary, But here we are :)

PR #1640, opened 2026-08-27
Fix agents keeping lab and craft assignments when fired or transferred

### LINKED ISSUE #1559: Agents transferred or fired are not correctly removed from their current research/work projects leading to an error

andremmfaria:
Fixes #1559

## Problem

Firing or transferring an agent left their lab assignment (and craft roster membership) intact. Saves then contained dangling agent references in `research.xml` and `buildings.xml`, which made them fail to load with errors like `AGENT_245` not found. The fire path relied on UI-side cleanup duplicated across screens, and `Agent::transfer()` only cleared the lab assignment, never the vehicle roster.

## Fix

- New shared `Agent::leaveLabAndCraft()` clears the lab assignment and vehicle roster membership in one place; called from `Agent::transfer()` and `Agent::die()` (the fire path).
- Removed the now-redundant inline lab-removal logic from `RecruitScreen` and `TransferScreen`.
- Added `test_transfer_removes_from_vehicle()` to `tests/test_lab_assignment.cpp`.

## Verification

- `OpenApoc_GameState` and `OpenApoc_GameUI` build clean (cmake + ninja), no new warnings.
- clang-format clean.
- New test compiles; it needs extracted game data to run, so it relies on CI for execution.

## Notes

- Scope is limited to the fire/transfer paths from the issue; other agent-removal paths (capture, base loss) are untouched.

### DIFF
--- a/game/state/shared/agent.cpp
+++ b/game/state/shared/agent.cpp
@@ -7,6 +7,7 @@
 #include "game/state/city/building.h"
 #include "game/state/city/city.h"
 #include "game/state/city/facility.h"
+#include "game/state/city/research.h"
 #include "game/state/city/vehicle.h"
 #include "game/state/gameevent.h"
 #include "game/state/gamestate.h"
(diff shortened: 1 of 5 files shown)


andremmfaria:
Closing: after checking against current master, this appears to already be fixed by #1565 (merged 2026-01-08). The reporter's save in #1559 was from build 20250822, which predates that fix. On current master the fire path goes through Agent::die() (which clears lab and craft assignments) and Agent::transfer() clears the lab assignment, so the dangling agent references should no longer occur. #1559 likely just needs verification on a current build.

PR #1646, opened 2026-08-31
Launcher: add CD/ISO autodetection

### LINKED ISSUE #1645: Launcher: autodetect X-COM: Apocalypse CD/ISO path
## Summary

Add CD/ISO autodetection to the launcher.

On startup, if `Framework.CD` is missing, default, or invalid, the launcher should search for a valid X-COM: Apocalypse data source and prefill the existing `CD Path` field. Also add an `Autodetect` button beside `CD Path` so users can rerun the same flow manually.

## Current state

- `Framework.CD` already exists and defaults to `./data/cd.iso`.
- Launcher already displays/saves `CD Path`.
- Runtime already mounts `Framework.CD` through PhysFS and checks for `music`.
- Windows installer already has Steam/GOG detection logic that can be used as reference, but it should be modernised for the launcher rather than copied verbatim.

## Desired behaviour

On launcher startup:

- keep an existing valid user-configured path
- otherwise search for a valid game-data source
- if exactly one candidate is found, populate `CD Path`
- if multiple candidates are found, let the user choose
- if none are found, leave the field unchanged and allow manual browsing

Add an `Autodetect` button next to `CD Path` that runs the same flow again.

## Detection sources

- local `./data/cd.iso`
- Steam app `7660`, using modern `steamapps/libraryfolders.vdf` + `appmanifest_7660.acf` discovery
- GOG install, checking `XCOM.cue` plus adjacent `XCOM.BIN`
- mounted/extracted retail CD directory

## Implementation notes

Likely files:

- `tools/launcher/launcherwindow.ui`
- `tools/launcher/launcherwindow.h`
- `tools/launcher/launcherwindow.cpp`
- `tools/launcher/CMakeLists.txt` if discovery is split into helper files

Best startup hook is in `LauncherWindow::LauncherWindow()` after config parsing and before creating `Framework("OpenApoc_Launcher", false)`, because `Framework` immediately consumes `Framework.CD`.

Reuse the existing `Framework.CD` option; no new config key should be needed.

Existing Windows installer detection reference:

- `install/windows/installer.nsi` has `ScanSteam`, `ScanSteamLibrary`, and `ScanGOG`
- that code currently checks Steam via older `config.vdf` `BaseInstallFolder_*`; launcher implementation should prefer modern `libraryfolders.vdf` discovery

## Non-goals

- Do not bundle or download original game data.
- Do not overwrite a valid explicit user path.
- Do not add physical CD imaging in this issue.

andremmfaria:
## Summary

Closes #1645.

On launcher startup, if the configured CD path (`Framework.CD`, default `./data/cd.iso`) is missing or doesn't look like a valid X-COM: Apocalypse data source, the launcher now scans a small set of known locations for one:

- local `./data/cd.iso`
- Steam library folders (appid 7660, `steamapps/common/<installdir>/cd.iso`, parsed from `libraryfolders.vdf` / `appmanifest_7660.acf`)
- GOG's registry install path (product 1445249430, `CD/XCOM.cue`, Windows only)
- mounted/extracted volumes (via `QStorageInfo::mountedVolumes()`, looking for a top-level `music` entry)

If exactly one plausible candidate is found, it's silently written to the CD path option. If several are found, a modal picker lets the user choose (Cancel leaves the value unchanged). If none are found, the field is left as-is and no dialog appears.

A new "Autodetect" button was added to the Paths tab (next to the existing browse buttons) that reruns the same scan on demand, always showing the picker even for a single match, since that's a deliberate user action.

This is detection/UX sugar only: no game data is bundled, downloaded, or imaged; no new config key was introduced (`Framework.CD` is reused); and full validation still happens where it already did (the `music` probe in `Framework`'s constructor). Candidate checks are lightweight plausibility checks (existence + extension, `.cue` needs an adjacent `.bin`, a directory needs a case-insensitive `music` entry) with no PhysFS mounting during detection.

The detection logic lives in new files `tools/launcher/cddetect.h`/`.cpp`, exposing two free functions (`isPlausibleCDPath`, `detectCandidates`); no new detector class hierarchy, no async/background scanning, no third-party VDF parsing library.

## Test plan

Manual test matrix run in isolated scratch directories (via `portable.txt`, with `HOME` overridden to fake scratch homes so no real user config or Steam install was touched):

- [x] No-config first run, no candidates present: CD path field stays at the default, no dialog, no crash.
- [x] Valid existing CD path (directory containing a `music` marker file) is left unchanged, no detection scan runs.
- [x] Fake Steam library fixture (`libraryfolders.vdf` + `appmanifest_7660.acf` + placeholder `cd.iso`, matching the real Steam appid 7660 `installdir` of `XCom Apocalypse`): exactly one candidate found, silently prefilled.
- [x] Two candidates present (local `./data/cd.iso` + Steam fixture): the modal picker is shown (confirmed the launcher blocks before Framework/logging init starts, i.e. before reaching the picker's result); interactive confirm/cancel of the picker itself isn't automatable headless (no synthetic Qt input injection available in this environment).
- [ ] Autodetect button rerun with 1 candidate still showing the picker: verified by code inspection only (`forceChooser=true` unconditionally skips the single-candidate shortcut) — not automatable headless since it requires clicking a button in the running GUI.

Build: `cmake --build build --target OpenApoc_Launcher` succeeds cleanly on Linux (Qt 6.10.2) with no new compiler warnings from `cddetect.cpp`/`launcherwindow.cpp`/`.h`.

**Not verified**: Windows registry paths (Steam `HKEY_CURRENT_USER\Software\Valve\Steam`, GOG `HKEY_LOCAL_MACHINE\Software\GOG.com\Games\1445249430` and its `WOW6432Node` fallback) are implemented per the existing NSIS installer's `ScanSteamLibrary`/`ScanGOG` logic but untested on a real Windows machine — flagging for reviewers with Windows access to spot-check, along with the "Autodetect" button's placement/visibility in the Paths tab grid (verified analytically via widget minimum-width sums given no screenshot tooling was available in this environment; not screenshotted).

### DIFF
--- a/tools/launcher/CMakeLists.txt
+++ b/tools/launcher/CMakeLists.txt
@@ -12,6 +12,8 @@ qt_standard_project_setup()
 set (LAUNCHER_SOURCE_FILES
 	main.cpp
 	launcherwindow.cpp
+	cddetect.h
+	cddetect.cpp
 	launcherwindow.ui)
 
 set(sources ${sources} "apocicon.rc")
--- a/tools/launcher/cddetect.cpp
+++ b/tools/launcher/cddetect.cpp
@@ -1,0 +1,258 @@
+#include "cddetect.h"
+
+#include <QDir>
+#include <QFile>
+#include <QFileInfo>
+#include <QRegularExpression>
+#include <QSettings>
+#include <QStorageInfo>
+#include <QTextStream>
+
+namespace OpenApoc::CDDetect
+{
+
+// Returns true if `dir` (non-recursively) contains a file named `name`, matched
+// case-insensitively.
+static bool caseInsensitiveFileExists(const QDir &dir, const QString &name)
+{
+	for (const auto &entry : dir.entryList(QDir::Files))
+	{
+		if (entry.compare(name, Qt::CaseInsensitive) == 0)
+		{
+			return true;
+		}
+	}
+	return false;
+}
+
+bool isPlausibleCDPath(const QString &path)
+{
+	if (path.isEmpty())
+	{
+		return false;
+	}
+
+	const QFileInfo info(path);
+
+	if (info.isFile())
+	{
+		const QString suffix = info.suffix();
+		if (suffix.compare("iso", Qt::CaseInsensitive) == 0)
+		{
+			return true;
+		}
+		if (suffix.compare("cue", Qt::CaseInsensitive) == 0)
+		{
+			const QDir cueDir = info.dir();
+			QFile cueFile(path);
+			if (cueFile.open(QIODevice::ReadOnly | QIODevice::Text))
+			{
+				const QString text = QTextStream(&cueFile).readAll();
+				static const QRegularExpression binRegex(
+				    QStringLiteral("FILE\\s+\"?([^\"\\r\\n]+\\.bin)\"?"),
+				    QRegularExpression::CaseInsensitiveOption);
+				const auto match = binRegex.match(text);
+				if (match.hasMatch())
+				{
+					const QString binName = QFileInfo(match.captured(1)).fileName();
+					if (caseInsensitiveFileExists(cueDir, binName))
+					{
+						return true;
+					}
+				}
+			}
+			// Fall back to the conventional GOG naming if the FILE directive didn't yield a
+			// match (or the referenced .bin doesn't exist alongside the .cue).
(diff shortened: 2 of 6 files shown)


andremmfaria:
Superseded by #1647 (same commit, branch renamed to include the issue number).

PR #1649, opened 2026-08-31
Fix crash when a battle unit dies mid-update (null owningTile deref) (#1556)

### LINKED ISSUE #1556: Crash in battlescape due to null Agent->tileObject
Reported on Discord by Peter987

Originally reported on OpenApoc-20250822 but reproduced locally on latest master too (b93cffe3a939915756616dba615ada200115d32d) on both win32 and linux

Repro steps:

- Load provided save
- Unpause

andremmfaria:
Fixes #1556.

## Root cause

A `BattleUnit` that dies with `destroy = true` partway through its own `BattleUnit::update()` tick keeps executing the rest of that tick's movement/falling logic. `die()` calls `tileObject->removeFromMap()`, which nulls the tile object's `owningTile`, but the update path only guarded on `retreated`, never `destroyed`. The movement loop then dereferences the null tile: `Tile::getCanStand(this=0x0)` via `BattleUnit::updateCheckBeginFalling()`, matching the backtrace in the issue (Popper dying mid-tick from its self-destruct).

## Fix

Two commits:

1. Mirror the existing `retreated` guards for `destroyed`: at `updateCheckBeginFalling()` entry, at the movement-loop break, and at the post-loop return. Also set `destroyed = true` before `tileObject->removeFromMap()` in `die()` so the flag is visible to anything that runs during removal.
2. While validating, a second entry point for the same bug class surfaced: a unit that dies before the movement loop is entered (e.g. bleeding out via `updateWoundsAndHealing()` inside `updateStateAndStats()`) crashed in `Tile::getRestingPosition(this=0x0)` via `updateMovementNormal()`. Added the same `destroyed || retreated` guard at the top of the movement-loop body.

## Verification

The issue's attached save no longer loads on current master (it predates 8c5621f0, which moved `SceneryTileType` to a single `StateRefMap` and broke old saves), so verification used a headless harness that recreates the condition: a base-defense battle against a single `AGENTTYPE_POPPER`, killed via a lethal fatal wound resolved inside `updateStateAndStats()` so `die()` runs mid-tick, taking the destroy branch through the Popper's self-destruct.

* With `battleunit.cpp` reverted to pre-fix master: SIGSEGV in `Tile::getCanStand(this=0x0)` from `updateCheckBeginFalling()`, the exact signature from the issue.
* With this branch: clean pass, and a further update tick on the destroyed unit is a safe no-op. Full ctest suite green (RelWithDebInfo, Linux).

The harness code is posted as a comment below for reference.

### DIFF
--- a/game/state/battle/battleunit.cpp
+++ b/game/state/battle/battleunit.cpp
@@ -2033,6 +2033,15 @@ void BattleUnit::update(GameState &state, unsigned int ticks)
 			lastHandsTicksRemaining = handsTicksRemaining;
 			lastTurnTicksRemaining = turnTicksRemaining;
 
+			// A unit can also die before this loop is ever entered (e.g. bleeding
+			// out from a fatal wound in updateStateAndStats(), called earlier in
+			// update()). Bail before touching movement/body/hands state, which
+			// assume a live tileObject with a non-null owningTile.
+			if (destroyed || retreated)
+			{
+				break;
+			}
+
 			updateCheckBeginFalling(state);
 			updateBody(state, bodyTicksRemaining);
 			updateHands(state, handsTicksRemaining);
--- a/game/state/battle/battleunit.cpp
+++ b/game/state/battle/battleunit.cpp
@@ -2039,9 +2039,14 @@ void BattleUnit::update(GameState &state, unsigned int ticks)
 			updateMovement(state, moveTicksRemaining, wasUsingLift);
 			updateTurning(state, turnTicksRemaining, handsTicksRemaining);
 			updateDisplayedItem(state);
+
+			if (destroyed || retreated)
+			{
+				break;
+			}
 		}
 	}
-	if (retreated)
+	if (destroyed || retreated)
 	{
 		return;
 	}
@@ -2636,7 +2641,7 @@ void BattleUnit::updateCrying(GameState &state)
 
 void BattleUnit::updateCheckBeginFalling(GameState &state)
 {
-	if (retreated)
+	if (destroyed || retreated)
 	{
 		return;
 	}
@@ -4873,8 +4878,8 @@ void BattleUnit::die(GameState &state, StateRef<BattleUnit> attacker, bool viole
 		{
 			shadowObject->removeFromMap();
 		}
-		tileObject->removeFromMap();
 		destroyed = true;
+		tileObject->removeFromMap();
 	}
 }
 


andremmfaria:
The headless validation harness used for verification (core portion). It stages a base-defense battle (fixed map, no random generation, player squad auto-filled from base personnel) against one Popper, then kills it with a lethal fatal wound so `die()` runs inside the unit's own `update()` tick, taking the destroy branch via the Popper's self-destruct item:

```cpp
// Stage a base-defense battle: a single Popper invades the player's own base.
building->current_crew[popperType] = 1;

std::list<StateRef<Agent>> playerAgents;
                    StateRef<Vehicle>{}, building);
Battle::enterBattle(*state);
```

```cpp
static bool test_destroyed_unit_mid_update_no_crash(const sp<GameState> &state)
{
	auto unit = findUnitOfType(state, "AGENTTYPE_POPPER");
	if (!unit || !unit->tileObject || !unit->tileObject->getOwningTile())
		return false;
	if (unit->destroyed || unit->retreated)
		return false;

	// Ground the unit so updateCheckBeginFalling()'s getCanStand() branch (the
	// original crash site) is actually reached on the next movement-loop pass.
	unit->falling = false;
	unit->current_movement_state = MovementState::None;

	// Lethal fatal wound: resolved by updateWoundsAndHealing(), called from
	// updateStateAndStats() before the movement loop, so die() runs *during*
	// this update() call rather than before it.
	unit->agent->modified_stats.health = 1;
	unit->fatalWounds[BodyPart::Body] = 100;

	// One wound-tick's worth of ticks so the bleedout resolves within this
	// single update() call.
	unit->update(*state, TICKS_PER_WOUND_EFFECT);

	if (!unit->destroyed)
		return false;
	// die()'s destroy branch calls tileObject->removeFromMap(), which nulls the
	// tile object's owningTile while the tileObject itself stays alive. This
	// null-owningTile-while-destroyed state is exactly what used to crash
	// unguarded code downstream.
	if (!unit->tileObject || unit->tileObject->getOwningTile())
		return false;

	// A further tick on the now-destroyed unit must be a safe no-op.
	unit->update(*state, TICKS_PER_WOUND_EFFECT);
	return true;
}
```

On pre-fix `battleunit.cpp` this segfaults in `Tile::getCanStand(this=0x0)` from `BattleUnit::updateCheckBeginFalling()` (the issue's exact backtrace); on this branch it passes. Happy to turn this into a proper CTest target in a follow-up if there's interest.

FilmBoy84:
>Happy to turn this into a proper CTest target in a follow-up if there's interest.

Definitely interested in anything you submit to resolve the segfault as a separate Issue/PR

For now, as this resolves the linked crashes, i shall merge 🍻

PR #1650, opened 2026-08-31
Fix out-of-bounds tile lookups during dogfights near the city border (#1617)

andremmfaria:
Refs #1617, complements #1634.

## Root cause

`FlyingVehicleTileHelper::findSidestep` applies a random ±1.0 float offset to x/y but clamps only z, so at the city border `getTile()` gets out-of-map coordinates, producing exactly the issue's `Incorrect tile coordinates` + `No 'to' position supplied` pair. Worse, float→int truncation maps a marginally negative result like `{-0.003, …}` back to tile 0, so it is accepted and the craft sits slightly off-map.

The projectile dodge generator has the same truncation problem: it builds `Vec3<int>` candidates from `vehicle.position + offset` floats, so near the low border a `-1` offset collapses onto the vehicle's own tile (`FromPos == ToPos` error) instead of the intended neighbour.

#1634 covered a third site (`PickNearbyPoint`); this PR covers the remaining two.

## Fix

1. Clamp x/y in `findSidestep` to `[0, size - 0.1]` alongside the existing z clamp; build dodge candidates from the owning tile's integer position instead of truncated float sums.
2. Harden `TileObject::setPosition`: it accepted positions up to a full tile past the map edge and clamped to `size + 1`, itself still invalid, yielding a nullptr tile after the object was already removed from the map. It now clamps anything outside `[0, size)` to `size - 0.1`.

## Verification

Headless harness (core posted as a comment below): a flying vehicle placed at each map corner, seeded RNG, 300 `findSidestep` calls per corner asserting in-bounds results with a valid backing tile.

* Pre-fix `vehiclemission.cpp`: fails on iteration 1 with `out-of-bounds position {-0.0027151704,1.2609513,6.5} from corner {0.3,0.3,6.5}`.
* This branch: clean pass; full ctest suite green (RelWithDebInfo, Linux, extracted CD data).

The dodge generator isn't reachable from a test binary (no header for `FlyingVehicleMover`) and its pre-fix failure is a wrong-but-filtered candidate, so it's fixed by construction. Reporter's save not replayed, hence `Refs` rather than `Fixes`.

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -288,6 +288,9 @@ class FlyingVehicleMover : public VehicleMover
 				//   Otherwise we think we'll be moving too slow to dodge (basically moving
 				//   backwards into projectile or forwards alongside it and still getting hit)
 				auto point2d = glm::normalize(Vec2<float>{point.x, point.y});
+				// Use the vehicle's owning tile (integer) as the base so offsets don't get
+				// truncated toward zero and collapse onto the vehicle's own tile near 0
+				auto ownTile = vehicle.tileObject->getOwningTile()->position;
 				// Gather all allowed dodge locations according to the rules above
 				std::list<Vec3<int>> possibleDodgeLocations;
 				for (int x = -1; x <= 1; x++)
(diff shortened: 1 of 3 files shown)


andremmfaria:
The headless verification harness (core portion). It takes a player flying vehicle still parked in a building, places it in the city via `leaveBuilding()`, then hammers `findSidestep` from each map corner with a seeded RNG, asserting every result is inside the map with a valid backing tile:

```cpp
auto &map = *state->current_city->map;
Vec3<int> mapSize = map.size;
float midZ = static_cast<float>(mapSize.z) / 2.0f;

std::vector<Vec3<float>> cornerPositions = {
    {0.3f, 0.3f, midZ},
    {mapSize.x - 0.4f, mapSize.y - 0.4f, midZ},
    {0.3f, mapSize.y - 0.4f, midZ},
    {mapSize.x - 0.4f, 0.3f, midZ},
};

vehicle->leaveBuilding(*state, cornerPositions.front(), 0.0f);

FlyingVehicleTileHelper helper(map, *vehicle);
const int iterationsPerPosition = 300;
const float distancePref = 2.0f;

state->rng.seed(1617);

for (auto &corner : cornerPositions)
{
	vehicle->setPosition(corner);

	for (int i = 0; i < iterationsPerPosition; i++)
	{
		auto result =
		    helper.findSidestep(*state, vehicle->tileObject, vehicle->tileObject, distancePref);

		if (result.x < 0.0f || result.x >= static_cast<float>(mapSize.x) || result.y < 0.0f ||
		    result.y >= static_cast<float>(mapSize.y) || result.z < 0.0f ||
		    result.z >= static_cast<float>(mapSize.z))
		{
			LogError("findSidestep returned out-of-bounds position {0} from corner {1} "
			        "(map size {2}), iteration {3}",
			        result, corner, mapSize, i);
			return false;
		}

		if (map.getTile(result) == nullptr)
		{
			LogError("findSidestep returned position {0} with no backing tile from corner "
			        "{1}, iteration {2}",
			        result, corner, i);
			return false;
		}
	}
}
```

On pre-fix `vehiclemission.cpp` this fails on the very first iteration from the `{0.3, 0.3}` corner with `findSidestep returned out-of-bounds position {-0.0027151704,1.2609513,6.5}`; on this branch all 4×300 iterations pass. Happy to turn this into a proper CTest target in a follow-up if there's interest.

PR #1657, opened 2026-09-03
Fix organisation tech level never increasing (#1587)

### LINKED ISSUE #1587: Org Tech level does not improve
Hello, couldn't find this anywhere. Sorry if I missed it.

Windows 11 Home 25H2
OpenApoc Build 20260123
No mods

**Describe the bug**
Organisations do not improve their tech level and equipment with time

**To Reproduce**
1. Pass through weeks
2. Raid orgs (e.g. Cult of Sirius)
3. Equip gained during organization raids stays the same over weeks - e.g. still finding Lawpistols in Cult raids in week 7.

**Expected behavior**
According to Wiki, each week the Org Tech levels should increase by 1. This means they eventually start using Alien tech - this definitely happened in OG. Cult should be at lvl 8 in week 7, dropping disruptor weapons and alien grenades, Lawpistol should not be dropped anymore.

**Savegame**
In the organisations.xml, you can see the Tech level is still at the starting value.
[NoTechIncrease.zip](https://github.com/user-attachments/files/25815737/NoTechIncrease.zip)

**Screenshots**
Here's the drop from 6 Cult raids in week 7
Image

**Additional context**
n/a

[settings.zip](https://github.com/user-attachments/files/25744255/settings.zip)

andremmfaria:
Fixes #1587.

**Root cause:** `Organisation::tech_level` is set once by the extractor and never changes at runtime. `GameState::updateEndOfWeek` has a per-organisation finance loop but nothing for tech level, so every organisation stays at its starting level for the whole game. The original game raises each organisation's tech level by 1 per week, capped at 12.

**Fix:** in `updateEndOfWeek`, when not `gameStart`, increment `tech_level` for every organisation except `player`, `aliens`, and `civilian`, capped at the highest `min_score` among `Human` `EquipmentSet`s (12 with the shipped data) instead of a hardcoded constant. The separate "+3 on full infiltration" rule is not part of this change.

**Verification**
- Pre-fix headless harness: all 25 non-exempt organisations unchanged after a week (`ORG_CULT_OF_SIRIUS tech_level 2 -> 2, expected 3`).
- Post-fix: weekly +1 with exemptions, cap holds at and beyond the derived value, `gameStart` week leaves levels untouched.
- Full ctest green (RelWithDebInfo, Linux); fork CI Lint + CMake green.

Harness core in the comment below.

### DIFF
--- a/game/state/gamestate.cpp
+++ b/game/state/gamestate.cpp
@@ -1465,6 +1465,22 @@ void GameState::updateEndOfWeek(bool gameStart)
 
 	if (!gameStart)
 	{
+		int maxOrgTechLevel = 1;
+		for (auto &es : equipment_sets)
+		{
+			if (es.second->type == EquipmentSet::Type::Human)
+			{
+				maxOrgTechLevel = std::max(maxOrgTechLevel, es.second->min_score);
+			}
+		}
+		for (auto &[id, org] : organisations)
+		{
+			if (id != player.id && id != aliens.id && id != civilian.id)
+			{
+				org->tech_level = std::min(org->tech_level + 1, maxOrgTechLevel);
+			}
+		}
+
 		for (auto &c : this->cities)
 		{
 			c.second->weeklyLoop(*this);


andremmfaria:
Temporary harness used for the before/after check (not committed). Loads difficulty0 like `test_lab_assignment` and runs one real weekly tick:

```cpp
int maxOrgTechLevel = 1;
for (auto &es : state->equipment_sets)
	if (es.second->type == EquipmentSet::Type::Human)
		maxOrgTechLevel = std::max(maxOrgTechLevel, es.second->min_score);

std::map<UString, int> before;
for (auto &org : state->organisations)
	before[org.first] = org.second->tech_level;
state->updateEndOfWeek(false);
for (auto &org : state->organisations)
{
	bool exempt = org.first == state->getPlayer().id || org.first == state->getAliens().id ||
	              org.first == state->getCivilian().id;
	int expected = exempt ? before[org.first] : std::min(before[org.first] + 1, maxOrgTechLevel);
	assert(org.second->tech_level == expected);
}
assert(EquipmentSet::getForType(*state, EquipmentSet::Type::Human, maxOrgTechLevel));
assert(EquipmentSet::getForType(*state, EquipmentSet::Type::Human, maxOrgTechLevel + 5));

for (auto &org : state->organisations) before[org.first] = org.second->tech_level;
state->updateEndOfWeek(true);
for (auto &org : state->organisations)
	assert(org.second->tech_level == before[org.first]);
```

Pre-fix (`727cd974`):
```
E Org 'ORG_CULT_OF_SIRIUS' tech_level 2 -> 2, expected 3
E Weekly increment test failed
```

Post-fix (`b69d29a3`): all checks pass, derived cap = 12.

FilmBoy84:
Really nice catch!
Will be great to have organisation tech advancing once more
Thanks as always 🍻

PR #1660, opened 2026-09-06
Battle: make the Mind Shield toggle psi defence (#1583)

### LINKED ISSUE #1583: Mind Shield doesn't work
Hello, couldn't find this anywhere. Feel free to re-classify as Feature request but it's a deviation from original, so I create it as a bug.

Windows 11 Home 25H2
OpenApoc Build 20260123
No mods

**Describe the bug**
Mind Shield cannot be used.

**To Reproduce**
1. Acquire Mind Shield from a Marsec raid
2. Click on Mind Shield to activate it
3. Nothing happens

**Expected behavior**
Using Mind Shield should grant 20 PSI Defense. In the original game, the bonus became permanent at end of mission, almost certainly a bug.

**Savegame**
[MindShield.zip](https://github.com/user-attachments/files/25737817/MindShield.zip)

**Screenshots**
Image

**Additional context**
Mind Shield was an unfinished/cut content, only available by raiding Marsec. Activating it provided a 20 point boost to PSI Defense. If mission finished or soldier left mission while active, the bonus became permanent, and could be repeated infinitely. Using two shields would stack.

 It's a nice secret item, could worth farming for, and wouldn't be OP if the perma-increase bug is removed. To make it legit part of the game, it would need a Ufopaedia entry. I would suggest it behaves like an alien artifact, including requiring research.

[settings.zip](https://github.com/user-attachments/files/25737820/settings.zip)

andremmfaria:
Fixes #1583.

## Current behaviour

The Mind Shield is inert. `BattleUnit::useItem()` and the `BattleView::orderUse()` click dispatch both list `MindShield` in their "items that do nothing" case groups, so clicking it in battle is a silent no-op (`useItem` returns `false`). The psi attack chance reads `modified_stats.psi_defence`, which nothing ever updates for this item. In the original game activating it granted +20 psi defence, with a bug that made the bonus permanent after the mission; this is the missing feature without that bug.

## Change

- `Agent::updatePsiDefence()` (new, mirrors `updateSpeed()`): recomputes `modified_stats.psi_defence = current_stats.psi_defence + 20` per equipped Mind Shield with `inUse` set. It always derives from `current_stats` and never mutates it, so unequipping or dropping an active shield clears the bonus and nothing can carry over between missions. The constant `MIND_SHIELD_PSI_DEFENCE_BONUS` sits next to `TELEPORT_TICKS_REQUIRED_AGENT` in `agent.h`.
- Called from every `updateSpeed()` site: `AgentGenerator::createAgent`, `Agent::addEquipment`, `Agent::removeEquipment`, `Agent::updateModifiedStats`, `AEquipment::fire`, plus the new `useItem` case.
- `BattleUnit::useItem`: `MindShield` toggles `item->inUse` like `MotionScanner`, then calls `updatePsiDefence()`.
- `battleview.cpp`: `MindShield` moved to the "usable items with no automatic mode" group so a click reaches `useItem`.
- No load-time hook: `inUse` and `modified_stats` already round-trip through `gamestate_serialize.xml`, so a save made with the shield active restores consistently.

## Open questions

- TU cost for the toggle is currently zero. No original-game source for a cost was found. `MotionScanner` charges 10% of `initialTU` in turn-based mode; happy to match that if preferred.
- The psi-attack consumption path was not exercised live in the harness (needs a MindBender attacker with psi energy and line of sight). Every psi-chance call site reads `modified_stats.psi_defence` directly, so the stat change is what matters.
- Two shields stack by construction (`updatePsiDefence()` counts all in-use shields), matching the reporter's description, but this was not separately tested.

## Verification

- Headless harness (difficulty0 gamestate, base defence battle, shield equipped in the right hand of a player soldier):
  - `3b2b59fe` (master): `useItem(MindShield)` returns `false`, `psi_defence` unchanged.
  - this branch: baseline 28, after activation 48, after deactivation 28, after removing an in-use shield 28 (no leak).
- Full ctest suite green (RelWithDebInfo, Linux); fork CI Lint + CMake green.

### DIFF
--- a/game/state/battle/battleunit.cpp
+++ b/game/state/battle/battleunit.cpp
@@ -5132,7 +5132,6 @@ bool BattleUnit::useItem(GameState &state, sp<AEquipment> item)
 		case AEquipmentType::Type::DimensionForceField:
 		case AEquipmentType::Type::DisruptorShield:
 		case AEquipmentType::Type::Loot:
-		case AEquipmentType::Type::MindShield:
 		case AEquipmentType::Type::MultiTracker:
 		case AEquipmentType::Type::StructureProbe:
 		case AEquipmentType::Type::VortexAnalyzer:
@@ -5164,6 +5163,10 @@ bool BattleUnit::useItem(GameState &state, sp<AEquipment> item)
 			// Initial use of medikit just brings up interface, action and TU spent happens
 			// when individual body part is clicked
 			item->inUse = !item->inUse;
+			return true;
+		case AEquipmentType::Type::MindShield:
+			item->inUse = !item->inUse;
+			agent->updatePsiDefence();
 			return true;
 		case AEquipmentType::Type::Brainsucker:
 			useBrainsucker(state);
--- a/game/state/shared/aequipment.cpp
+++ b/game/state/shared/aequipment.cpp
@@ -841,6 +841,7 @@ void AEquipment::fire(GameState &state, Vec3<float> targetPosition, StateRef<Bat
 		payloadType.clear();
 		loadAmmo(state);
 		ownerAgent->updateSpeed();
+		ownerAgent->updatePsiDefence();
 	}
 }
 
--- a/game/state/shared/agent.cpp
+++ b/game/state/shared/agent.cpp
@@ -187,6 +187,7 @@ StateRef<Agent> AgentGenerator::createAgent(GameState &state, StateRef<Organisat
 	}
 
 	agent->updateSpeed();
+	agent->updatePsiDefence();
 	agent->modified_stats.restoreTU();
 
 	return {&state, ID};
@@ -780,6 +781,7 @@ void Agent::addEquipment(GameState &state, Vec2<int> pos, sp<AEquipment> object)
 	}
 	this->equipment.emplace_back(object);
 	updateSpeed();
+	updatePsiDefence();
 	updateIsBrainsucker();
 	if (unit)
 	{
@@ -810,6 +812,7 @@ void Agent::removeEquipment(GameState &state, sp<AEquipment> object)
 	}
 	object->ownerAgent.clear();
 	updateSpeed();
+	updatePsiDefence();
 	updateIsBrainsucker();
 }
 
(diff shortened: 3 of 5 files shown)


andremmfaria:

```cpp
// tests/test_mind_shield.cpp (core), run against difficulty0 + base_gamestate.
// Setup: loadGame(common) + loadGame(gamestate), startGame(), initState(),
// fillPlayerStartingProperty(), then Battle::beginBattle(...) on the first
// player base's building (base defence), then:
bool test_mind_shield_toggle(sp<GameState> state)
{
	auto unit = findPlayerSoldier(state); // first player-owned Soldier unit
	auto mindShieldType = findMindShieldType(state); // AEquipmentType::Type::MindShield

	if (unit->agent->rightHandItem)
		unit->agent->removeEquipment(*state, unit->agent->rightHandItem);

	auto shield = unit->agent->addEquipmentByType(*state, mindShieldType,
	                                              EquipmentSlotType::RightHand, true);
	if (!shield)
		return false;

	int baseline = unit->agent->modified_stats.psi_defence;
	LogWarning("Baseline psi_defence: {0}", baseline);

	if (!unit->useItem(*state, shield))
	{
		LogError("useItem(MindShield) returned false - activation is a no-op");
		return false;
	}
	if (!shield->inUse || unit->agent->modified_stats.psi_defence != baseline + 20)
		return false;
	LogWarning("psi_defence after activation: {0}", unit->agent->modified_stats.psi_defence);

	if (!unit->useItem(*state, shield))
		return false;
	if (shield->inUse || unit->agent->modified_stats.psi_defence != baseline)
		return false;
	LogWarning("psi_defence after deactivation: {0}", unit->agent->modified_stats.psi_defence);

	// Re-activate, then remove the in-use shield: bonus must not leak.
	if (!unit->useItem(*state, shield) ||
	    unit->agent->modified_stats.psi_defence != baseline + 20)
		return false;
	unit->agent->removeEquipment(*state, shield);
	if (unit->agent->modified_stats.psi_defence != baseline)
		return false;
	LogWarning("psi_defence after removing in-use shield: {0}",
	           unit->agent->modified_stats.psi_defence);
	return true;
}
```

Observed output (baseline varies per run since the stat is rolled at agent creation; the deltas were identical across runs):

```text
# master 3b2b59fe
W ... Baseline psi_defence: 18
E ... useItem(MindShield) returned false - activation is a no-op
E ... test_mind_shield_toggle FAILED
(exit 1)

# this branch
W ... Baseline psi_defence: 28
W ... psi_defence after activation: 48
W ... psi_defence after deactivation: 28
W ... psi_defence after removing in-use shield: 28
W ... test_mind_shield success - all tests passed
(exit 0)
```

FilmBoy84:
Very nice! Especially with the fix for the permanent bonus in the OG
The fact that these were "cut" but still available to collect from MarSec missions meant that they often caused confusion
Thanks for resolving this

PR #1662, opened 2026-09-06
Battle: handle two brainsuckers attaching to one unit (#1221)

### LINKED ISSUE #1221: Two Brainsuckers simultaneously trying to "Brainsuck" the same soldier will cause a crash
- OS: Windows 11 Pro 22H2
- OpenApoc Version: 0.0.737
- List of Mod files in use: None

It crashes when two Brainsuckers try to "Brainsuck" the same soldier/agent. It only seems to work if the first Brainsucker successfully Brainsucks the soldier/agent and turns him into an enemy unit, then the second Brainsucker will cause the crash.

**Expected behavior**
If two Brainsuckers decide to jump on the same agent/soldier, then if the first succeeds, then the second one should bounce back to the ground to search for another victim to "Brainsuck".

**Savegame**
Just End the turn when you have loaded this save file. It will hopefully show you the same results.
https://www.dropbox.com/s/vskf5z7yz19y0rf/save_Brainsucker%20bug.save?dl=0

**Screenshots**
Brainsucker Bug

andremmfaria:
Fixes #1221.

## Root cause

`BattleUnit::updateFallingIntoUnit` (battleunit.cpp:2931) has an inverted eligibility check: a falling Brainsucker only bounces off an already-claimed target when the damage roll also fails (`!unit.brainSucker && dealDamage(...) == 0`), so a second lander always attempts to attach to a unit that already has one attached.

The `Brainsuck` guard in `BattleUnitMission::start` (battleunitmission.cpp:1862) then rejects the second attacher with `cancelled = true` but without the `resetGoal()` / `startFalling()` cleanup that `update()`'s cancel path performs for the same mission type (battleunitmission.cpp:1657-1663). The rejected unit is left hovering with a cancelled mission instead of falling, which matches the "hovers in place forever" observation in the issue thread.

Neither `BattleUnit::useBrainsucker()` nor `UnitAIVanilla::getBrainsuckerDecision()` checks `target->brainSucker`, so two AI Brainsuckers converge on the same nearest victim by design. The reporter's save shows exactly that: three Brainsuckers pick the same soldier and two issue `Brainsuck` starts on it.

## Fix

Three lines in two files:

* `updateFallingIntoUnit`: bounce/stun whenever the target is already claimed, independent of the damage roll (`unit.brainSucker || dealDamage(...) == 0`).
* `BattleUnitMission::start`: the rejection now calls `u.resetGoal(); u.startFalling(state);` alongside `cancelled = true`, symmetric with the `update()` cancel path.

## Verification

The reported crash was not reproduced on Linux: not in RT or TurnBased, not via the direct mission API or a real fall collision, not under ASan+UBSan, and not with the reporter's own save. That save loads and replays the scenario (three Brainsuckers converge on one soldier, two Brainsuck starts on it) through 50 tick blocks with repeated end turn, with no crash before or after this change.

What does change: a headless harness asserting that a rejected second attacher falls (or never queues a Brainsuck mission at all) instead of hovering with a cancelled mission fails on master `3b2b59fe` and passes on this branch, under both plain RelWithDebInfo and ASan+UBSan builds. Full ctest green (Linux); fork CI Lint + CMake green.

@XCOM-HUB, if you still have the setup, a retest on a build with this change would help confirm it covers the crash seen on Windows.

### DIFF
--- a/game/state/battle/battleunit.cpp
+++ b/game/state/battle/battleunit.cpp
@@ -2928,7 +2928,7 @@ void BattleUnit::updateFallingIntoUnit(GameState &state, BattleUnit &unit)
 			if (position.z < unit.getMuzzleLocation().z)
 			{
 				StateRef<DamageType> brainsucker = {&state, "DAMAGETYPE_BRAINSUCKER"};
-				if (!unit.brainSucker &&
+				if (unit.brainSucker ||
 				    brainsucker->dealDamage(100, unit.agent->type->damage_modifier) == 0)
 				{
 					// Cannot suck this head, get stunned
--- a/game/state/battle/battleunitmission.cpp
+++ b/game/state/battle/battleunitmission.cpp
@@ -1862,6 +1862,8 @@ void BattleUnitMission::start(GameState &state, BattleUnit &u)
 			if (targetUnit->brainSucker && targetUnit->brainSucker.id != u.id)
 			{
 				cancelled = true;
+				u.resetGoal();
+				u.startFalling(state);
 				return;
 			}
 			targetUnit->brainSucker = {&state, u.id};


andremmfaria:

```cpp
// Base-defense battle with three Brainsuckers invading the player's base
// (difficulty0, same idiom as the PR 1649 harness). fail(msg) = LogError + return EXIT_FAILURE.
StateRef<AgentType> brainsuckerType(state.get(), "AGENTTYPE_BRAINSUCKER");
building->current_crew[brainsuckerType] = 3;
Battle::beginBattle(*state, false, state->getAliens(), playerAgents, nullptr, nullptr,
                    nullptr, StateRef<Vehicle>{}, building);
Battle::enterBattle(*state);
// suckers = the three AGENTTYPE_BRAINSUCKER units; soldier, soldier2 = two player soldiers

// (a) two Brainsuck missions on the same target in the same tick (BattleUnitMission::start guard)
StateRef<BattleUnit> soldierRef(state.get(), soldier->id);
auto A1 = suckers[0], A2 = suckers[1];
A1->setMission(*state, BattleUnitMission::brainsuck(*A1, soldierRef, 0));
A2->setMission(*state, BattleUnitMission::brainsuck(*A2, soldierRef, 0));
LogWarning("A: soldier.brainSucker={0} A2.missions={1} A2.cancelled={2} A2.falling={3}", /* ... */);
if (!soldier->brainSucker || soldier->brainSucker.id != A1->id)
    fail("first attacher A1 should hold soldier.brainSucker");
if (A2->missions.empty() || !A2->missions.front()->cancelled)
    fail("rejected attacher A2's mission should be cancelled");
if (!A2->falling)
    fail("rejected attacher A2 should start falling on rejection, not hover with a cancelled mission");
for (int i = 0; i < 200; i++)
    state->current_battle->update(*state, TICKS_PER_SECOND / 4);

// (b) second sucker lands on an already-claimed target via updateFallingIntoUnit
StateRef<BattleUnit> soldier2Ref(state.get(), soldier2->id);
auto B1 = suckers[2], B2 = A1;
B1->setMission(*state, BattleUnitMission::brainsuck(*B1, soldier2Ref, 0));
auto pos = soldier2->position;
pos.z += 0.3f;
B2->velocity = {0.0f, 0.0f, -1.0f};
B2->falling = true;
B2->setPosition(*state, pos, true);
B2->updateFallingIntoUnit(*state, *soldier2);
LogWarning("B: soldier2.brainSucker={0} B2.missions={1} B2.front.cancelled={2}", /* ... */);
if (!soldier2->brainSucker || soldier2->brainSucker.id != B1->id)
    fail("first attacher B1 should still hold soldier2.brainSucker");
if (!B2->missions.empty() && B2->missions.front()->cancelled)
    fail("rejected lander B2 should take the stun/bounce path (no Brainsuck mission queued), not attach-then-cancel");
for (int i = 0; i < 200; i++)
    state->current_battle->update(*state, TICKS_PER_SECOND / 4);
```

Output on master `3b2b59fe` (abridged, unit ids omitted):
```
W A: soldier.brainSucker=A1 A2.missions=1 A2.cancelled=true A2.falling=false
E Invariant violated: rejected attacher A2 should start falling on rejection, not hover with a cancelled mission
```
Scenario (b) on master, from a run without the (a) assertion: `B2.missions=1 B2.front.cancelled=true`, followed by `E ... setMovementState 1 incompatible with current body states 7 to 7`.

Output on this branch:
```
W A: soldier.brainSucker=A1 A2.missions=1 A2.cancelled=true A2.falling=true
W Scenario (a) survived 7200 ticks of battle update without crashing
W B: soldier2.brainSucker=B1 B2.missions=0 B2.front.cancelled=false
W Scenario (b) survived 7200 ticks of battle update without crashing
```
Same results under an ASan+UBSan build, with no sanitizer findings on either side. A TurnBased variant (same two-mission setup, 30 end turns) fails and passes the same way.

XCOM-HUB:
@andremmfaria 

I don't have the setup, and the save files are too old for the latest AppVeyor Build (0.0.1521). I trust that this bug is fixed. I will keep my eyes open for when multiple Brainsuckers jump one soldier in the future.

Issue #987, opened 2021-02-08
Crush When End my turn


endy2qq:
Hello.
Please check my save file.

Thank you

[save_psimorph crush turn end.zip](https://github.com/OpenApoc/OpenApoc/files/5943663/save_psimorph.crush.turn.end.zip)

PR #1669, opened 2026-09-08
Battle: keep large and prone unit footprints on the map (#1007)

### LINKED ISSUE #787: CTD on END TURN - "Incorrect tile coordinates"
Suddenly on END TURN button click rises this error message:

Screenshot from 2019-10-20 17-55-23

gdb logs: [gdb.txt](https://github.com/OpenApoc/OpenApoc/files/3747984/gdb.txt)
game logs (150 MB): [log.txt.zip](https://github.com/OpenApoc/OpenApoc/files/3747987/log.txt.zip)

See #1007 and #1057

### LINKED ISSUE #1007: Multi-Tile Units in Battlescape can spawn, or move part of their form, outside of the map limits causing a "Incorrect Tile Coordinates" CTD
Still currently on a previous build (998) so that you know it isn't related to most recent merging. Basically I saved the game and got the error below with a CTD. Reloaded the game and it loaded in fine, then went to end the turn and it immediately CTD'd again with the same error message.

Capture

andremmfaria:
Fixes #1007.
Fixes #787.

## Root cause

Nothing enforces the invariant "a unit's whole footprint is on the map". A large unit occupies `{x-1..x} x {y-1..y} x {z..z+1}` (`tileobject_battleunit.cpp:279-291`) and a prone unit additionally occupies the legs tile behind it, but `TileObject::setPosition` (`tileobject.cpp:99-141`) only clamps the centre and silently drops footprint tiles that fall off the map. Readers that later walk that footprint then dereference `TileMap::getTile`, which returns `nullptr` and logs `Incorrect tile coordinates ...` for out-of-range coordinates (`tilemap.h:100-118`), without a null check:

* `BattleUnitTileHelper::canEnterTile` guards the TO side (`battleunitmission.cpp:226`) but not the FROM side (`:231-236`). This is the `Incorrect tile coordinates 3,-1,0` crash in the issue, and the mid-mission crash after `findShortestPath`.
* `BattleUnit::updateIdling`'s prone leg-support loop (`battleunit.cpp:2599-2606`) for a prone unit whose legs tile (`pos - facing`) sits off the map. That is the #787 backtrace.
* `Battle::initialMapPartRemoval` (`battle.cpp:435`) dereferencing `getTile(position + {0,0,1})` for a large unit standing at the top z level.

And three places that set a large unit's position never clamped the centre so the footprint stays on the map: `Battle::placeUnit(state, agent, position)`, the `spawnLocations` branch of `Battle::initialUnitSpawn`, and the edge/ceiling clamp in `BattleUnit::updateMovementFalling`. The LOS-block spawn, teleport, give-way and AI random-move paths were already guarded.

## Fix

Four surgical changes in `game/`, no change to `TileMap::getTile`/`isTileInBounds` semantics:

* `canEnterTile` now also rejects `fromPos.x < 1 || fromPos.y < 1` for a large unit, mirroring the existing TO-side fit check (returns `false` silently, same as that check).
* `updateIdling`'s prone-support loop gates the `getTile` call on `isTileInBounds` and treats an out-of-bounds tile as "no support", so the unit kneels up instead of crashing. Same guard style as `updateGiveWay` (`battleunit.cpp:2455`).
* `initialMapPartRemoval` null-checks the tile before reading `solidGround`.
* The three unguarded placement paths clamp a large unit's centre to `x >= 1`, `y >= 1`, `z <= size.z - 2`, so the invariant also holds at the source.

`initialMapPartRemoval` still only checks one of the four footprint tiles; that is pre-existing behaviour, unrelated to the crash, and left alone here.

## Verification

Headless harness (`test_map_bounds`, removed after validation) that builds a real `Battle` via `Battle::beginBattle`/`enterBattle`, places one unit with `Battle::placeUnit` and scripts each edge condition:

|---|---|---|
| a: large unit, from-side footprint off-map, `canEnterTile` | SIGSEGV at `battleunitmission.cpp:232`, after `Incorrect tile coordinates -1,30,0` | returns `false` |
| b: prone unit, legs tile off-map, `updateIdling` (#787) | SIGSEGV in `Tile::getCanStand(this=0x0)` at `tile.cpp:115`, from `battleunit.cpp:2601` | survives, tile treated as unsupported |
| c: large unit at top z, `initialMapPartRemoval` | SIGSEGV on `t->solidGround` at `battle.cpp:436` | survives |
| d: control, large unit fully in bounds | survives | survives |

Every pre-fix backtrace landed on the predicted line (`gdb -batch -ex run -ex bt`). Full ctest green (RelWithDebInfo, Linux); fork CI Lint + CMake green.

Related: #1003 is already closed as a duplicate of #1007. #987 (END TURN CTD on a Battleship) matches the same trigger pattern but was not reproduced, so it is not claimed as fixed here.

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -433,7 +433,7 @@ void Battle::initialMapPartRemoval(GameState &state)
 			for (int y = 0; y < 1; y++)
 			{
 				auto t = map->getTile(u.second->position + Vec3<float>{-x, -y, 1.0f});
-				if (t->solidGround)
+				if (t && t->solidGround)
 				{
 					std::list<sp<TileObjectBattleMapPart>> partsToKill;
 					for (auto &o : t->ownedObjects)
@@ -721,6 +721,12 @@ void Battle::initialUnitSpawn(GameState &state)
 		}
 		auto pos = pickRandom(state.rng, spawnLocations[u.second->agent->type]);
 		spawnLocations[u.second->agent->type].remove(pos);
+		if (u.second->isLarge())
+		{
+			pos.x = std::max(pos.x, 1);
+			pos.y = std::max(pos.y, 1);
+			pos.z = std::min(pos.z, map->size.z - 2);
+		}
 		auto tile = map->getTile(pos.x, pos.y, pos.z);
 		u.second->position = tile->getRestingPosition(u.second->isLarge());
 	}
@@ -1220,6 +1226,12 @@ sp<BattleUnit> Battle::placeUnit(GameState &state, StateRef<Agent> agent)
 sp<BattleUnit> Battle::placeUnit(GameState &state, StateRef<Agent> agent, Vec3<float> position)
 {
 	auto u = placeUnit(state, agent);
+	if (map && u->isLarge())
+	{
+		position.x = std::max(position.x, 1.0f);
+		position.y = std::max(position.y, 1.0f);
+		position.z = std::min(position.z, (float)map->size.z - 2.0f);
+	}
 	u->position = position;
 	if (map)
 	{
--- a/game/state/battle/battleunit.cpp
+++ b/game/state/battle/battleunit.cpp
@@ -2598,7 +2598,9 @@ void BattleUnit::updateIdling(GameState &state)
 				bool hasSupport = true;
 				for (auto &t : tileObject->occupiedTiles)
 				{
-					if (!tileObject->map.getTile(t)->getCanStand())
+					auto *supportTile =
+					    tileObject->map.isTileInBounds(t) ? tileObject->map.getTile(t) : nullptr;
+					if (!supportTile || !supportTile->getCanStand())
 					{
 						hasSupport = false;
 						break;
@@ -2853,6 +2855,12 @@ void BattleUnit::updateMovementFalling(GameState &state, unsigned int &moveTicks
 			newPosition.x = glm::clamp(newPosition.x, 0.0f, mapSize.x - 0.01f);
 			newPosition.y = glm::clamp(newPosition.y, 0.0f, mapSize.y - 0.01f);
 			newPosition.z = glm::clamp(newPosition.z, 0.0f, mapSize.z - 0.01f);
+		}
+		if (isLarge())
+		{
+			newPosition.x = glm::clamp(newPosition.x, 1.0f, mapSize.x - 0.01f);
+			newPosition.y = glm::clamp(newPosition.y, 1.0f, mapSize.y - 0.01f);
+			newPosition.z = glm::clamp(newPosition.z, 0.0f, mapSize.z - 1.01f);
 		}
 		// Fell below 0???
 		if (newPosition.z < 0 && !tileObject->map.getTile(newPosition))
--- a/game/state/battle/battleunitmission.cpp
+++ b/game/state/battle/battleunitmission.cpp
@@ -223,7 +223,8 @@ bool BattleUnitTileHelper::canEnterTile(Tile *from, Tile *to, bool allowJumping,
 	if (large)
 	{
 		// Can we fit?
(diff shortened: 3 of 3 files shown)


andremmfaria:

```cpp
// Real Battle from a UFO recovery mission against the first vehicle type with a battle_map,
// seeded with one alien so fillSquads() has a participant (same idiom as skirmish.cpp's
// loadBattleVehicle). Each case then places its own unit and pokes position/body state
// directly to script the edge condition. findAgentType(state, large) looks the type up by the
// bodyType flag rather than by ruleset name.
Battle::beginBattle(state, false, state.getAliens(), playerAgents, &aliens, {}, targetCraft);
Battle::enterBattle(state);
auto unit = state.current_battle->placeUnit(
    state, state.agent_generator.createAgent(state, state.getAliens(), type), position);

// (a) large unit whose FROM footprint is off the map (x == 0)
unit->setPosition(state, {0.0f, (float)y0, 0.0f});
BattleUnitTileHelper helper(map, *unit);
bool result = helper.canEnterTile(map.getTile(0, y0, 0), map.getTile(1, y0, 0));
LogWarning("Case a: canEnterTile from off-map large footprint returned {0}", result);

// (b) #787: small prone unit whose legs tile (pos - facing) is off the map (y == size.y)
unit->facing = {0, -1};
unit->goalFacing = unit->facing;
unit->current_body_state = unit->target_body_state = BodyState::Prone;
unit->movement_mode = MovementMode::Prone;
unit->atGoal = true;
unit->goalPosition = unit->position;
unit->setPosition(state, unit->position); // rebuild occupiedTiles under Prone + facing
unit->updateIdling(state);
LogWarning("Case b: survived updateIdling with off-map legs tile, queued kneel mission={0}",
           !unit->missions.empty());

// (c) large unit at the top z level, footprint tile at z + 1 is off the map
unit->setPosition(state, {(float)x0, (float)y0, (float)(map.size.z - 1)});
battle->initialMapPartRemoval(state);

// (d) control: large unit fully in bounds at x == 1, same canEnterTile call as (a)
```

Output on master `9804cfd9` (`gdb -batch -ex run -ex bt`, one case per run, abridged):
```
E TileMap::getTile(int, int, int) Incorrect tile coordinates -1,30,0
Thread 1 "test_map_bounds" received signal SIGSEGV, Segmentation fault.
#0  BattleUnitTileHelper::canEnterTile (...) at game/state/battle/battleunitmission.cpp:232
#2  runCaseA (state=...) at tests/test_map_bounds.cpp:143

E TileMap::getTile(int, int, int) Incorrect tile coordinates 30,60,0
Thread 1 "test_map_bounds" received signal SIGSEGV, Segmentation fault.
#0  Tile::getCanStand (this=0x0, large=false) at game/state/tilemap/tile.cpp:115
#1  BattleUnit::updateIdling (...) at game/state/battle/battleunit.cpp:2601

E TileMap::getTile(int, int, int) Incorrect tile coordinates 30,30,10
Thread 1 "test_map_bounds" received signal SIGSEGV, Segmentation fault.
#0  Battle::initialMapPartRemoval (...) at game/state/battle/battle.cpp:436

W Case d: control canEnterTile (in-bounds) returned false
W PASS case_d_control_in_bounds expected=survive
```

Output on this branch (all four cases in one process, exit 0):
```
W Case a: canEnterTile from off-map large footprint returned false
W PASS case_a_large_from_edge expected=survive
W Case b: survived updateIdling with off-map legs tile, queued kneel mission=true
W PASS case_b_prone_legs_edge expected=survive
W Case c: survived initialMapPartRemoval with large unit at top z level
W PASS case_c_top_z_edge expected=survive
W Case d: control canEnterTile (in-bounds) returned false
W PASS case_d_control_in_bounds expected=survive
```

The control behaves identically on both builds, so the added from-side check is a no-op in bounds. `ctest` with the harness in the tree: 12/12.

PR #1664, opened 2026-09-06
Battle: land falling units and items instead of killing them below z=0 (#844)

andremmfaria:
Fixes #844.

## Root cause

`BattleUnit::updateMovementFalling()` and the equivalent block in `BattleItem::update()` integrate a whole batch of ticks of gravity into `newPosition` in one step, more at higher game speed. Before the "reached ground" logic (resting-position snap for units, `findSupport()`/`getSupport()` for items) runs, both check `newPosition.z < 0` and, if true, log `... fell off the end of the world!?` and kill the unit or item.

A unit standing on a z=0 floor tile that starts falling and takes one coarse step overshoots to a small negative z within the same tile. The issue's debug output is exactly this: previous z `0.1218`, new z `-0.0286`, resting height `0.075`. The tile still exists; the check misreads a sub-tile overshoot as "no floor here". #581 noted in 2019 it only reproduced at medium or high speed. PR #1569 clamped x/y at the map edge but left the vertical `z < 0` kill in both files.

## Fix

Gate the kill on tile existence at the overshoot spot via `map.getTile(newPosition)`. `TileMap::getTile(Vec3<float>)` truncates toward zero, so a shallow negative z resolves to the z=0 tile. If it exists, the overshoot is left in place: units land through the existing `position.z < restingPosition.z` snap, and items set `collision = true` so `findSupport()`/`getSupport()` settle them. Death only happens when no tile exists, i.e. the overshoot exceeds a full tile.

## Verification

* Headless harness on a real base-defense battle (difficulty0, fixed RNG seed), three scenarios:
  * (a) standing on the z=0 tile, `startFalling()`, one `battle->update(state, 36)` step; must land.
  * (b) established fall (z velocity -50, 8 ticks); must land at the resting height, not tunnel through.
  * (c) overshoot beyond a full tile, no tile to land on; must still die with the original log line.
* master `3b2b59fe`: (a) fails with `Unit at 69.5 88.5 fell off the end of the world!?`, unit dies.
* This branch: all three pass across three runs.
* Full ctest suite green (RelWithDebInfo, Linux); fork CI Lint + CMake green.

Separately, the "Fell into a unit" check in the same function dereferences `map.getTile(newPosition)` without a null check, so a conscious unit overshooting by more than a full tile segfaults there. Pre-existing on master, left out of scope.

The harness core is posted as a comment below.

Issue #1330, opened 2024-02-05
Shooting between adjacent grav-lift tiles is impossible


dethstryke01:
[https://youtu.be/Lgxlf0U-09I?t=1920](url)

Window 10 x64 22H2
OpenApoc version 0.0.930
no mods in use

[settings.zip](https://github.com/OpenApoc/OpenApoc/files/14166034/settings.zip)

[save_test 7.zip](https://github.com/OpenApoc/OpenApoc/files/14166035/save_test.7.zip)

kgd192:
@dethstryke01 Unfortunately your Savefiles do not allow for an easy debug, we need you to make the save at the moment the issue is occuring and not at the End of the Session. So if you encounter en error like this make a separate safe file for each issue and send it to us so we can see/debug the exact situation you are having.

Cheers, Kai

dethstryke01:
Gotcha. I'll make sure to make a save when each bug is occurring when I'm playtesting in the future.

kgd192:
Perfect, thank you. Otherwise keep up the videos and issue reports they are a massive help 👌

dethstryke01:
Glad to hear it. It's been a lot of fun playtesting so far :)

dethstryke01:
I found the same grav-lift on the same map and managed to get the same "fell off the end of the world!" error by throwing grenades downwards. Some tiles in that area seem to generate the error while others don't. I made save files during a buggy state the happened after the error message where many of the animations in the game seemed to be running slightly faster for some reason. 

https://youtu.be/-029oEE0ynE?t=4268

[save_35.zip](https://github.com/OpenApoc/OpenApoc/files/14237843/save_35.zip)
[save_grenade fell off end of world.zip](https://github.com/OpenApoc/OpenApoc/files/14237844/save_grenade.fell.off.end.of.world.zip)
[save_2nd grenade fell off.zip](https://github.com/OpenApoc/OpenApoc/files/14237845/save_2nd.grenade.fell.off.zip)

dethstryke01:
Here are three save files from situations where shooting at enemies between tiles in grav-lifts was impossible for some reason.

[shooting in lift doesnt work.zip](https://github.com/OpenApoc/OpenApoc/files/14965538/shooting.in.lift.doesnt.work.zip)
[lift shooting problem 2.zip](https://github.com/OpenApoc/OpenApoc/files/14965539/lift.shooting.problem.2.zip)
[...]

PR #1670, opened 2026-09-08
Battle: stop falling items dropping through the floor they hit (#1057)

### LINKED ISSUE #1057: Agents can throw grenades through the floor - projectile collision and movement bugs
**Describe the bug**
When an agent throws a grenade from height down onto an adjacent square the grenade will go through the ground.
Only seems to activate for adjacent squares.
Possibly related to #844

**To Reproduce**
1. Start a new game and go to skirmish mode
2. Take an agent up a few height levels (must be about 3-4 levels above ground to activate)
3. Move camera down to ground level
4. Throw a grenade down onto an adjacent square

**Expected behavior**
Grenade lands or bounces

**Screenshots**
Agent throwing grenade

 - OS: Ubuntu 18.04.3 LTS
 - OpenApoc Version: Built from master at commit 3624277

andremmfaria:
Fixes #1057.
Refs #1330.

## Root cause

`BattleItem::update` integrates up to four ticks of gravity into one `newPosition` and then does a single collision check for the whole displacement (`battleitem.cpp:206-220`). The Ground branch of that check throws the collision point away:

```cpp
case TileObject::Type::Ground:
    newPosition = {previousPosition.x, previousPosition.y,
                   std::min(newPosition.z, previousPosition.z)};
    break;
```

It keeps the full pre-collision displacement, clamped only against where the item started this update, and never uses `c.position`, which `findCollision` already set to the voxel-precise collision point in the same tile-space units (`collision.cpp:108-109`, `:164-165`).

A floor's Ground voxels occupy roughly `[k, k+0.1)` of tile level k, so once an item has fallen about 1.6 tiles a single update step crosses the whole band and ends below the floor's tile row. Landing then depends entirely on `findSupport()`, which looks at the owning tile of the already-moved item, i.e. the tile below the floor it just tunnelled through, finds no support, and lets it keep falling. This matches the "must be 3-4 levels up" and "two of three grenades" reports on the issue.

PR #1664's `newPosition.z < 0` gate only protects level 0; an overshoot below a floor at z >= 1 lands on a positive z and never reaches it.

## Fix

One line in the Ground branch: floor the result at the collision surface as well.

```cpp
newPosition = {previousPosition.x, previousPosition.y,
               std::max(std::min(newPosition.z, previousPosition.z), c.position.z)};
```

The existing `min` against `previousPosition.z` stays (`newPosition` may already have been reset by a bounce earlier in the same call), and the item can no longer end below the voxel it just hit. `newPosition`'s owning tile is then the floor's own tile, so `findSupport()`/`getSupport()` snap the item to the resting height as designed. x/y handling, the throw trajectory code and unit falling are untouched. PR #1664's z-below-zero gate is kept as the genuine "no floor anywhere" fallback for items that never register a Ground collision at all.

## Verification

Headless harness (`test_item_fall`, removed after validation): boots a base defence battle from the difficulty0 state, scans the generated map for a column with a floor at a target z and clear tiles above it, drops an AP grenade with `falling = true`, ticks `battle->update(state, 4)` until it settles, and asserts it lands within 0.05 tile of that floor's resting height.

Pre-fix (`9804cfd9`, `battleitem.cpp` restored):

```
PASS item_fall_z0 expected=0.075 got=0.075
FAIL item_fall_z1 expected=1.95 got=0.975
```

This branch:

```
PASS item_fall_z0 expected=0.2 got=0.2
PASS item_fall_z1 expected=1.55 got=1.55
```

The z=0 control passes both ways, confirming PR #1664 is unaffected and isolating the regression to floors at z >= 1. The map is regenerated per run, so the coordinates and heights differ between runs; each case compares its own landing z against its own floor. Full ctest green (RelWithDebInfo, Linux); fork CI Lint + CMake green.

This closes sub-problem C of #1330; A and B are unrelated code paths and stay open. Unit falling has the same coarse-step problem in `battleunit.cpp` and is handled separately.

### DIFF
--- a/game/state/battle/battleitem.cpp
+++ b/game/state/battle/battleitem.cpp
@@ -242,9 +242,9 @@ void BattleItem::update(GameState &state, unsigned int ticks)
 			// Intentional fall-through
 			case TileObject::Type::Ground:
 				// Let item fall so that it can collide with scenery or ground if falling on top of
-				// it
+				// it, but never past the surface we just hit
 				newPosition = {previousPosition.x, previousPosition.y,
-				               std::min(newPosition.z, previousPosition.z)};
+				               std::max(std::min(newPosition.z, previousPosition.z), c.position.z)};
 				break;
 			default:
 				LogError("What the hell is this item colliding with? Type is {0}",


andremmfaria:

```cpp
// tests/test_item_fall.cpp (core; difficulty0 loading boilerplate omitted).
// Base defence battle on the player's own base, fixed RNG seed.
state->rng.seed(1057);
state->startGame();
state->initState();
state->fillPlayerStartingProperty();

StateRef<Building> building = state->player_bases.begin()->second->building;
std::list<StateRef<Agent>> playerAgents;
Battle::beginBattle(*state, false, state->getAliens(), playerAgents, nullptr, nullptr,
                    nullptr, StateRef<Vehicle>{}, building);
Battle::enterBattle(*state);

// Scan for a column with a floor at floorZ and clear tiles above it. The count is capped by
// map.size.z so the scan never queries an out-of-map tile.
int available = std::min(4, map.size.z - 1 - floorZ);
// ... find (x, y) with supportProviderForItems at floorZ and 'available' empty tiles above ...

auto grenadeType = findGrenadeType(state); // first AEquipmentType::Type::Grenade, no hardcoded id
auto eq = mksp<AEquipment>();
eq->type = grenadeType;
eq->ammo = grenadeType->max_ammo;
eq->ownerOrganisation = state->getPlayer();

auto item = state->current_battle->placeItem(
    *state, eq, {x + 0.5f, y + 0.5f, floorZ + available + 0.5f});
item->falling = true;
item->velocity = {0.0f, 0.0f, 0.0f};

for (int i = 0; i < 2000 && item->tileObject && item->falling; i++)
    state->current_battle->update(*state, 4); // Speed3 tick step

// Assert: item still alive, !item->falling, and position.z within 0.05 of
// map.getTile(x, y, floorZ)->getRestingPosition(false).z
```

Two cases: `item_fall_z0` (the level-0 control PR #1664 already covers) and `item_fall_z1` (the #1057 regression).

Output on master `9804cfd9` with `battleitem.cpp` reverted (exit 8):
```
Column for floor z=0: x=2 y=36 openAbove=4
PASS item_fall_z0 expected=0.075 got=0.075
Column for floor z=1: x=17 y=55 openAbove=3
FAIL item_fall_z1 expected=1.95 got=0.975
```

```
Column for floor z=0: x=22 y=103 openAbove=4
PASS item_fall_z0 expected=0.2 got=0.2
Column for floor z=1: x=37 y=37 openAbove=3
PASS item_fall_z1 expected=1.55 got=1.55
test_item_fall success - all tests passed
```

The grenade lands a full tile below its target floor pre-fix, on the floor underneath, which is the reported symptom. The level-0 control passes both ways, so PR #1664 is unaffected. The map is regenerated per process run, so the columns and heights differ between the two runs; each case compares its own landing z against its own floor. No `Incorrect tile coordinates` in either log. `ctest` with the harness in the tree: 12/12.

Issue #1203, opened 2023-04-26
Alien Containment Overflow addition


Nurulinde:
When returning from a mission where you obtained either Live or Dead aliens, the game will bring up an Alien Containment Overflow screen if you would overflow your Bases capacity.  It would be nice if there was some sort of indicator for each alien type whether or not you have researched it yet, to help you decide which aliens/bodies to keep or discard.

FilmBoy84:
I agree, this would be a good QoL feature

I will add the appropriate tags and request that whoever takes this on attempts to find an appropriate sprite from the original game that can be called, or an appropriate text string, before implementing a new one (that could create problems in of itself without careful design)

PR #1671, opened 2026-09-08
City: deliver combat loot to any owned base with containment (#1471)

### LINKED ISSUE #1471: Captured aliens from a mission can't be transported to a second base
- OS: Windows 11 Pro x64 23H2
 - OpenApoc Version: [0.0.1097](https://ci.appveyor.com/project/OpenApoc/openapoc/builds/49988825)
 - No Mods

SAVE-FILE: [Alien Containment in Base 2.zip](https://github.com/user-attachments/files/15806919/Alien.Containment.in.Base.2.zip)
"Base 1" vehicle (Valkyrie Interceptor 1) can't transport Aliens to "Base 2". Alien Containment only exists in "Base 2".

andremmfaria:
Fixes #1471.

## Root cause

`Battle::exitBattle` stamps every piece of recovered loot with the vehicle's home building as its `Cargo::destination` (`battle.cpp:3334-3335`). The only "we have arrived" test in the cityscape is exact destination equality (`vehicle.cpp:1742` in `getServiceDestination`, `vehicle.cpp:2410` in `updateCargo`), so a craft that lands at any other base is never considered a valid drop-off. The Alien Containment gate lives inside that branch, so it never even gets a chance to run at the second base, and the craft offers no ferry mission either.

## Fix

A single predicate, plus a rebind of the destination on landing:

```cpp
bool Vehicle::cargoDeliverableAtCurrentBuilding(const Cargo &c) const
{
	return currentBuilding && (c.destination == currentBuilding ||
	                           (!c.originalOwner && currentBuilding->base &&
	                            currentBuilding->owner == owner));
}
```

`getServiceDestination` uses it and sets `it->destination = currentBuilding` before the unchanged containment gate and `arrive()` call. The rebind matters: `Cargo::arrive` keys off `destination->base` (`vehicle.cpp:4040-4054`), so without it the loot would leave the vehicle and land in the wrong base's stock. `updateCargo` uses the same predicate when deciding whether a ferry mission is needed, with `c.originalOwner` kept as its own disjunct so transfers and purchases still force a ferry while in flight.

Only combat loot qualifies (`originalOwner == nullptr`), and only at a building that hosts a base owned by the vehicle's owner. Purchases and transfers keep the exact-destination path, and building freight (`Building::cargo`) is a different list that this change never touches.

Ownership is decided via `currentBuilding->owner == owner`, since `Base` has no owner field of its own; that matches the existing convention at `building.cpp:365`.

Loot is redirected to any owned base, not only ones with containment. That is deliberate: it re-triggers the existing "no containment here" dialog at the base the player actually flew to, instead of silently stranding the cargo. No capacity check is added, matching current `Cargo::arrive` behaviour (#1203 territory, out of scope).

## Verification

Headless harness (`test_alien_cargo`, removed after validation): difficulty0 state, a second base B, a Biotrans homed at base A landing at B with one live alien as combat loot.

|---|---|---|
| B has Alien Containment | no ferry mission, cargo kept aboard | unloaded into B, cargo empty |
| B has no Alien Containment | kept aboard | kept aboard (existing dialog) |
| transfer cargo addressed to A, landed at B | kept aboard | kept aboard |
| loot landed at a building without a player base | kept aboard | kept aboard |

Only the bug case flips; the three controls pass in both legs. Full ctest green (RelWithDebInfo, Linux); fork CI Lint + CMake green.

Two adjacent one-liners were found while sweeping the cargo consumers and deliberately left out of this diff: `battle.cpp:3354` uses `getMaxCargo()` where `getMaxBio()` looks intended, and `cityview.cpp:4024` is missing an iterator increment. Happy to file them separately.

### DIFF
--- a/game/state/city/vehicle.cpp
+++ b/game/state/city/vehicle.cpp
@@ -1739,8 +1739,9 @@ StateRef<Building> Vehicle::getServiceDestination(GameState &state)
 	// Step 01: Find first cargo destination and remove arrived cargo
 	for (auto it = cargo.begin(); it != cargo.end();)
 	{
-		if (it->destination == currentBuilding)
+		if (cargoDeliverableAtCurrentBuilding(*it))
 		{
+			it->destination = currentBuilding;
 			// Only add aliens if alien containment is available at base
 			const auto vehicleContainsAlienLoot = cargoContainsAlienLoot();
 			const auto alienContainmentExists =
@@ -2407,7 +2408,7 @@ void Vehicle::updateCargo(GameState &state)
 		// Either this is non-combat loot, or this loot belongs to this building
 		// Don't keep trying to ferry combat loot to other building as we're NOT going to move
 		// when we check it in getServiceDestination method!
-		if (c.originalOwner || c.destination == currentBuilding)
+		if (c.originalOwner || cargoDeliverableAtCurrentBuilding(c))
 		{
 			needFerry = true;
 			break;
@@ -3910,6 +3911,13 @@ const bool Vehicle::cargoContainsAlienLoot() const
 	}
 
 	return false;
+}
+
+bool Vehicle::cargoDeliverableAtCurrentBuilding(const Cargo &c) const
+{
+	return currentBuilding &&
+	       (c.destination == currentBuilding ||
+	        (!c.originalOwner && currentBuilding->base && currentBuilding->owner == owner));
 }
 
 Cargo::Cargo(GameState &state, StateRef<AEquipmentType> equipment, int count, int price,
--- a/game/state/city/vehicle.h
+++ b/game/state/city/vehicle.h
@@ -389,6 +389,7 @@ class Vehicle : public StateObject<Vehicle>,
 	const UString getFormattedVehicleNameForEventMessage(GameState &state) const;
 
 	const bool cargoContainsAlienLoot() const;
+	bool cargoDeliverableAtCurrentBuilding(const Cargo &c) const;
 
 	// Following members are not serialized, but rather setup during game
 


andremmfaria:

```cpp
// tests/test_alien_cargo.cpp (core; difficulty0 loading boilerplate omitted).
// baseA = the starting base, baseB = a second base created the way test_lab_assignment does.
// Containment at B is toggled by editing base->facilities directly, since Base::destroyFacility
// refuses in this synthetic setup. bio = the first AEquipmentType with bioStorage set.
static sp<Vehicle> spawnLootCarrier(sp<GameState> state, StateRef<Building> home,
                                    StateRef<Building> landedAt)
{
	auto v = state->current_city->placeVehicle(
	    *state, StateRef<VehicleType>{state.get(), UString("VEHICLETYPE_BIOTRANS")},
	    state->getPlayer(), landedAt);
	v->homeBuilding = home;
	return v;
}

// (1) the bug: B has containment, loot is addressed to A, craft is landed at B
setAlienContainment(*state, baseB->base, true);
auto v = spawnLootCarrier(state, baseA, baseB);
auto before = baseB->base->inventoryBioEquipment[bio.id];
v->cargo.emplace_back(*state, bio, 1, 0, nullptr, baseA); // originalOwner null == combat loot

v->updateCargo(*state);
bool gotFerryMission = !v->missions.empty() &&
                       v->missions.back().type == VehicleMission::MissionType::OfferService;
v->getServiceDestination(*state);
auto after = baseB->base->inventoryBioEquipment[bio.id];
// expect: gotFerryMission, v->cargo.empty(), after == before + 1

// (2) control: same, but setAlienContainment(*state, baseB->base, false)
//     expect cargo kept aboard as Cargo::Type::Bio, inventory unchanged
// (3) control: same as (1) but emplace_back(..., state->getPlayer(), baseA), i.e. a transfer
//     expect cargo kept aboard even though B has containment
// (4) control: craft landed at the first building with no base at all
//     expect cargo kept aboard
```

Output on master `9804cfd9` (`vehicle.cpp`/`vehicle.h` restored, exit 1):
```
FAIL bio_loot_delivers_to_second_base_with_containment
PASS bio_loot_stays_aboard_without_containment
PASS transfer_cargo_not_unloaded_at_foreign_base
PASS loot_not_unloaded_at_non_player_building
```

```
PASS bio_loot_delivers_to_second_base_with_containment
PASS bio_loot_stays_aboard_without_containment
PASS transfer_cargo_not_unloaded_at_foreign_base
PASS loot_not_unloaded_at_non_player_building
```

Only the bug case flips. The three controls pass in both legs, so transfers and non-base buildings keep the exact-destination path. `ctest` with the harness in the tree: 12/12.

PR #1672, opened 2026-09-08
Battle: let a stunned unit rise on an adjacent tile when blocked (#742)

### LINKED ISSUE #742: [Battlescape] Stunned units that are stood on by an agent or enemy can wake up but remain stuck
Stunned units that are stood on can wake up, but remain stuck (sometimes causing mission ending to not occur as the user believes everything is unconscious)

In OG, stunned units under another unit would either remain stunned, or with a tiny (i.e. rare) chance wake up, but appear next to or behind the agent - which got around the UFO/TFTD abuse of just sitting on stunned units

OpenApoc needs to implement the behaviour of a stunned unit appearing next to or behind an agent when it wakes up (if it wakes up)

andremmfaria:
Fixes #742.

## Root cause

`BattleUnit::tryToRiseUp` (`battleunit.cpp:4430-4434`) hard-returns when a conscious unit occupies the downed unit's tile, and does nothing else:

```cpp
// Do not rise up if unit is standing on us
if (tileObject->getOwningTile()->getUnitIfPresent(true, true, false, tileObject))
{
	return;
}
```

`updateRegen` retries this once per game second, so a small unit whose stun has fully decayed stays in the Downed limbo (not conscious, not unconscious, no AI, not selectable) for as long as anyone stands on it, which a player or the AI can keep up indefinitely just by not moving.

## Fix

When the tile is blocked and the unit is not large, walk the eight compass headings (`dir_facing_map`) and for each try the tile at the unit's own z, then z-1, then z+1, taking the first that is in bounds and passes `BattleUnitTileHelper::canEnterTile`. If one is found, `setPosition` the unit to that tile's resting position with `goal = true` and call `resetGoal()`, then fall through into the existing rise-body-state logic unchanged. If no free tile exists, or the unit is large, return exactly as before, so the per-second retry keeps trying.

Large-unit displacement is deliberately out of scope, matching Grichmann's note on the issue.

One thing worth flagging for review: the obvious move is to copy `updateGiveWay`'s `canEnterTile(from, to) && canEnterTile(to, from)` pattern, and that is wrong here. The reverse call swaps the roles, so its "is the target passable" step examines the unit's own tile for a blocking occupant, which in this scenario is occupied by definition. It returns `false` for every candidate, unconditionally. `updateGiveWay` gets away with it because the unit giving way is standing on its own unblocked tile. Only the forward check is used here; the harness caught this immediately, on a hand-picked fully clear 3x3 patch.

## Verification

Headless harness (`test_stunned_rise`, removed after validation): a real base defence battle against the player's own base with zero hostiles, a unit downed and stripped of equipment, a second unit moved onto its tile, the stun cleared, then one game second of ticks.

Pre-fix (`9804cfd9`):

```
FAIL blocked_rise expected=displaced_and_rising got=stuck_or_wrong
PASS control_rise_in_place expected=in_place_and_rising got=in_place_and_rising
PASS large_unit_blocked expected=stayed_down got=stayed_down
```

This branch:

```
PASS blocked_rise expected=displaced_and_rising got=displaced_and_rising
PASS control_rise_in_place expected=in_place_and_rising got=in_place_and_rising
PASS large_unit_blocked expected=stayed_down got=stayed_down
```

Only the blocked case changes; the unblocked control and the large-unit early return are identical in both directions. Full ctest green (RelWithDebInfo, Linux); fork CI Lint + CMake green.

Separately, ticking this randomly generated base map turns up two latent issues that are present on master, reproduce with this change reverted, and are not touched here: a unit or dropped item occasionally starts falling out of map bounds and one call site segfaults on the resulting null tile, and a freshly relocated unit is occasionally not found by `Tile::getUnitIfPresent` right afterwards even though `occupiedTiles` looks correct. Neither is mentioned in this issue; I can file them separately if that is useful.

### DIFF
--- a/game/state/battle/battleunit.cpp
+++ b/game/state/battle/battleunit.cpp
@@ -4431,7 +4431,39 @@ void BattleUnit::tryToRiseUp(GameState &state)
 {
 	// Do not rise up if unit is standing on us
 	if (tileObject->getOwningTile()->getUnitIfPresent(true, true, false, tileObject))
-		return;
+	{
+		if (isLarge())
+		{
+			return;
+		}
+		auto from = tileObject->getOwningTile();
+		auto helper = BattleUnitTileHelper{tileObject->map, *this};
+		Tile *freeTile = nullptr;
+		for (int dir = 0; dir <= 7 && !freeTile; dir++)
+		{
+			auto heading = dir_facing_map.at(dir);
+			for (int dz : {0, -1, 1})
+			{
+				Vec3<int> pos = {position.x + heading.x, position.y + heading.y, position.z + dz};
+				if (!tileObject->map.isTileInBounds(pos))
+				{
+					continue;
+				}
+				auto to = tileObject->map.getTile(pos);
+				if (helper.canEnterTile(from, to))
+				{
+					freeTile = to;
+					break;
+				}
+			}
+		}
+		if (!freeTile)
+		{
+			return;
+		}
+		setPosition(state, freeTile->getRestingPosition(isLarge()), true);
+		resetGoal();
+	}
 
 	// Find state we can rise into (with animation)
 	auto targetState = BodyState::Standing;


andremmfaria:

```cpp
// tests/test_stunned_rise.cpp (core; difficulty0 loading boilerplate omitted).
// Base defence battle against the player's own base with no hostiles: the starting soldiers are
// already in that building, so no vehicle or UFO setup is needed. All but 4 soldiers have their
// currentBuilding cleared first, so BattleMap::createBattle only pulls those 4 in.
state->rng.seed(742);
                    noVehicle, state->current_base->building);
Battle::enterBattle(*state);
// isolateUnit() then moves three of the four to mutually distant tiles whose whole 3x3 patch
// passes Tile::getSolidGround(false), so nobody starts falling once ticks are advanced.

bool downUnit(GameState &state, sp<BattleUnit> unit, unsigned int maxTicks)
{
	stripEquipment(state, unit); // no DropItem missions, keeps item physics out of the test
	unit->stunDamage = unit->agent->getHealth() + 10;
	unit->fallUnconscious(state);
	unsigned int elapsed = 0, step = TICKS_PER_SECOND / 4;
	while (unit->current_body_state != BodyState::Downed && elapsed < maxTicks)
	{
		state.current_battle->update(state, step);
		elapsed += step;
	}
	return unit->current_body_state == BodyState::Downed;
}

bool blockTile(GameState &state, sp<BattleUnit> blocker, sp<BattleUnit> occupant)
{
	auto tile = occupant->tileObject->getOwningTile();
	blocker->setPosition(state, tile->getRestingPosition(blocker->isLarge()), true);
	blocker->resetGoal();
	// the exact predicate tryToRiseUp checks
	return tile->getUnitIfPresent(true, true, false, occupant->tileObject) != nullptr;
}

// blocked_rise: down the unit, stand a second unit on its tile, clear the stun, tick one
// game second (enough for updateRegen's once-per-second retry to fire).
auto originalPos = unit->tileObject->getOwningTile()->position;
downUnit(state, unit, TICKS_PER_SECOND * 2);
blockTile(state, blocker, unit);
auto blockerPos = blocker->tileObject->getOwningTile()->position;
unit->stunDamage = 0;
state.current_battle->update(state, TICKS_PER_SECOND);
auto newPos = unit->tileObject->getOwningTile()->position;
bool pass = newPos != originalPos &&                                  // displaced
            std::abs(newPos.x - originalPos.x) <= 1 &&                // to a neighbour
            std::abs(newPos.y - originalPos.y) <= 1 &&
            std::abs(newPos.z - originalPos.z) <= 1 &&
            blocker->tileObject->getOwningTile()->position == blockerPos && // blocker unmoved
            (unit->isConscious() || unit->target_body_state != BodyState::Downed);

// control_rise_in_place: same without a blocker, must rise where it lies.
// large_unit_blocked: same as blocked_rise with bodyType->large forced true for that one tick
// (restored right after), must still stay down.
```

Output on master `9804cfd9`:
```
FAIL blocked_rise expected=displaced_and_rising got=stuck_or_wrong
PASS control_rise_in_place expected=in_place_and_rising got=in_place_and_rising
PASS large_unit_blocked expected=stayed_down got=stayed_down
```

Output on this branch:
```
PASS blocked_rise expected=displaced_and_rising got=displaced_and_rising
PASS control_rise_in_place expected=in_place_and_rising got=in_place_and_rising
PASS large_unit_blocked expected=stayed_down got=stayed_down
test_stunned_rise success - all cases passed
```

Only the blocked case changes; the unblocked control and the large-unit early return behave the same in both directions. Full `ctest`: 11/11 with the harness removed again.

This is also what caught the `canEnterTile(to, from)` trap described in the PR body: with the reverse check in place, `blocked_rise` reported no free tile on a hand-picked, fully clear 3x3 patch.

Issue #264, opened 2017-11-23
DEVELOPMENT #2:  To reach a mostly complete state, we also require:


makus82:
- [ ] Proper portal location in alien dimension
- [ ] Proper portal movement
- [ ] Alien Takeover screen
- [ ] Score screen and graphs
- [ ] Proper collapse algorithm for alien tubes
- [ ] Victory & Defeat screens
- [ ] UFO growth halting when relevant building is destroyed
- [ ] Organisations properly buying their vehicles
- [ ] Organisations making treaties, raiding each other and x-com, sending illegal flyers
- [ ] Proper Music (not just 3 looped tracks)
- [ ] Overspawn

JonnyH:
Removed the alien infiltration graph and moved to alpha - it's pretty much unplayable without knowing that stuff :)

makus82:
Every Day Score screen 
![image](https://user-images.githubusercontent.com/11581207/86574234-2ffb3c00-bf7e-11ea-824e-a13d9212bccc.png)
![image](https://user-images.githubusercontent.com/11581207/86574265-3f7a8500-bf7e-11ea-9500-a50d11cec76a.png)

Weekly Funding Assessment screen
![image](https://user-images.githubusercontent.com/11581207/86574927-3342f780-bf7f-11ea-8012-a6a6bbbdaa22.png)
And after
![image](https://user-images.githubusercontent.com/11581207/86575041-5372b680-bf7f-11ea-9e40-5dd74f67ba34.png)
![image](https://user-images.githubusercontent.com/11581207/86575061-5a99c480-bf7f-11ea-88a5-0eec3aae0202.png)
![image](https://user-images.githubusercontent.com/11581207/86575083-61c0d280-bf7f-11ea-9975-cb1db1bfd562.png)

Istrebitel:
Does this
**"Organisations order their craft to retreat when heavily damaged and other AI improvements"**
not already work? I believe a craft is retreating when it's low on health, as per its aggressiveness level? Regardless of the org, even player craft will do it?

makus82:
I will test ... but i think u right

makus82:
> Does this
> **"Organisations order their craft to retreat when heavily damaged and other AI improvements"**
> not already work? I believe a craft is retreating when it's low on health, as per its aggressiveness level? Regardless of the org, even player craft will do it?

it seems its not implemented even if mode is defensive and phoenix hovercar is HP left 5%

makus82:
[...]

PR #1673, opened 2026-09-08
City: end the game on the final victory and the last base lost (#766)

### LINKED ISSUE #766: End Game does not trigger on destruction of dimension gates
The end game video and prompt does not trigger on leaving the dimension gate generators with a successful mission.

Instead the player is able to go back to the city and cause chaos (see today's quickmind stream)

Maybe freeplay should be an option longer term, but for now, the game should end as per OG

On mission successful, as soon as return to cityscape display victory prompt, then run video, then return to main menu (or if we add it, prompt to free-play)

andremmfaria:
Fixes #766.

## Root cause

`Building::victory` is data-driven and read correctly, but the result never leaves the function. `Battle::exitBattle` sets a local `bool victory` from `building->victory` for `RaidAliens` (`battle.cpp:3400`) and `RaidHumans` (`:3445`), then drops it:

```cpp
LogError("You won, but we have no screen for that yet LOL!");
```

The defeat side is the same stub: `Base::die` logs `Player lost, but we have no screen for that yet!` in the `state.player_bases.empty()` branch and pushes no event. There is no `GameWon` or `GameLost` in `GameEventType`, so `CityView` has nothing to react to and the game simply continues.

## Fix

The minimal slice that makes both outcomes reach the player:

* `gameeventtypes.h`: `GameWon` and `GameLost` appended to the Misc group, no reordering.
* `gameevent.cpp`: two `GameEvent::message()` cases returning `tr(...)` text. No new event subclass, neither carries a payload.
* `battle.cpp:3471`: the `LogError` becomes `state.eventFromBattle = GameEventType::GameWon;` on victory, reusing the already-serialised field the mission-complete events use.
* `gamestate.cpp` `updateAfterBattle()`: a `GameWon` case that pushes the plain event, identical in shape to the existing `MissionCompletedBuildingAlien` case.
* `base.cpp` `Base::die`: the `LogError` becomes a `GameLost` event. When a battle is active the event goes through `eventFromBattle` (`exitBattle` calls `die` before `CityView` exists); otherwise it is pushed directly, which is what `Base::die` already does for `BaseDestroyed`. The `player_bases.empty()` guard means it fires exactly once, whichever caller killed the base.
* `cityview.cpp`: `GameWon` and `GameLost` cases in the unconditional switch, modelled on `FacilityCompleted`: `setUpdateSpeed(Pause)` plus a single-OK `MessageBox` whose callback does `stageQueueCommand({REPLACEALL, mksp<MainMenu>()})` (the `ingameoptions.cpp:212` precedent).
* `gamestate_serialize.xml`: the two names appended to the `GameEventType` list.

Save format is additive. The enum is serialised by name, nothing is removed, renamed or renumbered, and no existing save can contain either value.

## Out of scope, deliberately

* Proper victory and defeat art screens and a final score screen (#264 item 5).
* The ending video: the dev CD image here has no `SMK` directory and I could not confirm a real filename from the extractor sources, so this does not invent one.
* Free play / continue after victory.
* Alien takeover as a second defeat trigger (`organisation.cpp` `takeOver`, structurally a different path, no last-base check) (#264 item 6).

One tradeoff worth naming: on a winning raid, `eventFromBattle` is overwritten, so the player sees the victory box instead of the "Mission completed in Alien building" ticker line for that one mission.

## Verification

Headless harness (`test_endgame`, removed after validation): difficulty0 state, a bare `Battle` driven through `updateBeforeBattle`/`exitBattle`, and the last base killed via `Base::die` with a battle active. Assertions compare `magic_enum::enum_name(eventFromBattle)` to a string so the same source compiles on the base commit.

|---|---|---|
| won raid on the Dimension Gate Generator | `MissionCompletedBuildingAlien` | `GameWon` |
| won raid on an Alien Farm (control) | `MissionCompletedBuildingAlien` | `MissionCompletedBuildingAlien` |
| last base dies during a battle | `eventFromBattle = None` | `GameLost` |

Full ctest green (RelWithDebInfo, Linux); fork CI Lint + CMake green.

The `cityview.cpp` switch cases themselves are covered by reading rather than a live run: `Framework`'s event queue is private, draining it needs `Framework::run()`, and `CityView` loads UI forms, so there is no headless precedent in this codebase to copy. If an end-to-end run is wanted, the smallest path is a `UNIT_TEST`-gated queue accessor, which is a larger change than this issue.

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -3470,7 +3470,7 @@ void Battle::exitBattle(GameState &state)
 
 	if (victory)
 	{
-		LogError("You won, but we have no screen for that yet LOL!");
+		state.eventFromBattle = GameEventType::GameWon;
 	}
 
 	state.current_battle = nullptr;
--- a/game/state/city/base.cpp
+++ b/game/state/city/base.cpp
@@ -144,7 +144,14 @@ void Base::die(GameState &state, bool collapse)
 	state.current_base.clear();
 	if (state.player_bases.empty())
 	{
-		LogError("Player lost, but we have no screen for that yet!");
+		if (state.current_battle)
+		{
+			state.eventFromBattle = GameEventType::GameLost;
+		}
+		else
+		{
+			fw().pushEvent(new GameEvent(GameEventType::GameLost));
+		}
 		return;
 	}
 	else
--- a/game/state/gameevent.cpp
+++ b/game/state/gameevent.cpp
@@ -74,6 +74,11 @@ UString GameEvent::message()
 			return tr("X-COM returning from UFO mission.");
 		case GameEventType::BuildingDisabled:
 			return tr("Building has been disabled");
+		case GameEventType::GameWon:
+			return tr("The link with the Alien dimension is broken forever. The Aliens are "
+			          "vanquished and victory is ours!");
+		case GameEventType::GameLost:
+			return tr("The last X-COM base has fallen. Earth is lost to the Alien invasion.");
 		default:
 			break;
 	}
--- a/game/state/gameeventtypes.h
+++ b/game/state/gameeventtypes.h
@@ -115,6 +115,8 @@ enum class GameEventType
 	BuildingDisabled,
 	MissionCompletedVehicle,
 	MissionCompletedBuildingAlien,
+	GameWon,
+	GameLost,
 
 	None
 };
--- a/game/state/gamestate.cpp
+++ b/game/state/gamestate.cpp
@@ -1780,6 +1780,8 @@ void GameState::updateAfterBattle()
 			break;
 		}
 		case GameEventType::MissionCompletedVehicle:
+		case GameEventType::GameWon:
+		case GameEventType::GameLost:
 		{
 			fw().pushEvent(new GameEvent(eventFromBattle));
 			break;
--- a/game/state/gamestate_serialize.xml
+++ b/game/state/gamestate_serialize.xml
@@ -414,6 +414,8 @@
     <value>BuildingDisabled</value>
     <value>MissionCompletedVehicle</value>
     <value>MissionCompletedBuildingAlien</value>
+    <value>GameWon</value>
+    <value>GameLost</value>
     <value>None</value>
   </enum>
   <enum>
--- a/game/ui/tileview/cityview.cpp
+++ b/game/ui/tileview/cityview.cpp
@@ -65,6 +65,7 @@
 #include "game/ui/components/locationscreen.h"
 #include "game/ui/general/aequipscreen.h"
 #include "game/ui/general/ingameoptions.h"
+#include "game/ui/general/mainmenu.h"
 #include "game/ui/general/messagebox.h"
 #include "game/ui/general/messagelogscreen.h"
 #include "game/ui/general/notificationscreen.h"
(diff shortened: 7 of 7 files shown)


andremmfaria:

```cpp
// tests/test_endgame.cpp (core; difficulty0 loading boilerplate omitted).
// Assertions compare magic_enum::enum_name(eventFromBattle) to a string, so the same source
// compiles against the base commit where GameWon/GameLost do not exist yet.
static bool reportResult(const UString &name, const UString &expected, GameEventType got)
{
	UString gotName = UString(magic_enum::enum_name(got));
	bool ok = gotName == expected;
	LogWarning(format("{0} {1} expected={2} got={3}", ok ? "PASS" : "FAIL", name, expected, gotName));
	return ok;
}

static GameEventType runRaidAliensWin(sp<GameState> state, StateRef<Building> building)
{
	state->updateBeforeBattle();
	auto battle = mksp<Battle>();
	battle->mission_type = Battle::MissionType::RaidAliens;
	battle->mission_location_building = building;
	battle->playerWon = true;
	state->current_battle = battle;
	Battle::exitBattle(*state);
	return state->eventFromBattle;
}

// (1) the victory path
StateRef<Building> gate = {state.get(), "BUILDING_DIMENSION_GATE_GENERATOR"}; // victory == true
reportResult("victory_on_gate_building", "GameWon", runRaidAliensWin(state, gate));

// (2) control: an ordinary alien building, victory unset
StateRef<Building> farm = {state.get(), "BUILDING_ALIEN_FARM"};
reportResult("no_victory_on_normal_building", "MissionCompletedBuildingAlien",
             runRaidAliensWin(state, farm));

// (3) the defeat path: shrink to one base, kill it with a battle active
while (state->player_bases.size() > 1)
	state->player_bases.erase(state->player_bases.begin()->first);
auto lastBase = state->player_bases.begin()->second;
state->current_battle = mksp<Battle>();
state->eventFromBattle = GameEventType::None;
lastBase->die(*state, false); // the same call battle.cpp:3430 makes on a lost BaseDefense
state->current_battle = nullptr;
reportResult("defeat_on_last_base_in_battle", "GameLost", state->eventFromBattle);
```

Output on master `9804cfd9` (the seven touched files restored, exit 1):
```
FAIL victory_on_gate_building expected=GameWon got=MissionCompletedBuildingAlien
PASS no_victory_on_normal_building expected=MissionCompletedBuildingAlien got=MissionCompletedBuildingAlien
PASS defeat_last_base_removed expected=true got=true
FAIL defeat_on_last_base_in_battle expected=GameLost got=None
```

```
PASS victory_on_gate_building expected=GameWon got=GameWon
PASS no_victory_on_normal_building expected=MissionCompletedBuildingAlien got=MissionCompletedBuildingAlien
PASS defeat_last_base_removed expected=true got=true
PASS defeat_on_last_base_in_battle expected=GameLost got=GameLost
test_endgame success - all tests passed
```

Winning the Dimension Gate Generator raid and winning an ordinary alien building raid are indistinguishable pre-fix, which is the bug. `ctest` with the harness in the tree: 12/12.

The `cityview.cpp` switch cases are not exercised here: `Framework`'s event queue is private, draining it needs `Framework::run()`, and `CityView` loads UI forms, so there is no headless precedent in this codebase to copy. This covers the code that actually changed in `battle.cpp`, `base.cpp` and `gameevent.cpp`.

PR #1674, opened 2026-09-08
Battle: use the original game's combat experience rules (#1665)

### LINKED ISSUE #1665: Battlescape experience gains do not match the original game
Split out of #263 item 1 ("Agent Training from Battlescape"), which is the only item on that roadmap still needing code.

### Symptom

Agents gain far too much from a single mission. From #263 (makus82, 2020-07-22, superhuman, two agents vs 1-2 Anthropods, 4 Spitters, 2 Hyperworms, 3 Brainsuckers):

| Agent 1 | before | after | | Agent 2 | before | after |
|---|---|---|---|---|---|---|
| Health | 56 | 62 | | Health | 48 | 54 |
| Accuracy | 36 | 40 | | Accuracy | 38 | 40 |
| Reactions | 37 | 39 | | Reactions | 16 | 17 |
| Speed | 80 | 80 | | Speed | 80 | 88 |
| Stamina | 33 | 41 | | Stamina | 34 | 41 |
| Strength | 58 | 64 | | Strength | 63 | 66 |

### Why

`BattleUnit::processExperience` (game/state/battle/battleunit.cpp, unchanged since 2017) is a placeholder using the X-COM 1/2 primary/secondary system, and the code says so:

```
// FIXME: Ensure correct
// For now, using X-Com 1/2 system of primary/secondary stats,
```

The original game's rules are documented on UFOpaedia ([Combat Experience (Apocalypse)](https://www.ufopaedia.org/index.php/Combat_Experience_(Apocalypse))) and differ on every point:

| | Original game | OpenApoc today |
|---|---|---|
| Randomness | none, deterministic | random rolls everywhere |
| Strength | never from combat | +rand(0..2) + 10% of missing stat, every mission |
| Psi stats | never from combat | +1/+2 XP per psi attempt, converted with a roll |
| Health | +1 if kill score 15..29, +2 if 30+ | +rand(0..2) + 10% of missing stat, every mission |
| Bravery | +10 if kill score / bravery > 0.5 | +1 XP per turn interval whenever morale is above 50, then a roll |
| Accuracy | +ceil((96 - acc)/100 * floor(hits/4)), hits capped at 40, stat capped at 96 | roll of 0..6 from hit count |
| Reactions | +min(floor(reaction hits / reactions), 5), turn-based only | roll from interrupt count (TB), or from total XP (RT) |
| Stamina / Speed | stamina XP from running; stamina += floor(XP/stamina)/2; speed += min(floor(XP/stamina), 5) | +rand(0..2) + 10% of missing stat, every mission |
| Androids | no gains at all | gain at 10% rate |
| Trigger for secondary stats | n/a | total XP > 1, which bravery XP alone guarantees on any mission longer than one turn interval |

The bravery timer is what makes the health/speed/stamina/strength block fire after essentially every mission for humans and hybrids, which matches the table above.

The kill-score input the original uses already exists: `BattleUnit::combatRating` accumulates `agent->type->score` per kill.

### Proposed change

Replace `processExperience` with the documented deterministic rules, drop psi and strength gains from combat, skip androids, and add stamina XP accrual to the run code path. Difficulty is not a factor in the original formulas. Base training (Training Area / Psi-Gym) is out of scope.

andremmfaria:
Fixes #1665.

## Root cause

`BattleUnit::processExperience` (`battleunit.cpp:4158`) and its helper `rollForPrimaryStat` (`:4134`) are a 2017 placeholder, marked `FIXME ... for now, using X-Com 1/2 system` in the source: random dice rolls instead of Apocalypse's deterministic formulas. The accumulators it reads do not match the original rules either:

* accuracy XP is `+1` per hit, which is right, but uncapped (`battleunit.cpp:1942`);
* reaction XP goes to the interrupt **receiver** just for being offered an interrupt chance (`battle.cpp:2426`), not to the firer for landing a reaction shot;
* bravery XP comes from a per-turn morale-check timer (`battleunit.cpp:2213`), not from kills;
* psi stats accrue XP on psi attempts (`battleunit.cpp:1128-1137`) even though psi never changes from combat;
* there is no kill-score and no stamina-spent tracking at all, so health, bravery, speed and stamina can never move for the right reasons.

## Fix

`processExperience` now implements the documented rules ([ufopaedia: Combat Experience (Apocalypse)](https://www.ufopaedia.org/index.php/Combat_Experience_(Apocalypse))), deterministic, no RNG, gated on `agent->type->canTrain` so androids gain nothing:

| Stat | Trigger | Accrual cap | Gain | Stat cap |
|---|---|---|---|---|
| Accuracy | +1 per hit landed on a living unit | 40 | `ceil((96 - accuracy) / 100 * floor(accXP / 4))` | 96 |
| Reactions | +1 per hit landed by a reaction shot, turn based only | none | `min(floor(reactXP / reactions), 5)` | 100 |
| Speed | stamina spent running | none | `min(floor(stamXP / stamina), 5)` | 100 |
| Stamina | stamina spent running | none | `floor(stamXP / stamina) * 10` internal | 2000 |
| Health | sum of `score` of enemies killed | n/a | `+1` at 15-29, `+2` at 30+ | 100 |
| Bravery | same kill score | n/a | `+10` if `killScore / bravery > 0.5` | 100 |

Strength and every psi stat are never written. Division guards: no reaction gain at `reactions == 0`, no stamina or speed gain at displayed `stamina == 0`, and `bravery == 0` with any kill counts as above the threshold.

`AgentStats experiencePoints` is replaced by a small `BattleUnitExperience` struct holding `accuracyHits`, `reactionHits`, `staminaSpent` and `killScore`, registered in `gamestate_serialize.xml` so mid-battle saves keep working. `rollForPrimaryStat` is removed (no other caller). The accrual sites move accordingly: accuracy hits capped at 40, reaction hits recorded on the firer when it is in `Battle::interruptUnits` in turn-based mode, kill score added next to the existing `combatRating += agent->type->score` on the hostile-kill branch, stamina spend recorded alongside the existing `modified_stats.stamina -=` lines, and the psi, morale-timer and interrupt-receiver increments deleted.

`interruptUnits` membership is used as the "this is a reaction shot" marker because no explicit flag exists on `BattleUnit` or `Projectile`; a unit only enters that map through the interrupt decision flow, so any shot fired while it is there is a reaction shot by construction.

## Assumptions worth a maintainer's eye

* **Stamina XP unit.** The source does not state it. Raw internal spend (10 or 30 per event, matching the existing magnitudes) is accumulated and converted once via `/ 20` at the end, so a per-event `10/20` division does not throw away half an XP on every prone-drag tick.
* **Stamina and speed share one raw term with different caps.** The source gives `min(floor(stamXP/stamina), 5)` for speed and `floor(stamXP/stamina)/2` displayed for stamina. Read literally, only the speed gain carries the per-mission ceiling of 5. Implemented literally; if the intent was for both to share it, that is a one-line change.
* Denominators in the reaction and stamina formulas are the unit's current displayed stat before this mission's gain is applied, matching how the accuracy formula uses live `accuracy`.
* The accuracy ceiling is 96 rather than the general 100, per the source.
* `improvementPercentagePhysical` / `improvementPercentagePsi` are left in place but become unused, per the issue discussion. Androids are excluded via `canTrain`, which is already `false` for them; `improvementPercentagePhysical` is 10 for androids, so comparing it against zero would not have excluded them.

## Verification

Headless harness (`test_battle_experience`, removed after validation): difficulty0 state, real agents from `AgentGenerator`, a bare `BattleUnit` with only `->agent` set, `processExperience` called, assertions on `current_stats`.

```
case                       expected  pre-fix (9804cfd9)       this branch
accuracy 50, 40 hits         55        56                       55
accuracy 96, 40 hits         96        98                       96
reactions 20, 19 r-hits      20        26                       20
reactions 0                   0         3                        0
stamina 400, 2000 spent     450       400                      450
speed 50, 2000 spent         55        50                       55
health 50, killScore 30      52        50                       52
bravery 20, killScore 11     30        20                       30
bravery 0, killScore 1       10         0                       10
android, everything maxed  unchanged  acc+2 rea+2 sta+36 hp+1  unchanged
strength 45                  45        50                       45
```

Pre-fix 19 of 33 cases fail (random rolls, androids training, strength growing from combat, kills doing nothing). This branch: 33 of 33 pass. Full ctest green (RelWithDebInfo, Linux); fork CI Lint + CMake green.

### DIFF
--- a/game/state/battle/battle.cpp
+++ b/game/state/battle/battle.cpp
@@ -2423,7 +2423,6 @@ void Battle::giveInterruptChanceToUnit(GameState &state, StateRef<BattleUnit> gi
 				fw().pushEvent(new GameLocationEvent(GameEventType::ZoomView, receiver->position));
 			}
 			interruptQueue.emplace(receiver, receiver->agent->getTULimit(reactionValue));
-			receiver->experiencePoints.reactions++;
 		}
 	}
 }
--- a/game/state/battle/battleunit.cpp
+++ b/game/state/battle/battleunit.cpp
@@ -1125,16 +1125,12 @@ bool BattleUnit::startAttackPsiInternal(GameState &state, StateRef<BattleUnit> t
 	}
 	int chance = getPsiChanceForEquipment(target, status, item);
 	int roll = randBoundsExclusive(state.rng, 0, 100);
-	experiencePoints.psi_attack++;
-	experiencePoints.psi_energy++;
 	LogWarning("Psi Attack #{0} Roll {1} Chance {2} {3} Attacker {4} Target {5}", (int)status, roll,
 	           chance, roll < chance ? (UString) "SUCCESS" : (UString) "FAILURE", id, target->id);
 	if (roll >= chance)
 	{
 		return false;
 	}
-	experiencePoints.psi_attack += 2;
-	experiencePoints.psi_energy += 2;
 
 	// Attack hit, apply effects
 
@@ -1939,7 +1935,16 @@ bool BattleUnit::handleCollision(GameState &state, Collision &c)
 			notifyHit(position - glm::normalize(projectile->velocity) * 1.41f);
 			if (projectile->firerUnit)
 			{
-				projectile->firerUnit->experiencePoints.accuracy++;
+				if (projectile->firerUnit->experience.accuracyHits < 40)
+				{
+					projectile->firerUnit->experience.accuracyHits++;
+				}
+				if (state.current_battle->mode == Battle::Mode::TurnBased &&
+				    state.current_battle->interruptUnits.find(projectile->firerUnit) !=
+				        state.current_battle->interruptUnits.end())
+				{
+					projectile->firerUnit->experience.reactionHits++;
+				}
 			}
 			return applyDamage(state, projectile->damage, projectile->damageType, partHit,
 			                   DamageSource::Impact, projectile->firerUnit);
@@ -2208,11 +2213,7 @@ void BattleUnit::updateMorale(GameState &state, unsigned int ticks)
 		{
 			moraleTicksAccumulated -= LOWMORALE_CHECK_INTERVAL;
 
-			if (randBoundsExclusive(state.rng, 0, 100) >= 100 - 2 * agent->modified_stats.morale)
-			{
-				experiencePoints.bravery++;
-			}
-			else
+			if (randBoundsExclusive(state.rng, 0, 100) < 100 - 2 * agent->modified_stats.morale)
 			{
 				moraleStateTicksRemaining = TICKS_PER_LOWMORALE_STATE;
 				moraleState = (MoraleState)(randBoundsInclusive(state.rng, 1, 3));
@@ -2363,6 +2364,7 @@ void BattleUnit::updateRegen(GameState &state, unsigned int ticks)
 					if (agent->modified_stats.stamina > 10)
 					{
 						agent->modified_stats.stamina -= 10;
+						experience.staminaSpent += 10;
 					}
 					else
 					{
@@ -2374,6 +2376,7 @@ void BattleUnit::updateRegen(GameState &state, unsigned int ticks)
 					if (agent->modified_stats.stamina > 30)
 					{
 						agent->modified_stats.stamina -= 30;
+						experience.staminaSpent += 30;
 					}
 					else
 					{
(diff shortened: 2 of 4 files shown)


andremmfaria:

```cpp
// tests/test_battle_experience.cpp (core; difficulty0 loading boilerplate omitted).
// processExperience only needs unit->agent, so a bare BattleUnit is enough. Agents come from
// AgentGenerator so the stat ranges are real; each case overwrites the stats it cares about.
static StateRef<Agent> makeAgent(sp<GameState> state, const UString &typeId)
{
	StateRef<AgentType> type{state.get(), typeId};
	return state->agent_generator.createAgent(*state, state->getPlayer(), type);
}

static sp<BattleUnit> makeUnit(StateRef<Agent> agent)
{
	auto unit = mksp<BattleUnit>();
	unit->agent = agent;
	return unit;
}

// accuracy: ceil((96 - accuracy) / 100 * floor(accXP / 4)), capped at 96
auto agent = makeAgent(state, "AGENTTYPE_X-COM_AGENT_HUMAN");
agent->current_stats.accuracy = 50;
auto unit = makeUnit(agent);
unit->experience.accuracyHits = 40;
unit->processExperience(*state);
checkStat("accuracy_gain", 55, agent->current_stats.accuracy);
// same at accuracy 96 -> stays 96; at 3 hits -> no gain

// reactions: min(floor(reactXP / reactions), 5), and no gain (no divide) at reactions == 0
unit->experience.reactionHits = 100;   // reactions 20 -> 25
unit->experience.reactionHits = 19;    // reactions 20 -> unchanged
unit->experience.reactionHits = 3;     // reactions  0 -> stays 0

// stamina and speed share floor(stamXP / stamina); speed is capped at +5, stamina is x10 internal
unit->experience.staminaSpent = 2000;  // stamina 400 -> 450, speed 50 -> 55

// health and bravery from the kill score
unit->experience.killScore = 30;       // health 50 -> 52 (+1 at 15..29, +2 at 30+)
unit->experience.killScore = 11;       // bravery 20 -> 30 (killScore / bravery > 0.5)
unit->experience.killScore = 1;        // bravery  0 -> 10 (zero counts as above the threshold)

// androids: makeAgent(state, "AGENTTYPE_X-COM_AGENT_ANDROID") with every accumulator maxed,
// every stat must be unchanged (canTrain == false). Strength and all psi stats are asserted
// unchanged for humans too.
```

The pre-fix leg is the same 33 cases against the old `AgentStats experiencePoints` accumulator; it also has to give `state.current_battle` a TurnBased `Battle`, because the old `processExperience` dereferences it.

Output on master `9804cfd9` (19 of 33 fail, exit 1), the notable ones:
```
FAIL accuracy_cap_96 expected=96 got=98
FAIL reactions_zero_no_crash expected=0 got=3
FAIL stamina_gain expected=450 got=400
FAIL health_killscore_30 expected=52 got=50
FAIL bravery_zero_baseline_gain expected=10 got=0
FAIL android_stamina_unchanged expected=400 got=436
FAIL strength_unchanged expected=45 got=50
```

Output on this branch:
```
PASS accuracy_gain expected=55 got=55
PASS accuracy_cap_96 expected=96 got=96
PASS reactions_gain expected=25 got=25
PASS reactions_zero_no_crash expected=0 got=0
PASS stamina_gain expected=450 got=450
PASS speed_gain expected=55 got=55
PASS health_killscore_30 expected=52 got=52
PASS bravery_gain expected=30 got=30
PASS android_stamina_unchanged expected=400 got=400
PASS strength_unchanged expected=45 got=45
... 33/33 PASS
```

Pre-fix the failures are exactly the placeholder's behaviour: random rolls overshooting the caps, androids training, strength growing from combat, and kills doing nothing at all. `ctest` with the harness in the tree: 12/12.

Issue #755, opened 2019-09-12
[Battlescape] Agents falling or jumping off the map, or through tiles and off the map will die


FilmBoy84:
From Quickmind in issue #581 

Using the save file, make the agent jump 2 tiles forward. They will jump through the wall on the first jump, and then appear to land on the stairs but actually fall through the world on the second jump. This will also kill the agent. I was unable to reproduce a fight with explosions to get this error but stumbled upon this technique.

This is using build OpenApoc-v0.2.0.2-12-g8f8b94c1 with V6.1 of EMW.

https://github.com/OpenApoc/OpenApoc/files/3588351/save_test.zip

PR #1676, opened 2026-09-08
Battle: land falling units on the floor they cross and enforce the jump rules (#1581)

### LINKED ISSUE #1581: Agents can jump through walls, floors and up obstacles and stairs
Hello, creating a new issue from [#1057 comment](https://github.com/OpenApoc/OpenApoc/issues/1057#issuecomment-3985732767) as it's a separate topic.

**Describe the bug**
Agents can use Jump indiscriminately and jump up otherwise impassable obstacles, flights of stairs, or through walls and floors. Also, jumping agent rotates in the direction of jump, and pressing J for jump uses up time units even if you did not actually jump (cursor over agent tile or non-adjacent tile, or built area).

**To Reproduce**
1. Place agent next to a wall, stairs, obstacle
2. Put cursor over the adjacent tile you want to jump to
3. Press J
4. Agent will jump to the space
Jumping through floor described in [#1057 comment](https://github.com/OpenApoc/OpenApoc/issues/1057#issuecomment-3985732767).

**Expected behavior**
Jump in the original game could only be used to jump down. You had to stand next to a ledge, hold J and click to the adjacent tile. You could not jump up something, and you could not jump on an even floor. Also, the landing area needs to be clear (passable). You could jump down a flight of stairs. Jumping did not change the direction the agent was facing, and jumping didn't actually cost any TUs.

**Savegame**
See [#1057 comment](https://github.com/OpenApoc/OpenApoc/issues/1057#issuecomment-3985732767)

**Screenshots**
See [#1057 comment](https://github.com/OpenApoc/OpenApoc/issues/1057#issuecomment-3985732767)

**Additional context**
Even if Jumping is re-defined (some of the differences would make sense), you should definitely not be able to jump through floors and walls. You probably shouldn't be able to jump up on impassable objects. You might not want to jump up a flight of stairs (saves a lot of TUs).

See [#1057 comment](https://github.com/OpenApoc/OpenApoc/issues/1057#issuecomment-3985732767)

andremmfaria:
Fixes #1581.
Refs #1057, #1330.

Two commits, reviewable separately: the falling clamp first, the jump rules on top of it.

## Root cause

**Falling.** `BattleUnit::updateMovementFalling` (`battleunit.cpp:2752-2919`) integrates a whole tick batch in one step, never runs a collision check for conscious units, and only lands the unit when the resting surface of the tile it *ends up in* is above it (`:2893-2896`), and only while `collisionIgnoredTicks` is zero. So a drop that in one update exceeds the 0.2 band of the floor it passes goes straight through to the level below. A J jump makes it worse: it sets `collisionIgnoredTicks = 73` for a 72-tick flight (`:3979`) and aims at the target tile's base z (`battleview.cpp:2293`), so the unit is already below the target floor when the first landing test runs. The same block dereferences `getTile(newPosition)` without a null check (`:2824`).

**Jump rules.** The only gate on a manual jump is `canLaunch` (`:3928-3965`): not flying, an adjacent column, and a parabola that exists. Nothing checks that the target is *lower*, that the landing tile is passable or free of units, or that the edge is not a wall or a closed door. On top of that the UI passes `requireFacing = true`, so the unit turns first, and `BattleUnitMission::update` charges 4 TU plus 1 per turn step. A refused jump gets charged too: `start()` sets `cancelled` and `isFinished` and then calls `update()` once more, and the Jump branch never looked at `cancelled`, so it spent the TU and called `launch()`, which silently did nothing. That is the reporter's "TUs are consumed even when no jump happens", and I could not confirm it from reading alone; the harness reproduced it.

## Fix

**Commit 1, `fix(battle): land falling units on the floor they cross`.** A descending unit is clamped to the first standable surface it crosses in its own column, scanning down from its previous z. The clamp and the ordinary owning-tile test are re-enabled inside the collision-ignore window once the unit is over the column it was launched at, so a launch still leaves its own tile unhindered but lands correctly at the far end. `getTile(int,int,int)` returns null for out-of-map x/y, so the scan is safe at the map edge, and the bottom of the range is clamped to level 0 so a z=0 floor is still found. The `getTile` lookup in the "fell into a unit" block is null-checked, the same lookup PR #1664 gated the below-zero kill on.

**Commit 2, `fix(battle): only let agents jump down to a clear adjacent tile`.** A new `canJumpDown` applies the original rules: the target must be an adjacent column; walking down that column from the unit's level, the first standable tile must be strictly below the surface the unit stands on now (so no jumping up a stair step, no jumping across level floor); that landing tile must be passable and free of static units; and the edge to the target column must not be blocked by a wall, a corner wall on a diagonal, or a closed door (via `canEnterTile`, which already encodes all three). `jumpDown` then steps the unit off the ledge at a constant horizontal speed with no vertical throw and lets gravity plus the commit-1 clamp bring it down on whatever floor it reaches, so a drop of any depth works.

The order validates before it creates the mission, asks for no facing change, and costs nothing, and a cancelled mission no longer reaches the code that spends TU. Brainsucker leaps keep `canLaunch`, their own facing and their 4 TU, so AI behaviour is unchanged.

Not adopted, deliberately: fall damage (in the original, not in OpenApoc, existing `FIXME` at `battleunit.cpp:2914`), and the original's "silently try another movement method when the jump is invalid" (a refused jump just does nothing here).

Two open questions for a maintainer, neither blocking: if railings are wall parts with movement cost 255, the edge test refuses jumping over them where the original vaulted, which is a tileset data question I could not settle. And "down a flight of stairs" is implemented as landing on the first lower surface, i.e. the top step; if it should be the bottom of the flight, that needs a definition of "flight" and is a follow-up.

## Verification

Headless harness (`test_jump_rules`, not part of this PR) that boots a base defence battle, switches it to turn based, searches the generated map for the geometry each case needs, and asserts on landing tile, facing and TU. Same harness both ways.

```
                                              9804cfd9              this branch
F0  3.3 tile fall, 4 tick batches (control)   PASS                  PASS
F1  same fall, 16 tick batches                FAIL {50,71,0}        PASS {50,55,1}
F2a jump off a ledge, 4 tick batches          PASS                  PASS
F2b same jump, 16 tick batches                FAIL {2,86,0}         PASS {1,42,1}
J0  own tile / three tiles away refused       FAIL 4 TU charged     PASS TU unchanged
J1  jump up onto a higher surface             FAIL accepted, 4 TU   PASS refused
J2  jump across a blocked edge                FAIL accepted, 4 TU   PASS refused
J3  jump down to a clear ledge                PASS                  PASS
J5  facing unchanged                          PASS                  PASS
J6  no TU for an accepted jump                FAIL 74 -> 70         PASS 72 -> 72
J7  flying unit refused                       FAIL 4 TU charged     PASS TU unchanged
J4  landing tile occupied                     FAIL landed on it     PASS refused
                                              4 pass / 8 fail       12 pass / 0 fail
```

An intermediate run against commit 1 alone gives 6 pass and 6 fail: the four falling cases fixed, every commit-2 case still showing the old behaviour.

F1 and F2b need 16-tick batches because the landing test is sampled once per update; at 4 ticks the per-update drop stays inside the floor's 0.2 band on this map. F2 also needs a ledge above level 0, since `getTile` truncates a z between -1 and 0 back to level 0 and the unit is caught by that floor anyway.

Full ctest green (RelWithDebInfo, Linux); fork CI Lint + CMake green. No save format change: `launched`, `launchGoal`, `collisionIgnoredTicks`, `jumpTarget` and `jumped` are all already serialised.

#755 looks like a duplicate of this and is worth a re-test once this lands.

### DIFF
--- a/game/state/battle/battleunit.cpp
+++ b/game/state/battle/battleunit.cpp
@@ -4009,6 +4009,84 @@ void BattleUnit::launch(GameState &state, Vec3<float> targetPosition, BodyState 
 	beginBodyStateChange(state, bodyState);
 }
 
+bool BattleUnit::canJumpDown(Vec3<int> target, Vec3<float> &landing)
+{
+	if (!tileObject || isLarge() || canFly() || !isConscious())
+	{
+		return false;
+	}
+	auto &map = tileObject->map;
+	auto *ownTile = tileObject->getOwningTile();
+	Vec3<int> from = ownTile->position;
+	if (target.x == from.x && target.y == from.y)
+	{
+		return false;
+	}
+	if (std::abs(target.x - from.x) > 1 || std::abs(target.y - from.y) > 1)
+	{
+		return false;
+	}
+	if (target.z > from.z || target.z < 0 || !map.tileIsValid({target.x, target.y, from.z}))
+	{
+		return false;
+	}
+	float ownSurface = ownTile->getRestingPosition(false).z;
+	auto *edgeTile = map.getTile(target.x, target.y, from.z);
+	Tile *landingTile = nullptr;
+	for (int z = from.z; z >= 0; z--)
+	{
+		auto *t = map.getTile(target.x, target.y, z);
+		if (!t->getCanStand(false))
+		{
+			continue;
+		}
+		if (t->getRestingPosition(false).z >= ownSurface)
+		{
+			return false;
+		}
+		landingTile = t;
+		break;
+	}
+	if (!landingTile || target.z < landingTile->position.z)
+	{
+		return false;
+	}
+	BattleUnitTileHelper helper(map, *this);
+	if (!helper.canEnterTile(nullptr, landingTile) ||
+	    landingTile->getUnitIfPresent(true, true, false, tileObject))
+	{
+		return false;
+	}
+	BattleUnitTileHelper edgeHelper(map, false, false, true, agent->type->bodyType->maxHeight,
+	                                tileObject);
+	bool jumped = false;
+	bool doorInTheWay = false;
+	float cost = 0.0f;
+	if (!edgeHelper.canEnterTile(ownTile, edgeTile, landingTile != edgeTile, jumped, cost,
+	                             doorInTheWay, true, true, true) ||
+	    doorInTheWay)
+	{
+		return false;
+	}
+	landing = landingTile->getRestingPosition(false);
+	return true;
+}
+
+void BattleUnit::jumpDown(GameState &state, Vec3<float> landing, BodyState bodyState)
+{
+	Vec3<float> targetVectorXY = {landing.x - position.x, landing.y - position.y, 0.0f};
+	if (glm::length(targetVectorXY) == 0.0f)
+	{
+		return;
+	}
+	startFalling(state);
+	launchGoal = landing;
+	launched = true;
+	velocity = glm::normalize(targetVectorXY) * 0.5f * VELOCITY_SCALE_BATTLE;
+	collisionIgnoredTicks = (int)ceilf(36.0f / glm::length(velocity / VELOCITY_SCALE_BATTLE)) + 1;
+	beginBodyStateChange(state, bodyState);
+}
+
 void BattleUnit::startMoving(GameState &state)
 {
 	auto targetMovementMode = missions.empty()
--- a/game/state/battle/battleunit.cpp
+++ b/game/state/battle/battleunit.cpp
@@ -2821,14 +2821,43 @@ void BattleUnit::updateMovementFalling(GameState &state, unsigned int &moveTicks
 		}
 	}
 
+	auto &map = tileObject->map;
+	bool descending = newPosition.z < previousPosition.z;
+	bool landedOnFloor = false;
+	bool inGoalColumn = launched && descending && (int)newPosition.x == (int)launchGoal.x &&
+	                    (int)newPosition.y == (int)launchGoal.y;
+	if (descending && (collisionIgnoredTicks == 0 || inGoalColumn) && newPosition.x >= 0.0f &&
+	    newPosition.x < map.size.x && newPosition.y >= 0.0f && newPosition.y < map.size.y)
+	{
+		int topZ = std::min((int)previousPosition.z, map.size.z - 1);
+		int bottomZ = std::max((int)std::floor(newPosition.z), 0);
+		for (int z = topZ; z >= bottomZ && !landedOnFloor; z--)
+		{
+			auto *t = map.getTile((int)newPosition.x, (int)newPosition.y, z);
+			if (!t->getCanStand(isLarge()))
+			{
+				continue;
+			}
+			float surfaceZ = t->getRestingPosition(isLarge()).z;
+			if (surfaceZ <= previousPosition.z && surfaceZ > newPosition.z)
+			{
+				newPosition.z = surfaceZ;
+				landedOnFloor = true;
+			}
+		}
+	}
+
 	// Fell into a unit
 	if (isConscious())
 	{
-		auto presentUnit =
-		    tileObject->map.getTile(newPosition)->getUnitIfPresent(true, true, false, tileObject);
-		if (presentUnit)
+		auto *newTile = map.getTile(newPosition);
+		if (newTile)
 		{
-			updateFallingIntoUnit(state, *presentUnit->getUnit());
+			auto presentUnit = newTile->getUnitIfPresent(true, true, false, tileObject);
+			if (presentUnit)
+			{
+				updateFallingIntoUnit(state, *presentUnit->getUnit());
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


andremmfaria:

```cpp
// tests/test_jump_rules.cpp (core; difficulty0 loading and geometry search omitted).
// Base defence battle against the player's own base, switched to turn based so spendTU is not
// a no-op. The map (128 x 128 x 5, standing surfaces on levels 0 to 2) is searched per run for
// the geometry each case needs: findFallColumn(), findLedge(minLandingZ), findLedge(wantBlocked)
// and findStepUp(). Every case first resets the unit to KneelingMode::None,
// MovementMode::Walking and BodyState::Standing, since idle kneeling costs 8 TU a time.
                    noVehicle, state->current_base->building);
Battle::enterBattle(*state);
state->current_battle->setMode(Battle::Mode::TurnBased);
state->current_battle->currentActiveOrganisation = state->getPlayer();

void orderJumpLike(GameState &state, sp<BattleUnit> unit, Vec3<int> target, bool requireFacing)
{
	unit->setMission(state, BattleUnitMission::jump(*unit, tileClickPoint(target),
	                                                BodyState::Standing, requireFacing));
}

// F0 / F1: a unit dropped over a floor must land on that floor, at 4 and at 16 tick batches
Vec3<float> start = {floorTile.x + 0.5f, floorTile.y + 0.5f, floorTile.z + dropHeight + 0.5f};
resetUnit(state, unit, start);
unit->falling = true;
for (int i = 0; i < 400 && unit->falling && unit->isConscious(); i++)
	state.current_battle->update(state, batch);
bool pass = !unit->falling && unit->tileObject->getOwningTile()->position == floorTile;

// F2: the reporter's jump, aimed at the base of the tile below the ledge as the UI used to
placeAt(state, unit, ledge.from);
unit->setMission(state, BattleUnitMission::jump(*unit, tileClickPoint(ledge.landing),
                                                BodyState::Standing, false));
settle(state, unit, batch, 2000);
// expect the owning tile == ledge.landing

// J0 / J1 / J2 / J4: every refusal must leave the unit where it was and charge nothing
placeAt(state, unit, from);
int tu0 = tu(unit);
orderJumpLike(state, unit, target, false); // own tile / three columns away / a step up /
settle(state, unit, 4, 500);               // across a blocked edge / onto a parked blocker
bool pass = unit->missions.empty() && !unit->falling &&
            unit->tileObject->getOwningTile()->position == from && tu(unit) == tu0;

// J3 / J5 / J6: an accepted jump lands on the ledge, keeps the facing and costs nothing.
// The unit is first turned away from the target so a facing change would be visible.
unit->setFacing(state, away);
for (int i = 0; i < 20; i++) state.current_battle->update(state, 4); // let idle work settle
Vec2<int> facing0 = unit->facing;
int tu0 = tu(unit);
orderJumpLike(state, unit, ledge.target, false);
settle(state, unit, 4, 2000);
// expect owning tile == ledge.landing, unit->facing == facing0, tu(unit) == tu0
```

Same harness both ways, geometry re-searched per run:

```
                                              9804cfd9              this branch
F0  3.3 tile fall, 4 tick batches (control)   PASS                  PASS
F1  same fall, 16 tick batches                FAIL {50,71,0}        PASS {50,55,1}
F2a jump off a ledge, 4 tick batches          PASS                  PASS
F2b same jump, 16 tick batches                FAIL {2,86,0}         PASS {1,42,1}
J0  own tile / three tiles away refused       FAIL 4 TU charged     PASS TU unchanged
J1  jump up onto a higher surface             FAIL accepted, 4 TU   PASS refused
J2  jump across a blocked edge                FAIL accepted, 4 TU   PASS refused
J3  jump down to a clear ledge                PASS                  PASS
J5  facing unchanged                          PASS                  PASS
J6  no TU for an accepted jump                FAIL 74 -> 70         PASS 72 -> 72
J7  flying unit refused                       FAIL 4 TU charged     PASS TU unchanged
J4  landing tile occupied                     FAIL landed on it     PASS refused
                                              4 pass / 8 fail       12 pass / 0 fail
```

An intermediate run of the same harness against commit 1 alone gives 6 pass and 6 fail: F0, F1, F2a, F2b, J3 and J5, that is the four falling cases fixed and every commit-2 case still showing the old behaviour.

F1 and F2b need 16-tick batches because the landing test is sampled once per update: at 4 ticks the per-update drop stays inside the floor's 0.2 band on this geometry, at 16 it does not. F2 also needs a ledge whose landing is above level 0, since `getTile` truncates a z between -1 and 0 back to level 0 and the unit is caught by that floor anyway. J4 runs last because pre-fix it lands the unit on the blocker and the fall knocks it out for everything after it.

One thing the harness turned up that I could not confirm from reading alone: a refused jump still cost 4 TU (J0 and J7 pre-fix). `start()` sets `cancelled` and `isFinished` and then calls `update()` once more, and the Jump branch never looked at `cancelled`, so it spent the TU and called `launch()`, which then did nothing. Fixed here too.

Full `ctest` is 11/11 with the harness out of the tree, 12/12 with it in. Happy to turn it into a real CTest case as a follow-up.

FilmBoy84:
Nice work

One thing i noticed is we are using "STANDART" instead of "STANDARD" in the phrase "STANDART_MOVE_TU_COST"

Not sure if this an existing typo that this PR perpetuates or there is a reason for it

image

In either case, probably one for another PR

PR #1681, opened 2026-09-12
City: end attack missions when the target leaves the city (#1325)

### LINKED ISSUE #1325: When targeted UFOs return to the alien dimension, friendly ships switch targets to attack random point on map
While attacking an enemy ship, my hoverbikes decided to switch targets onto a random point on the map. This appears to have occurred exactly when I clicked on the "organizations" tab with the map open.

https://youtu.be/BRogQm5x4Iw?t=1146

The random point on the map also had a red "enemy" box, but no enemies listed on the enemy ships tab were at that location. 

https://youtu.be/BRogQm5x4Iw?t=1207

Window 10 x64 22H2
OpenApoc version 0.0.930
no mods in use

[settings.zip](https://github.com/OpenApoc/OpenApoc/files/14152382/settings.zip)

[save_test 3.zip](https://github.com/OpenApoc/OpenApoc/files/14152385/save_test.3.zip)

Here are two save files where the bug is happening, and more video: 

https://youtu.be/-029oEE0ynE?t=3916
https://youtu.be/-029oEE0ynE?t=4041

[save_ships attack random spot.zip](https://github.com/OpenApoc/OpenApoc/files/14237787/save_ships.attack.random.spot.zip)
[save_ships attack random spot 2.zip](https://github.com/OpenApoc/OpenApoc/files/14237791/save_ships.attack.random.spot.2.zip)

andremmfaria:
Fixes #1325.

## Root cause

`VehicleMission::isFinishedInternal` validates an `AttackVehicle` target's existence, tile object, crashed/sliding/falling state and health (`vehiclemission.cpp:1544-1560`), but never checks that the target is still in the same city as the pursuer. The `FollowVehicle` branch immediately below it does exactly that, at `:1569` and `:1575`.

A UFO leaving through a dimension gate never presents the null-`tileObject` window that would end the mission: `Vehicle::enterDimensionGate` (`vehicle.cpp:1318`) removes it from the map and `leaveDimensionGate` (`vehicle.cpp:1273`) re-adds it in the other city, both inside the same synchronous step of that UFO's own update. A pursuer's next check therefore sees a live target with a valid tile object, and keeps flying at the target's alien-city coordinates reinterpreted on the human map. That is the "random empty point".

Both savegames attached to the issue show the end state directly: live `AttackVehicle` missions whose `targetVehicle` sits in `CITYMAP_ALIEN` while the pursuers are in `CITYMAP_HUMAN`, with the stored `targetLocation` X/Y matching the UFO's alien-side position exactly.

## Fix

End the mission when the target is no longer in the pursuer's city, mirroring the existing `FollowVehicle` pattern:

```cpp
if (t->city != v.city)
{
    LogInfo("Vehicle attack mission: Target left the city");
    return true;
}
```

`getNextDestination()` and `setFollowPath()` were checked and deliberately left alone: `Vehicle::getNewGoal()` (`vehicle.cpp:3336`) calls `popFinishedMissions()` immediately before `getNextDestination()` in the same synchronous call, and a gate transition happens inside a different vehicle's update, so there is no tick in which a stale cross-city destination can be consumed.

Not adopted: any replacement mission (fly to the portal, return to base). The issue thread suggests the original did one of those, but the original's `Pursue` / `No pursuit` UI strings have no code references anywhere in `UFO2P.EXE`, and no code path was found that gives a pursuing craft a replacement order when its target leaves. Cancelling the mission appears to be the faithful behaviour.


andremmfaria:
Harness used for the before/after check. It is committed as `tests/test_city_dimension_gate.cpp` and wired into ctest, so it runs in CI rather than being a one-off.

```cpp
// Place pursuer and target in the same city, give the pursuer an AttackVehicle mission.
auto pursuer = cityHuman->placeVehicle(*state, pursuerType, state->getPlayer(), {20, 20, 2}, 0.0f);
auto target  = cityHuman->placeVehicle(*state, targetType, state->getAliens(), {30, 30, 2}, 0.0f);
StateRef<Vehicle> targetRef{state.get(), target};
pursuer->setMission(*state, VehicleMission::attackVehicle(*state, *pursuer, targetRef));

if (pursuer->missions.front().isFinished(*state, *pursuer))
    fail("Mission reported finished before the target ever left the city");

// Move the target to the other city the way a dimension-gate transition does: gone from
// one map, re-added live on the other, in a single atomic step.
target->removeFromMap(*state);
target->city = cityAlien;
cityAlien->map->addObjectToMap(*state, target);

if (!pursuer->missions.front().isFinished(*state, *pursuer))
    fail("Mission still reports unfinished after the target left the city");
if (!pursuer->popFinishedMissions(*state))
    fail("popFinishedMissions() did not pop the stale AttackVehicle mission");
if (!pursuer->missions.empty())
    fail("Pursuer still carries a mission after the phantom AttackVehicle was popped");
```

The relocation is deliberately done at the storage level rather than by calling `enterDimensionGate`/`leaveDimensionGate`, because the point being tested is the state a pursuer observes afterwards: a live target, with a valid `tileObject`, in a different city.

Output on master `aed6f412` (fix reverted, harness unchanged):
```
12: E bool test_attack_mission_ends_when_target_leaves_city(...) Mission still reports unfinished after the target left the city
12: E int main(int, char**) AttackVehicle dimension-gate test failed
1/1 Test #12: test_city_dimension_gate .........***Failed    1.00 sec
```

Output on this branch:
```
1/1 Test #12: test_city_dimension_gate .........   Passed    1.11 sec
```

Full suite on this branch is 12/12, the existing 11 plus this one.

PR #1682, opened 2026-09-12
Battle: force the mission-end recheck on unit drop and retreat (#647)

### LINKED ISSUE #647: Mothership Mission Won't Auto-end
It appears I have killed all of the aliens, yet the mission will not end. I debugged and revealed everything. Nothing is alive. I did not see any errors, though that doesn't mean there weren't any.
[save_Mothership No Finish.zip](https://github.com/OpenApoc/OpenApoc/files/3414732/save_Mothership.No.Finish.zip)

andremmfaria:
Fixes #647.

## Root cause

Not the mechanism suggested in the thread. `BattleUnit::dropDown` calls `checkMissionEnd` unconditionally as its first statement (`battleunit.cpp:4621`) and has done since 69d625f6 (Apr 2017); every death path reaches it through `die()` -> `dropDown()`, including fatal wounds, stun and hazard ticks. Toxin C is not a damage-over-time effect at all: no `effectType`, no `hazardType`, `non_violent=true`, an ordinary instant-hit projectile.

The failure is the guard inside `Battle::checkMissionEnd` (`battle.cpp:2005`):

```cpp
else if (missionEndTimer > 0) { return; }
```

Only 2 of its 5 call sites pass `forceReCheck = true` — the brainsucker-egg hatch (`battleunit.cpp:3907`) and `tryToRiseUp` (`:4616`) — and both were added later as point fixes for this same class of bug (10adf502). `dropDown()` and `retreat()` rely on the invariant that `missionEndTimer` is still zero, which a prior unrelated recheck can violate. The recompute is then swallowed and the win state is never reached.

The attached savegame shows exactly that end state: 25/25 `ORG_ALIEN` units `Dead`, all 12 `ORG_X-COM` units alive, `playerWon=true`, and `missionEndTimer` absent.

## Fix

Force the recheck at the two unforced call sites, matching the existing pattern:

```
dropDown():  checkMissionEnd(state, false)  ->  checkMissionEnd(state, false, true)
retreat():   checkMissionEnd(state, true)   ->  checkMissionEnd(state, true,  true)
```

Forcing only removes an early-return short-circuit, so it can make `checkMissionEnd` recompute more often but cannot change what it computes.

Not adopted: the `destroy=true` branch of `die()` (`battleunit.cpp:4992-5007`, violent death with Popper equipment) skips `dropDown()` entirely and therefore skips the mission-end check. That is a real and separate gap, but it is causally unrelated to this report — Toxin C is `non_violent`, so that branch is unreachable from this kill — and it deserves its own issue and test.

The exact turn ordering that leaves `missionEndTimer` non-zero before the final kill could not be reconstructed: only a final snapshot save exists, with no replay or turn log. The mechanism is confirmed reachable and the fix is verified against the real save, but the precise triggering sequence remains unproven.


andremmfaria:
Harness used for the before/after check. It is committed as `tests/test_mission_end_recheck.cpp`, deliberately not wired into ctest because it needs the savegame attached to #647 rather than generated state; it is run by hand.

Two cases, each on a freshly loaded copy of that save so they cannot interfere:

```cpp
// The reported scenario: missionEndTimer is already nonzero, as if a prior unrelated
// recheck left it running, and playerWon is stale. Then the true last hostile dies.
void runGuardedTimerCase(GameState &state)
{
    auto &battle = *state.current_battle;
    auto unit = findAlienUnit(state);
    check((bool)unit, "found an alien unit in the save to use as the last hostile");

    reviveUnit(*unit);
    battle.missionEndTimer = 7; // simulate a prior, unrelated recheck already running
    battle.playerWon = false;   // stale conclusion from that prior recheck

    unit->die(state, nullptr, false); // non-violent kill, matches a Toxin C hit

    check(battle.missionEndTimer > 0, "missionEndTimer > 0 after the kill");
    check(battle.playerWon == true, "playerWon == true after the kill");
}

// Control: timer starts at 0, no interfering prior recheck. Must keep working.
void runPlainCase(GameState &state)
{
    /* ... same, but battle.missionEndTimer = 0 ... */
}
```

`die(state, nullptr, false)` is the non-violent path, which is what a Toxin C hit actually takes, so the Popper `destroy=true` branch is never entered here.

Output on master `aed6f412` (fix reverted, harness unchanged):
```
Loaded save. missionEndTimer=0 playerWon=1
Case: last kill with pre-seeded missionEndTimer
  PASS: found an alien unit in the save to use as the last hostile
  PASS: missionEndTimer > 0 after the kill
  FAIL: playerWon == true after the kill
Case: last kill with missionEndTimer starting at 0
  PASS: found an alien unit in the save to use as the last hostile
  PASS: missionEndTimer > 0 after the kill
  PASS: playerWon == true after the kill
SOME CHECKS FAILED
```

Output on this branch:
```
Loaded save. missionEndTimer=0 playerWon=1
Case: last kill with pre-seeded missionEndTimer
  PASS: found an alien unit in the save to use as the last hostile
  PASS: missionEndTimer > 0 after the kill
  PASS: playerWon == true after the kill
Case: last kill with missionEndTimer starting at 0
  PASS: found an alien unit in the save to use as the last hostile
  PASS: missionEndTimer > 0 after the kill
  PASS: playerWon == true after the kill
ALL CHECKS PASSED
```

Note which assertion moves: `missionEndTimer > 0` passes either way, because `dropDown()` does reach `checkMissionEnd` and the timer is already running. It is `playerWon` that stays stale on master, because the guard returns before the win state is recomputed. That is the bug, and it is why the timer being non-zero beforehand is the necessary precondition.

The plain case passes on master too, which is the point of including it: the unguarded path was never broken, so the fix must not change it.

`ctest` is 11/11 on this branch, unchanged from master.

Issue #601, opened 2019-06-13
Adding agents to a squad during battlescape.


Quickmind01:
I confirmed in the Steam version that selecting agents and then pressing Ctrl + (the squad number) moved those selected agents to that squad. This may have been a shortcut way of doing it, bypassing the need for clicking the "move to squad button." Could this be implemented in OpenAPOC? I understand this is probably a low priority.

makus82:
Its can be next hotkeys like 
CTRL + choose any agents and add to selection
ALT and click at already selected agent will substract it from selection
And then CTRL 1-6 add selection to selected squad slot
and if some squad already selected for ex 1 sqd and then you press CTRL +2 you add all 1 sqd units to 2sqd

also add summ spotted aliens dots above sqds like this
![sqdspots](https://user-images.githubusercontent.com/11581207/59433227-ba13b800-8df1-11e9-90c1-8c47f115dccc.png)

PR #1683, opened 2026-09-12
Battle: assign the current selection to a squad with ctrl+1-6 (#601)

andremmfaria:
Refs #601.

Three of the eight requests in the issue already work on master and need no code:

- Ctrl+click to add an agent to the selection: `battleview.cpp:3804`.
- Mousewheel to change map level: `BattleView::handleMouseScroll` (`battleview.cpp:3561-3580`), gated by `OpenApoc.NewFeature.BattlescapeVertScroll`, default on. It was undocumented, and this PR adds it to `README_HOTKEYS.md`.
- Numpad Enter opening the inventory: `SDLK_RETURN` and `SDLK_KP_ENTER` already fall through together at `battleview.cpp:3478`.

## What this adds

Ctrl+1 through Ctrl+6 assign the current selection to squad N, wiring `BattleForces::insert` (`battleforces.cpp:10-15`) — until now reached only from the pre-battle squad screen — into a new branch of the existing number-key handler.

Over-capacity behaviour matches the precedent already in this file rather than inventing new UX: the squad-assign-mode path around `battleview.cpp:605-633` loops the selection and checks `getNumUnits() < 6` per unit, silently skipping those that do not fit. The new branch does the same. `insertAt` already calls `removeAt` on the unit's previous squad, so no second removal path was added.

The new branch sits after the existing Shift test, so Shift+1-6 (select unit N within the squad) and Ctrl+Shift+1-6 (additive select) are unchanged and still resolve before it.

## Not adopted

The remaining four requests need design decisions rather than code:

- Alt+click to subtract from the selection: right-click already removes, and Alt+LMB is already force-fire-at-ground in battlescape.
- Ctrl+2 merging squads 1 and 2 into the selection: needs a decision on whether Ctrl+N means "add squad N" or "assign to squad N", which cannot be both.
- Shift+2 moving squad 1's units into squad 2: directly conflicts with the existing Shift+1-6 binding.
- Summed spotted-alien counts over the on-map squad markers: a HUD rendering request, not a hotkey. The underlying `spottedInfo` data already exists and is already surfaced in the portrait bar (`battleview.cpp:4549-4557`).

Worth noting for anyone checking against the wiki: the vanilla game had no numbered squad keys, no Ctrl/Shift+number combos and no mousewheel level switch. Ufopaedia's Input Device Reference for Apocalypse is exhaustive and silent on all three, so these are OpenApoc additions and there is no original binding to be faithful to.

Testing: the state-layer behaviour (assignment, the 6-unit cap as a clean no-op, consistent removal from the previous squad) is covered by a new test. The key dispatch itself is verified by inspection only — there is no harness anywhere in `tests/` that simulates SDL key events against a UI class.


andremmfaria:
Harness used for the before/after check, committed as `tests/test_squad_assignment.cpp` and wired into ctest.

A caveat worth stating up front: this tests the state-layer behaviour the new hotkey relies on, not the key dispatch itself. There is no harness anywhere in `tests/` that simulates SDL key events against a UI class, so the `SDLK_1`..`SDLK_6` branch placement was verified by inspection only. Reverting `battleview.cpp` therefore does not make this test fail, and presenting that as a before/after would be misleading.

What the test does cover is the two things the feature can get wrong: the 6-unit cap, and consistent removal from the previous squad.

```cpp
static bool test_full_squad_is_noop(BattleForces &forces)
{
    std::vector<sp<BattleUnit>> units;
    for (int i = 0; i < 6; i++)
    {
        auto unit = makeUnit();
        if (!forces.insert(0, unit))
            fail("Unit {0} failed to join squad 0 while it had free space", i);
        units.push_back(unit);
    }
    if (forces.squads[0].getNumUnits() != 6)
        fail("Squad 0 should have 6 units, has {0}", forces.squads[0].getNumUnits());

    auto overflow = makeUnit();
    bool assigned = forces.insert(0, overflow);
    if (assigned)
        fail("Assigning a 7th unit to a full squad should fail, but succeeded");
    /* ... also asserts the rejected unit's squadNumber stays -1 and the
           existing six units and their positions are undisturbed ... */
}
```

To show the test actually bites, the "before" here is a deliberate mutation rather than a revert: the capacity guard (`getNumUnits() == 6`) broken to `== 999`. It has to be broken in **both** `BattleForces::insert` and `BattleForces::insertAt`, because the guard is duplicated in the two functions and breaking only one leaves the test passing, which would have looked like a false positive.

Output with the capacity guard mutated in both functions:
```
10: E bool test_full_squad_is_noop(OpenApoc::BattleForces&) Assigning a 7th unit to a full squad should fail, but succeeded
10: E int main(int, char**) Full squad no-op test failed
1/1 Test #10: test_squad_assignment ............***Failed    0.01 sec
```

Output unmutated, on this branch:
```
1/1 Test #10: test_squad_assignment ............   Passed    0.03 sec
```

Full suite on this branch is 12/12, the existing 11 plus this one.

Issue #989, opened 2021-02-08
①My Retiliator 4 stuck and　② I cant fast forward (x5)


endy2qq:
①My Retiliator 4 stuck and　② I cant fast forward (x5)
[save_city2.zip](https://github.com/OpenApoc/OpenApoc/files/5944097/save_city2.zip)

Please check my savefile. 

sorry for my terrible English.

Thank you

FilmBoy84:
Another instance of a stuck vehicle from SAM at discord

To replicate, send all craft home, the second hovercar will refuse to do so, even under manual control (N Key)

[save_TACTICAL.zip](https://github.com/OpenApoc/OpenApoc/files/7875459/save_TACTICAL.zip)

The main difference with this example is that ultrafast time can be used - though the hovercar remains stuck

JaySwizzle:
I am experiencing this issue now also. I have a hover car stuck but am still able to fast forward. I'm new here, so catching up on protocol, but if there's something I can do to help or a save / log I could provide to assist, just let me know!

FilmBoy84:
> ...if there's something I can do to help or a save / log I could provide to assist, just let me know!

Thanks for the update that this is still a problem Jay

As always, if you can upload a save with a record of what's stuck and where, and how (if known) to replicate it, that'd be very welcome, thanks :beers:

JaySwizzle:
Attached you will find the save file and a txt file with a quick w
rite-up. I'm around if anything else I can provide would help!
[saves.zip](https://github.com/OpenApoc/OpenApoc/files/9228004/saves.zip)

dethstryke01:
This vehicle got stuck in a building and prevented the fastest game speed. I never told it to move there. 

I first noticed that the fastest game speed was impossible here: 
https://youtu.be/3F5is1O6n0U?t=5785

And then after reloading a save I managed to order the vehicle to move out of the building. It slowly returned to base, and then the fastest speed became possible again: 
https://youtu.be/XSoMAQXj19M?t=5

Here's the save:

[save_103.zip](https://github.com/OpenApoc/OpenApoc/files/14323255/save_103.zip)
[settings.zip](https://github.com/OpenApoc/OpenApoc/files/14323257/settings.zip)

andremmfaria:
[...]

PR #1684, opened 2026-09-12
City: dispatch every idle rescue craft per organisation update (#1008)

### LINKED ISSUE #983: Crashed Vehicle is not recovered/destroyed in Cityscape preventing fast passing of time. Vehicles also become stuck
Hello. 

I can't fast forward after UFO Fight.

Please check my save file.

Sorry about my terrible English ( I dont speak English well .....) 

[save_city.zip](https://github.com/OpenApoc/OpenApoc/files/5936811/save_city.zip)

### LINKED ISSUE #1008: Crashed Vehicles Are Not Recovered at Ultra Speed
When playing OpenApoc with vehicle crashes and recovery enabled, using ultra speed will cause the vehicle to be destroyed, not recovered after a few hours. This includes X-COM craft; that currently appear to sit there and then explode when they should be picked up and returned for repair automatically.

Additionally, mission generation for vehicle recovery by the city seems flaky at best; in earlier versions of OpenApoc, if a vehicle crashed, a Construction Vehicle was quickly sent to recover the unit and bring it back to it's parent organisation (usually point of spawn in the case of organisation, base of origin in the case of X-COM). This no longer seems to happen immediately and sometimes not at all. I'm trying to work out what's bugged the mechanism but need further data.

andremmfaria:
Fixes #1008.
Refs #983.

## Root cause

`Organisation::updateMissions` dispatches at most **one** rescue mission per organisation per call (`organisation.cpp:399-513`): it picks the first idle `canRescueCrashed` craft and breaks, then assigns the first eligible victim and breaks again. There is no loop over the remaining idle craft or the remaining wrecks.

`GameState::update()` runs once per rendered frame at every game speed (`UPDATE_EVERY_TICK = false`, `gamestate.h:55`; `CityView::update` issues a single `state->update(ticks)` with the whole batch, `cityview.cpp:2121-2126`, and `updateTurbo` does the same, `gamestate.cpp:1707-1719`). So the number of dispatch opportunities is frame-rate bound and roughly constant, while the game time each one covers scales with speed: on the order of 12 chances per game-hour at Ultra against thousands at speeds 1-4.

Meanwhile each crashed vehicle's `SelfDestruct` timer is independent and is not paused when a rescue is assigned (`vehiclemission.cpp:521-534`, decremented at `:1484-1491`); only an actual pickup clears it. With a backlog of several wrecks and one or two rescue craft, the timers win and the vehicles self-destruct instead of being recovered. A single crash dispatches on the very first call and recovers fine even at Ultra, which matches the report that it worked once.

Not the mechanism previously suspected: `updateMissions` is called on every update at every speed, it is never skipped.

## Fix

Dispatch every idle rescue-capable craft the organisation owns per call, instead of stopping after the first, with each victim claimed at most once.

Deliberately out of scope: the second, unrelated half of #983 ("preventing fast passing of time"). That is `GameState::canTurbo()` (`gamestate.cpp:950-977`) returning false for as long as any vehicle holds a pending `AttackBuilding` / `AttackVehicle` mission, with no timeout, so a permanently unreachable target locks the game out of Ultra. Different file, different mechanism, and it needs its own issue and test. This PR does not close #983 on its own.


andremmfaria:
To be explicit about the scope boundary: this PR closes the recovery-throughput half of #983 (which is also #1008) and deliberately does not address the other symptom in that report, "preventing fast passing of time".

That second symptom is a different mechanism in a different file, `GameState::canTurbo()` at `gamestate.cpp:950-977`, and it is already tracked at #989. I have posted the mechanism there: the gate has three independent blockers (any projectile in flight, any live hostile aggressive vehicle in the city, any pending AttackBuilding/AttackVehicle mission), none of which has a timeout or a reachability check.

andremmfaria:
Harness for the before/after check, committed as `tests/test_crashed_vehicle_recovery.cpp`. It is deliberately not wired into ctest: it is a comparative measurement rather than a pass/fail regression gate, so it is run by hand with the fleet size as arguments.

The shape is to drive the same total game-time window two ways and compare outcomes:

```cpp
// One organisation owning N rescue-capable craft, M vehicles forced to crashed at once.
// Same one-hour window driven two ways:
//   (a) via updateTurbo(), i.e. the fastest speed, few calls each covering a lot of game time
//   (b) via update(1) at normal cadence, many calls each covering little
// Then compare how many were recovered versus self-destructed on each path.
```

The scale matters, and it is worth being explicit about it: at the default 2 rescuers / 8 wrecks, **both paths recover everything and there is no difference to see**, because even the old one-dispatch-per-call code gets enough opportunities at that fleet size. The bug needs a backlog large enough that dispatch throughput, rather than rescue-craft availability, becomes the limit.

Output on master `aed6f412` (fix reverted), 4 rescuers / 16 wrecks:
```
=== Scenario: turbo path, 4 rescuers, 16 victims, 1h window ===
turbo: 13 updateTurbo() calls, 561600 ticks elapsed, recovered=4 selfDestructed=12 unresolved=0 aborted=false
=== Scenario: normal-cadence path, 4 rescuers, 16 victims, 1h window ===
normal: 518401 update(1) calls, recovered=16 selfDestructed=0 unresolved=0
RESULT turbo_recovered=4 turbo_selfDestructed=12 normal_recovered=16 normal_selfDestructed=0
MISMATCH: turbo and normal-cadence recovery rates differ
```

Output on this branch:
```
=== Scenario: turbo path, 4 rescuers, 16 victims, 1h window ===
turbo: 13 updateTurbo() calls, 561600 ticks elapsed, recovered=16 selfDestructed=0 unresolved=0 aborted=false
=== Scenario: normal-cadence path, 4 rescuers, 16 victims, 1h window ===
normal: 518401 update(1) calls, recovered=16 selfDestructed=0 unresolved=0
RESULT turbo_recovered=16 turbo_selfDestructed=0 normal_recovered=16 normal_selfDestructed=0
MATCH: turbo and normal-cadence recovery rates are equal
```

13 `updateTurbo()` calls against 518401 `update(1)` calls over the same simulated hour is the whole problem in one line: dispatch opportunities are bound to call count, not to game time, so at the fastest speed the old code had 13 chances to place 16 rescues while the 1-hour self-destruct timers ran to completion regardless.

`ctest` is 11/11 on this branch, unchanged from master.

Issue #1457, opened 2024-05-23
Megapol Hostile yet Allied


XCOM-HUB:
I got the following message when I tried to buy equipment from Megapol that I am allied with (see pictures below):
![Megapol Hostile yet Allied](https://github.com/OpenApoc/OpenApoc/assets/55395993/3da89891-40f1-4785-a73f-cea0fa67e107)
![Megapol is allied with X-COM](https://github.com/OpenApoc/OpenApoc/assets/55395993/26d3177c-8c4d-4e7f-952d-ae223f438922)

SAVE-FILE: [Megapol Allied yet Hostile.zip](https://github.com/OpenApoc/OpenApoc/files/15422871/Megapol.Allied.yet.Hostile.zip)

Kurtsley:
Are you hostile with transtellar? They would be making the actual delivery.

XCOM-HUB:
> Are you hostile with transtellar? They would be making the actual delivery.

They have been hostile with me but I have paid them to be "Friendly" with me. I guess paying orgs doesn't work!?

Kurtsley:
Hmm. I'll check the save. Was the game unpaused after you paid them? I'm not sure if that updates immediately.

XCOM-HUB:
> Hmm. I'll check the save. Was the game unpaused after you paid them? I'm not sure if that updates immediately.

I think it was paused, because I prefer to pause the game when I do stuff like that.

FilmBoy84:
> I got the following message when I tried to buy equipment from Megapol that I am allied with (see pictures below): 

This is quite common as MegaPol has issues with Transtellar - The message refers to their relations not yours to MegPol

This happens in the Orignal Game though is slightly rarer as each organisation has a secondary relationship value in each two way relationship that creates a stabilising bias - This is not implemented in OpenApoc as yet

Thus, without the stabilisation MegaPol and transtellar quickly fall out over the slightest thing and you cannot buy from MegaPol any longer

FilmBoy84:
Probably worth mentioning that if the option to allow X-COM to transport cargo is enabled, you can click proceed to the above message and then send a craft with a cargo module to the building in a yellow circle on the cityscape map to pick up the goods yourself

You may need to make multiple trips (until the yellow circle disappears) but it means you can still get the order even if Transtellar is not playing ball

Kurtsley:
[...]

PR #1685, opened 2026-09-12
Organisations: add the original's ally ripple and event-gated long-term relations (#1457)

andremmfaria:
Refs #1457.

Three commits, reviewable separately. Only the third changes player-visible balance.

## What the issue asks for does not exist

The report describes a "second value" per organisation relation acting as a slow-moving stabilising bias. There is no such formula in the original game to port.

`UFO2P.EXE` holds two 28x28 signed-byte matrices, current relations at `0x0016ec28` and a long-term value at `0x0016f078`, row-major, stride 28. Across every instruction referencing those addresses (~90 sites in ~25 functions), **no instruction reads the long-term matrix to compute a new current value**. Its only writes are a one-time bulk copy at load (`FUN_00014854`) and a handful of narrative-event handlers that set `bias = current`. Its only reads are two UI text builders (`FUN_0009379c`, `FUN_00093a30`) that classify it into Allied/Friendly tiers beside the current value to build a two-part label — which is what the Midnight Editor screenshot in the thread is showing. There is no periodic sweep nudging either value toward the other.

So OpenApoc's existing `updateRelations` stub is closer to the original's intent than the requested feature would be. What it gets wrong is only that it re-baselines every day instead of on qualifying events.

## What is actually missing

The original's real adjuster, `FUN_0005faf0`, applies an **ally ripple** that OpenApoc does not implement: every relation change propagates to third-party organisations in proportion to how much each one likes the affected party, with extra damping when it dislikes them, dropped rather than clamped when the result would leave range.

OpenApoc already approximates this by hand at four call sites. Their constants are the original's formula sampled at tier boundaries:

| case | original formula | OpenApoc constant |
|---|---|---|
| `vehicle.cpp` allied, delta -30 | `(100/2 * -30)/100 = -15` | -15 |
| `vehicle.cpp` rel > 20 | -3 | -3 |
| `scenery.cpp` allied, delta -5 | -2.5 | -2.5 |

Because those loops call `adjustRelationTo` per organisation, adding the ripple inside that function without a gate would fire N extra ripple passes per event and produce second-order propagation the original never does — it writes into the matrix directly and does not recurse.

## The three commits

1. **`event-gate long-term relations`** — `long_term_relations` re-baselines only on qualifying events via a new `establishRelation`, not every day. A new `previous_relations` field preserves the existing day-over-day `playerRelationshipDelta` used for diplomacy triggers, so that behaviour is unchanged; the raid heuristic at `organisation.cpp:342` needed no change and now measures drift from an established baseline rather than yesterday's dip.
2. **`add opt-in reputation ripple and unconditional player mirror`** — the ripple behind a new `applyReputationRipple` parameter, defaulting false so every existing caller is untouched, writing `current_relations` directly, dropping out-of-range results, and excluding the victim from its own loop. The X-COM mirror and the X-COM/Aliens -100 pin move out from behind the tactical-battle guard, matching the original, which runs them outside it.
3. **`convert vehicle/scenery relation hooks to the shared ripple`** — deletes the hand-rolled loops in `Vehicle::adjustRelationshipOnDowned` and `Scenery::updateRelationWithAttacker` and passes `true` instead.

## Balance effect of commit 3, measured

Downing a craft, delta -30, third party's reaction by its relation to the victim:

| relation to victim | before | after |
|---|---|---|
| 74 (Friendly) | -3 | **-11.1** |
| 50 (Friendly) | -3 | **-7.5** |
| 0 (Neutral) | 0 | 0 |
| -100 (Hostile) | +5 | **+7.5** |
| the victim itself | **-45** | **-30** |

The Friendly band is where the tiers diverge hardest from the curve: everything between 25 and 74 currently collapses to a flat -3. The victim's own penalty changes because it is not currently excluded from its own loop, and `getRelationTo(self)` returns 100, so it takes an extra -15 as its own ally; the original excludes it.

Also surfaced by the conversion: `vehicle.cpp:2060` (`else if (rel < -20.0f)`) is dead code, unreachable behind `rel < -10.0f` at `:2056`.

The two `battleunit.cpp` sites are deliberately left alone. Their constants are about 3x off the original's line, and they run during tactical battle, where the original suppresses the ripple entirely.

Float arithmetic is used rather than reproducing the original's integer truncation, since `current_relations` is already `float` and existing callers pass values like `1.5f`. The behaviourally significant quirk — dropping an out-of-range ripple instead of clamping it — is reproduced exactly.


andremmfaria:
Harness for the before/after check, committed as part of `tests/test_org_relations.cpp` and wired into ctest.

A note on why this particular test exists. The first eight sub-tests in that file exercise `Organisation::adjustRelationTo` directly on hand-built organisations, which covers the ripple mechanism added by the second commit but does **not** touch the third commit at all: that one only changes `Vehicle::adjustRelationshipOnDowned` and `Scenery::updateRelationWithAttacker`. Checking those two files out at the previous commit and rebuilding produced byte-identical test output. Since the third commit is the only one that changes player-visible balance, it needed a test that actually drives the call site.

```cpp
static bool test_vehicle_downed_relation_adjustment(sp<GameState> state)
{
    auto victimOrg   = state->getOrganisation("ORG_GOVERNMENT");
    auto attackerOrg = state->getOrganisation("ORG_MEGAPOL");
    auto thirdHigh   = state->getOrganisation("ORG_MARSEC");      // relation 74 to victim
    auto thirdMid    = state->getOrganisation("ORG_CYBERWEB");    // relation 50
    auto thirdZero   = state->getOrganisation("ORG_DIABLO");      // relation 0
    auto thirdNeg    = state->getOrganisation("ORG_TRANSTELLAR"); // relation -100

    state->current_battle = nullptr;

    // Not hostile, so the "lose 30 points" branch fires rather than the -5 one.
    victimOrg->current_relations[attackerOrg] = 20.0f;
    attackerOrg->current_relations[victimOrg] = 20.0f;
    thirdHigh->current_relations[victimOrg] = 74.0f;  /* ... etc, attacker seeded at 0 ... */

    // Real Vehicle objects registered in state->vehicles so StateRef<Vehicle> resolves
    // normally, then the actual call site is invoked:
    victimVehicle->adjustRelationshipOnDowned(*state, attackerRef);

    // third parties, by their relation to the victim
    expect(nearlyEqual(thirdHigh->current_relations[attackerOrg], -11.1f));
    expect(nearlyEqual(thirdMid ->current_relations[attackerOrg],  -7.5f));
    expect(nearlyEqual(thirdZero->current_relations[attackerOrg],   0.0f));
    expect(nearlyEqual(thirdNeg ->current_relations[attackerOrg],  +7.5f));
    // victim's own penalty: 20 -> -10, i.e. -30 rather than the previous -45
    expect(nearlyEqual(victimOrg->current_relations[attackerOrg], -10.0f));
}
```

Output with `game/state/city/vehicle.cpp` and `game/state/city/scenery.cpp` reverted to the commit before the conversion:
```
12: E bool test_vehicle_downed_relation_adjustment(...) Expected a third party at relation 74 to the victim to move -11.1 toward the attacker, got -3
1/1 Test #12: test_org_relations ...............***Failed    1.83 sec
```

`got -3` is the old hand-rolled tier value: everything in the Friendly band from 25 to 74 collapsed to a flat -3, which is where the tier table diverges furthest from the original's continuous curve.

Output on this branch:
```
1/1 Test #12: test_org_relations ...............   Passed
100% tests passed, 0 tests failed out of 12
```

Full suite is 12/12, the existing 11 plus this one.

The victim's own penalty assertion is the one most worth reviewing: it moves from -45 to -30 because the victim was previously not excluded from its own ripple loop, and `getRelationTo(self)` returns 100, so it took an extra -15 as its own "ally". The original excludes the row being adjusted.

FilmBoy84:
Wow! Nice to finally have this implemented
Tested over the course of 4 weeks game time (to April 2084)

Things seem much more stable in the way Orgs work 🍻

PR #1686, opened 2026-09-12
Battlescape: optional gravlift line of fire and sight (#1330)

andremmfaria:
Refs #1330.

**This is a deliberate deviation from the original game, behind an opt-in flag. It is not a bug fix and not a restoration of original behaviour.** Three commits, reviewable separately.

## What the original does

The original blocks these shots too. `TACP.EXE`'s two voxel routines — `FUN_0005bad8` (line-of-sight raycast) and `FUN_0005d25c` (projectile trajectory stepper) — each read exactly one per-tile field per map-part category, offset `+0xf`, `loftemps_lof`. Neither reads `is_floor` (0x47) or `is_gravlift` (0x48). There is no grav-lift special case anywhere in the original's voxel chain, and a grav-lift floor pad assigns LOFTEMPS index 6 to z-slices 0 and 1, which decodes to a fully solid 24x24 plane.

(The original's routines that *do* exempt grav-lifts, `FUN_0004fc38` and `FUN_0004fd1c`, are a movement/reachability constraint with no call edges to or from the voxel chain.)

## Why change it anyway

The original exempts grav-lift floors for **movement** but not for **line of fire**, so a unit can float up through a plate it cannot shoot or see through. That asymmetry is the wart this addresses. OpenApoc already has 44 `OpenApoc.NewFeature.*` options for exactly this kind of deliberate improvement, so this follows that pattern rather than changing behaviour unconditionally.

## The three commits

1. **`add GravliftLineOfFire NewFeature option`** — the option alone, default on, listed in the options UI. No behaviour change.
2. **`exempt gravlift floor pads from line of fire under flag`** — `TileObjectBattleMapPart::getVoxelMap` returns no voxel map for `floor && gravlift` parts, for both LOF and LOS, when the flag is set. With the flag off the behaviour is byte-identical to today.
3. **`add gravlift line-of-fire regression test and manual battle harnesses`**.

## AI behaviour

Exempting LOS feeds `BattleUnit::calculateVisionToUnit` and therefore `visibleUnits`, which drives AI target selection and reaction fire, so the obvious risk was an AI fixating on a unit it can see through a shaft but cannot reach. A harness boots a real battle, places a hostile-organisation unit and a player unit at opposite ends of a grav-lift shaft and drives `Battle::update()` (which calls `aiBlock.think()` every tick) for 8640 ticks:

- Flag off: no line, no engagement decision, both units idle for the whole run.
- Flag on: visibility goes to 1 immediately, exactly **one** attack decision fires at tick ~1, then zero further decisions across the remaining ~8600 ticks. Per-batch wall-clock time fell to 0.29x the first batch rather than growing. Both units alive, positions stable.
- Control, `07corphq` (no grav-lift floor pad): identical decision, displacement and timing with the flag on and off.

## Scope, stated plainly

This only clears shots where a grav-lift **floor pad** is the blocker. In `12shops` and `06office` the same shaft also contains a non-floor grav-lift-flagged piece (a right wall or a frame) which continues to block either way, and `07corphq` has no grav-lift floor pad at all so the flag does nothing there. Those pieces are geometry the original genuinely consults, and exempting them is not proposed.

Sub-problem C of the issue (grenades and units falling off the end of the world down a lift) was already fixed by #1670 and #1676. Sub-problem A (shooting between adjacent lift tiles at the same level) was not reproducible at real muzzle heights in any tileset tested: the only blocking band sits at z 0.00-0.25, well below kneeling (0.55) and standing (0.80).

### DIFF
--- a/framework/options.cpp
+++ b/framework/options.cpp
@@ -346,6 +346,9 @@ ConfigOptionBool optionNoScrollSounds("OpenApoc.NewFeature", "NoScrollSounds",
                                       tr("Disable scrolling sounds"), false);
 ConfigOptionBool optionNoInstantThrows("OpenApoc.NewFeature", "NoInstantThrows",
                                        tr("Throwing requires proper facing and pose"), true);
+ConfigOptionBool optionGravliftLineOfFire("OpenApoc.NewFeature", "GravliftLineOfFire",
+                                          tr("Gravlift floors don't block line of fire/sight"),
+                                          true);
 ConfigOptionBool optionFerryChecksRelationshipWhenBuying(
     "OpenApoc.NewFeature", "FerryChecksRelationshipWhenBuying",
     tr("Transtellar checks relationship when buying items"), true);
--- a/framework/options.h
+++ b/framework/options.h
@@ -80,6 +80,7 @@ extern ConfigOptionBool optionInstantExplosionDamage;
 extern ConfigOptionBool optionGravliftSounds;
 extern ConfigOptionBool optionNoScrollSounds;
 extern ConfigOptionBool optionNoInstantThrows;
+extern ConfigOptionBool optionGravliftLineOfFire;
 extern ConfigOptionBool optionFerryChecksRelationshipWhenBuying;
 extern ConfigOptionBool optionAllowManualCityTeleporters;
 extern ConfigOptionBool optionAllowManualCargoFerry;
--- a/game/state/tilemap/tileobject_battlemappart.cpp
+++ b/game/state/tilemap/tileobject_battlemappart.cpp
@@ -1,4 +1,5 @@
 #include "game/state/tilemap/tileobject_battlemappart.h"
+#include "framework/configfile.h"
 #include "framework/image.h"
 #include "framework/renderer.h"
 #include "game/state/battle/battlemappart.h"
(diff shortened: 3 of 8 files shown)


andremmfaria:
Harnesses for the before/after checks. Both are committed: `tests/test_gravlift_lineoffire.cpp` is a registered ctest, `tests/test_gravlift_ai_battle.cpp` is a manual harness.

## Line of fire

The ctest asserts behaviour at both flag settings within a single run, setting the option itself rather than reading a command-line value:

```cpp
config().set("OpenApoc.NewFeature.GravliftLineOfFire", false);
// ... assert the vertical shot through the grav-lift floor pad IS blocked ...
config().set("OpenApoc.NewFeature.GravliftLineOfFire", true);
// ... assert it is clear, swept across kneeling z=0.55 and standing z=0.80 and
//     across positions within the lift column ...
```

so it is a regression gate for both states rather than a manual A/B. `1/1 passed`; full suite 12/12.

## AI behaviour, and a correction

Exempting line of sight feeds `BattleUnit::calculateVisionToUnit` and therefore `visibleUnits`, which drives AI target selection and reaction fire, so the risk worth testing was an AI fixating on a unit it can see through a shaft but cannot reach. The harness boots a real battle, places a hostile-organisation unit and a player unit at opposite ends of a grav-lift shaft, and drives `Battle::update()` (which calls `aiBlock.think()` every tick).

The first version of this comparison was **not controlled**, and the fourth commit on this branch fixes that. `GameState::rng` is not part of a save, and `OpenApoc.NewFeature.SeedRng` defaults to true, so `GameState::startGame()` reseeds from `std::time(nullptr)` on every process run. `BattleMap::generateMap` consumes that same rng for vertical stacking and sector packing, so two separate invocations produced different map geometry (3 z-levels against 6) purely from wall-clock time. The flag was not the only variable. The harness now takes an optional seed argument and pins the rng:

```cpp
config().set("OpenApoc.NewFeature.SeedRng", false);
state->rng.seed(rngSeed);   // default 424242, overridable as a 4th positional arg
```

Determinism was then checked before drawing any conclusion: the same arm run twice with the same seed produced byte-identical output across all 240 batches once the wall-clock column was stripped.

Controlled result, same seed, 240 batches x 144 ticks = 34,560 ticks per arm. Map geometry now matches between arms ({32,32,3} both). Exactly three differences:

```
flag off: 0 AI attack decisions, visible on 0 of 49 samples
flag on:  1 AI attack decision at batch 0 ("decided to Action: [Attack BATTLEUNIT_10...]"),
          visible on 1 of 49 samples
```

Everything else is identical between the arms: positions, mission state (`idle` throughout in both), target (`(none)` in both), hostile net displacement (0.583 tiles in both), and final state (both units alive). Per-batch timing fell in both arms rather than growing (0.27x flag off, 0.44x flag on), so there is no re-decision spam and no runaway.

So the AI sees the unit through the shaft, makes a single engagement decision, and does not thrash. The conclusion is the same one the uncontrolled run suggested, but it is now measured under a controlled comparison rather than inferred across two different maps.

FilmBoy84:
Thanks for this, ive ran a few battles in UFOs and Buildings with and without the option checked to no issue. Merging

Issue #968, opened 2020-12-24
Bug (vehicle stuck)


cTacuk7:
dont know how to replicate this issue, but i think it happens after ufo attack on the city

[save_sts.zip](https://github.com/OpenApoc/OpenApoc/files/5738506/save_sts.zip)

EmperorLol:
Adding another save for checking. I've just had the same bug. Happened during an attack but from what I can see it happens when a UFO disappears back into the dimension gate. This makes me think the bug is connected to having an impossible to complete order (attack a disappeared ufo).

[save_Overworld .zip](https://github.com/OpenApoc/OpenApoc/files/6086803/save_Overworld.zip)

makus82:
Thx for reports guys 
and yes vehicles not moving cant be recoered by valkiry with cargo modul and agents
nothing helps..

makus82:
and position its seems its stuck in tunnel @EmperorLol 
![image](https://user-images.githubusercontent.com/11581207/110796041-331e9880-8280-11eb-8fdf-afab0a583f61.png)

admiral-crunch:
I've had this happen as well, you can open and edit the vehicles.xml file within the save file to delete the pathing.

sot010174:
Just happened to me with four different vehicles. What I've done is switch the vehicles.xml with an older (and acceptable) version. Just extract both saves - the one with stuck vehicles and the older one using 7-zip or something preserving paths. Swap the vehicles.xml file and zip all the files back and rename the extension to .save. Done. This may open the door to cheating (spawning downed ufo missions).

sot010174:
Just happened to me. I've managed to recover the vehicles by replacing the entries on the vehicles.xml with those from an older savegame (where the vehicles were at the base) . This workaround seems to work fine. Just unzip (the .save files are actually zip files) the affected save and an older one, open the vehicles.xml file on your prefered editor and replace the entries for the stuck vehicles (from <entry> to </entry>). Backup the saves before attempting because editing the files wrongly will crash the game.

Artytoo:
Is a "Mcgyver it" mod a possibility where it edits the AI pathing information of the selected vehicle?

AI Pathing is always going to be a nightmare as I have watched for a very long time.

dethstryke01:
[...]

PR #1687, opened 2026-09-12
City: stop idle and already-finished vehicle missions locking the fastest speed (#989)

### LINKED ISSUE #989: ①My Retiliator 4 stuck and　② I cant fast forward (x5)

andremmfaria:
Fixes #989.

Three commits, reviewable separately.

## What actually locks the speed

`GameState::canTurbo()` (`gamestate.cpp:950-977`) has three independent blockers, any one of which disables the fastest speed, and none of them has a timeout or a liveness check: a projectile in flight, a live hostile aggressive vehicle in the city, or a pending `AttackBuilding`/`AttackVehicle` mission.

The thread has always framed this as a consequence of a vehicle being stuck. That framing does not survive contact with the savegames. Loading all five attached to this issue and evaluating each blocker gives:

| save | canTurbo() | blocker (1) | blocker (2) hits | blocker (3) hits |
|---|---|---|---|---|
| save_city2 | false | clear | 64 | 1 |
| save_speed_locked | false | clear | 8 | 7 |
| save_TACTICAL | false | **fires**, 4 projectiles | 32 | 4 |
| saves/save_stuck | false | clear | 16 | 0 |
| save_103 | false | clear | 12 | 1 |

**Every one of the 132 blocker (2) hits is a vehicle with no missions at all** - ordinary parked fleets sitting at their organisation's home building (Cult of Sirius, Osiron, Psyke, Mutant Alliance). None is scenery-blocked or building-blocked, and none of the "stuck" vehicles people reported appears among them; those are Allied or Friendly, not Hostile. So blocker (2) fires simply because an organisation has become Hostile (relation below -50) and owns parked vehicles, which makes the lock permanent for the rest of that game.

Blocker (3) splits cleanly in two. Eight hits are genuine ongoing combat - target present, same city, `isFinished()` false - and all eight are NPC against NPC (Sensovision, Mutant Alliance and Psyke attacking Megapol). Five hits are stale: the target has moved to `CITYMAP_ALIEN` while the attacker is in `CITYMAP_HUMAN`, its `tileObject` is null, and **`isFinished()` already returns true**. `gamestate.cpp:1098-1108` even has a self-heal that clears exactly these on the next tick. `canTurbo()` is counting a mission the rest of the engine already considers finished.

## The fix

**Blocker (3)**: count a mission only if it is not already finished, reusing the engine's own check rather than a parallel copy.

```cpp
if ((m.type == VehicleMission::MissionType::AttackBuilding ||
     m.type == VehicleMission::MissionType::AttackVehicle) &&
    !m.isFinished(*this, *v.second, false))
```

`callUpdateIfFinished` is false, so this stays a pure read. This required dropping `const` from `canTurbo()`; all four call sites (`updateTurbo` and three in `cityview.cpp`) already hold non-const access. The alternative, an inline liveness test, would duplicate `isFinishedInternal`'s handling of `attackCrashed`, `crashed`, `sliding`, `falling`, `health == 0` and `canDamageBuilding` - a second copy that can silently drift from the real completion semantics, which is the same class of problem this issue is about.

**Blocker (2)**: a hostile aggressive vehicle holding no missions no longer blocks. An idle parked fleet is not something the player is being asked to react to; one actually executing a mission still blocks.

Blocker (1) is left alone. Projectiles resolve in seconds.

## Effect on the reported saves

| save | before | after | why |
|---|---|---|---|
| saves/save_stuck | false | **true** | its only blockers were 16 idle parked hostiles |
| save_103 | false | **true** | its one attack mission had a stale cross-city target |
| save_TACTICAL | false | false | four projectiles genuinely in flight |
| save_city2 | false | false | genuine live NPC combat |
| save_speed_locked | false | false | genuine live NPC combat |

`save_103` is the save behind the two videos in this thread showing the speed locked and then restored after the vehicle was moved, so that case now clears while every save with real combat stays correctly locked.

## Notes and limits

A player-ownership filter was considered and rejected on the data: the genuine-combat hits are all NPC-versus-NPC and the stale ones are all X-COM's, so filtering to player-owned vehicles would exempt real fights and keep blocking the stale missions - exactly backwards.

Two things this does not settle. `save_TACTICAL` and `saves/save_stuck` were expected to be cases where the speed still worked, per their reporters, but both evaluate to false at the moment they were saved; `canTurbo()` gates only the *fastest* speed, so "fast forward worked" may have meant speed 4. And since the stale missions self-heal on the next tick, a static save cannot explain why the lock reportedly persisted for minutes - either those saves were captured while paused, or there is a live sequence that has not been reproduced. Neither affects the change here, which is about `canTurbo()` not counting conditions that are already resolved or were never activity in the first place.

Separately, the stuck-vehicle behaviour itself (#968) is untouched. This only stops a stuck or idle vehicle from permanently disabling the fastest speed.


andremmfaria:
Two harnesses. `tests/test_city_turbo.cpp` is the committed regression test, wired into ctest and built entirely from constructed state so it needs no attached savegame. `tests/test_city_turbo_diag.cpp` is a manual diagnostic, deliberately not `add_test`'d, used to produce the per-save table in the description.

The regression test runs four scenarios, each starting from a verified `canTurbo() == true` baseline so only the fixture it sets up can change the result:

```cpp
// An idle hostile aggressive vehicle must NOT block: it is a parked fleet, not activity.
bool test_idle_hostile_aggressive_does_not_block(sp<GameState> state)
{
    resetToBaseline(*state);
    if (!expectCanTurbo(*state, true, "baseline-before-idle-hostile"))
        return false;

    auto type = makeTestVehicleType(*state, "VEHICLETYPE_TEST_IDLE", 1.0f);
    auto org  = makeHostileOrg(*state, "ORG_TEST_IDLE");
    auto v    = spawnVehicle(*state, type, org, testSpawnPosition(*state, 0));
    (void)v; // missions left empty deliberately

    return expectCanTurbo(*state, true, "idle-hostile-aggressive-no-missions");
}

// The same hostile aggressive vehicle WITH any mission at all must still block turbo -
// it is now actually doing something, so the player should be made aware.
bool test_hostile_aggressive_with_mission_blocks(sp<GameState> state) { /* ... */ }

// A live AttackVehicle mission against an in-city target must block.
// An AttackVehicle mission whose target has moved to another city must not.
```

Output on master `aed6f412` (fix reverted, test unchanged):
```
12: E [idle-hostile-aggressive-no-missions] expected canTurbo()==true, got false
12: E [baseline-before-busy-hostile]        expected canTurbo()==true, got false
12: E [baseline-before-live-attack]         expected canTurbo()==true, got false
12: E [baseline-before-stale-attack]        expected canTurbo()==true, got false
12: E test_city_turbo: one or more scenarios failed
1/1 Test #12: test_city_turbo ..................***Failed    0.99 sec
```

Note the three `baseline-before-*` failures: on master even the empty baseline cannot reach the fastest speed, because the test city already contains hostile organisations with parked vehicles. That is the reported bug reproduced in miniature, before any scenario fixture is added.

Output on this branch:
```
1/1 Test #12: test_city_turbo ..................   Passed    0.97 sec
```

Full suite is 12/12, the existing 11 plus this one.

The diagnostic harness was used to produce the five-save table in the description. It replicates `canTurbo()`'s three blockers and dumps, per vehicle, the owner, aggressiveness, relation to the player, crashed flag, position, mission list, and for attack missions the target's city, tile object and `isFinished(state, v, false)` result. It loads with a narrow `initCurrentCityOnly` rather than full `initState()`, because several of these saves are old enough that a full init trips over stale alien-city scenery references unrelated to this issue.

PR #1691, opened 2026-09-16
City: survive losing the last base on return from a battle (#1689)

### LINKED ISSUE #1689: Aborting a base defense mission segfaults on return to the city (base destroyed, current_base null in CityView::resume)
**Build:** master, code references against `3dc29d54`. Reproduced on a Linux RelWithDebInfo build of master and on a 2026-08-31 master build under gdb; the four functions involved are unchanged between them.

## What happens

Start a new campaign (or load any save with one base), trigger a base defense mission (the debug cheat is enough: F1, then `b`, "Spawning base defense mission"), accept the alert, enter the battle, then leave it through Escape, Exit Battle, confirm, Debriefing, OK. The process segfaults on the OK click, before the city view is drawn. Nothing is logged.

Related: #1690 (another battle-flow crash found in the same headless playtest).

## Backtrace

```
Thread 1 "OpenApoc" received signal SIGSEGV, Segmentation fault.
#0  OpenApoc::Label::setText (this=..., Text=<error reading variable: Cannot access memory at address 0x120>) at forms/label.cpp:87
#1  OpenApoc::CityView::resume (this=...) at game/ui/tileview/cityview.cpp:1856
#2  OpenApoc::CityView::CityView (this=..., state=...) at game/ui/tileview/cityview.cpp:1214
#3  ... std::make_shared<CityView> ...
#4  operator() (lambda) at game/ui/battle/battledebriefing.cpp:25
#5  OpenApoc::Control::triggerEventCallbacks (forms/control.cpp)
#6  OpenApoc::GraphicButton::eventOccured (forms/graphicbutton.cpp)
#7  OpenApoc::Control::eventOccured (forms/control.cpp)
#8  OpenApoc::BattleDebriefing::eventOccurred
#9  OpenApoc::Framework::processEvents (framework/framework.cpp)
#10 OpenApoc::Framework::run (framework/framework.cpp)
#11 main (game/main.cpp)
```

## Cause

1. Aborting a mission never sets `Battle::playerWon`, so [`Battle::exitBattle()`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/state/battle/battle.cpp#L2964) takes the "not won" branch of the `BaseDefense` case and treats the abort as a lost base: it sets `eventFromBattle = BaseDestroyed` and calls `building->base->die(state, false)` ([`game/state/battle/battle.cpp:3426-3443`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/state/battle/battle.cpp#L3426-L3443)).
2. [`Base::die()`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/state/city/base.cpp#L81) collapses the building, erases the base from `player_bases` and clears `state.current_base` ([`game/state/city/base.cpp:81`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/state/city/base.cpp#L81) onward). With a single base the player now has none.
3. The debriefing OK handler then unconditionally builds a new `CityView` ([`game/ui/battle/battledebriefing.cpp:21-26`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/ui/battle/battledebriefing.cpp#L21-L26)), whose constructor calls `resume()` ([`cityview.cpp:1214`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/ui/tileview/cityview.cpp#L1214)), and `resume()` reads `state->current_base->name` with no check ([`cityview.cpp:1856`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/ui/tileview/cityview.cpp#L1856)). `current_base` is an empty `StateRef`, so this dereferences null and `Label::setText` faults on the garbage reference.

The `GameLost` handling that exists for an empty base list never runs because the view construction crashes first.

## Expected

Either or both of:
- `exitBattle()` distinguishes an aborted base defense (no units lost, mission not fought to the end) from a lost one, instead of destroying the base on any non-win.
- The debriefing OK path checks for a missing `current_base` (or the pending `GameLost` event) before constructing `CityView`, and `CityView::resume()` tolerates an empty `current_base`.

andremmfaria:
Fixes #1689.

## Root cause

Aborting a base defense mission retreats every player unit (`Battle::abortMission`, `game/state/battle/battle.cpp:1984-1994`), and a retreated unit is excluded from `isConscious()`, so `Battle::checkMissionEnd` finds no conscious player units and sets `playerWon = false` even though nobody was killed. `Battle::exitBattle`'s `BaseDefense` case (`game/state/battle/battle.cpp:3426-3442`) treats any non-win as the base being overrun and calls `building->base->die(state, false)`, with no distinction between a genuine wipeout and a clean abort with zero losses.

`Base::die` (`game/state/city/base.cpp:143-159`) erases the base, clears `current_base`, and, when `player_bases` is now empty and `current_battle` is still set, sets `state.eventFromBattle = GameEventType::GameLost` and returns without pushing an actual `fw()` event yet. That deferred delivery is normally picked up by `GameState::update()` on the city's next tick (`game/state/gamestate.cpp:1086`), but `BattleDebriefing`'s OK handler builds a new `CityView` synchronously right after `Battle::exitBattle` (`game/ui/battle/battledebriefing.cpp:19-25`), and `CityView::resume()` unconditionally dereferences `state->current_base->name` to set the base-name label (`game/ui/tileview/cityview.cpp:1856` before this fix). With `current_base` cleared, that is a null-pointer dereference: `Label::setText` on a null `Base*`, segfaulting before the deferred `GameLost` ever gets a chance to show the defeat message.

## Fix

`CityView::resume()` and `CityView::refreshBaseView()` now skip the base-name label update when `current_base.id` is empty, so `CityView` survives construction on a null base instead of crashing.

That alone only stops the crash, so a first attempt at this fix also tried realizing the pending `GameLost` synchronously, from `BattleDebriefing`'s OK handler right after `Battle::exitBattle`. That was wrong: `Framework::processEvents()` dispatches every queued event to whatever stage is current at the time it drains the queue (`framework/framework.cpp`), and stage commands like the `REPLACEALL` to `CityView` are only applied after `processEvents()` returns for that frame. Pushing the `GameLost` event from inside the OK handler queues it while `BattleDebriefing` is still current, so it gets popped and silently dropped there before `CityView` ever exists to receive it, and the call also zeroed `gameTimeBeforeBattle`, permanently disarming the pre-existing deferred delivery in `GameState::update()` that would otherwise have caught it on a later tick.

The fix now realizes the pending update from `CityView::begin()` instead, guarded on `player_bases.empty() && gameTimeBeforeBattle.getTicks() != 0`. `StageStack::push()` calls `begin()` synchronously while applying the `REPLACEALL` (`framework/stagestack.cpp:7-15`), so by the time the event is queued, `CityView` is genuinely the current stage, and `Framework::processEvents()` dispatches it there on the next pump. `CityView::handleGameStateEvent()`'s existing `GameLost` branch (added for #766) then shows the DEFEAT message box and transitions to the main menu, unchanged from what it already did for a base lost outside of battle.

A normal battle return with bases left, and a base lost outside battle entirely, are both unaffected: the former still goes through `GameState::update()`'s existing deferred path untouched, and the latter never touches `gameTimeBeforeBattle` at all (`Base::die` pushes its event directly when there is no `current_battle`).

Both the abort-without-losses path and a genuine last-stand wipeout go through the same `Base::die` route, so this fix covers both, and a regression test exercises both.

Left alone on purpose: whether a lossless abort should forfeit the base at all is a separate, more opinionated question (every other mission type in that same `exitBattle` switch treats a non-win as simply no bonus, never as losing an asset), and it needs its own design call from maintainers rather than being folded into a crash fix.

## Verification

`ctest` is 16/16 on this branch (master's 15 plus this branch's new regression test), with every pre-existing test unchanged. A headless playtest (Xvfb, scripted input) reproduced the original crash against master, confirmed the fix as first attempted swallowed the defeat silently instead of crashing, and then confirmed the reworked fix end to end twice: an abort with zero losses and a real fought battle with a casualty both now show the DEFEAT message box on return and go cleanly to the main menu, with a fresh campaign afterward starting normally.


andremmfaria:
Harness for the before/after check, committed as `tests/test_base_defense_abort.cpp` and wired into ctest.

```cpp
// One real BaseDefense battle against the player's own (single, auto-created) base.
// Two cases: abort with zero losses (every unit retreats) and a genuine wipeout
// (every unit killed outright), both routing through the same Base::die() path.

Battle::beginBattle(*state, false, state->getAliens(), agents, aliens, guards, civilians, {},
                    baseBuilding);
Battle::enterBattle(*state);

// Case A: mirrors BUTTON_EXIT_BATTLE's Yes callback, zero units lost.
for (auto &u : playerUnits(*state))
    if (u->isConscious())
        u->retreat(*state);

// Case B: genuine wipeout, no abort involved.
// for (auto &u : playerUnits(*state)) { u->agent->modified_stats.health = 0; u->die(*state, nullptr, false); }

Battle::finishBattle(*state);
Battle::exitBattle(*state);

// State Battle::exitBattle() leaves behind once the last base is lost: this is the
// exact state CityView's constructor has to survive.
check(state->player_bases.empty(), "player_bases is empty after the base is lost");
check(state->current_base.id.empty(), "current_base.id is empty after the base is lost");
check(state->eventFromBattle == GameEventType::GameLost, "eventFromBattle == GameLost after the base is lost");

// CityView must come up on a null current_base without throwing.
auto cityView = mksp<CityView>(state);

// CityView::begin() is the hook StageStack::push() calls synchronously while applying
// the debriefing screen's REPLACEALL; it is the only place allowed to realize the
// pending after-battle update, and only once CityView is genuinely current.
check(state->gameTimeBeforeBattle.getTicks() != 0, "after-battle update still pending once CityView is merely constructed");
cityView->begin();
check(state->gameTimeBeforeBattle.getTicks() == 0, "CityView::begin() realized the pending after-battle update");

// Framework's own event queue is private, so feed the event begin() would have queued
// straight into the same handler that queue eventually dispatches to.
GameEvent gameLostEvent(GameEventType::GameLost);
bool handled = cityView->handleGameStateEvent(&gameLostEvent);
check(handled, "CityView::handleGameStateEvent() accepts a GameLost event");
check(state->messages.back().text == gameLostEvent.message(),
     "the GameLost message is logged, matching what begin()'s queued event would produce");
```

Output on master `d06cc102` (fix reverted):
```
Case: abort with zero losses
  PASS: loaded common and gamestate for the abort case
  PASS: exactly one player base after fillPlayerStartingProperty()
  PASS: Battle::beginBattle created a battle
  PASS: battle came up as BaseDefense
  PASS: player units spawned for the base defense battle
  PASS: spawned player unit has a tile position   [x25]
  PASS: player units present to retreat
  PASS: playerWon == false once every unit has retreated
```
```
Thread 1 "test_base_defen" received signal SIGSEGV, Segmentation fault.
OpenApoc::Label::setText (this=0x5555600b0c90, Text=<error reading variable: Cannot access memory at address 0x120>) at ../forms/label.cpp:87
#0  OpenApoc::Label::setText (...) at ../forms/label.cpp:87
#1  OpenApoc::CityView::resume (this=...) at ../game/ui/tileview/cityview.cpp:1856
#2  OpenApoc::CityView::CityView (this=..., state=...) at ../game/ui/tileview/cityview.cpp:1214
#3  std::_Construct<OpenApoc::CityView, ...> (...) at .../bits/stl_construct.h:133
...
#11 (anonymous namespace)::runAbortCase () at ../tests/test_base_defense_abort.cpp:193
#12 main (...) at ../tests/test_base_defense_abort.cpp:287
```

Output on this branch:
```
Case: abort with zero losses
  PASS: loaded common and gamestate for the abort case
  PASS: exactly one player base after fillPlayerStartingProperty()
  PASS: Battle::beginBattle created a battle
  PASS: battle came up as BaseDefense
  PASS: player units spawned for the base defense battle
  PASS: spawned player unit has a tile position   [x25]
  PASS: player units present to retreat
  PASS: playerWon == false once every unit has retreated
  PASS: player_bases is empty after the base is lost
  PASS: current_base.id is empty after the base is lost
  PASS: eventFromBattle == GameLost after the base is lost
  PASS: CityView constructed with an empty current_base
  PASS: after-battle update still pending once CityView is merely constructed
  PASS: CityView::begin() realized the pending after-battle update
  PASS: CityView::handleGameStateEvent() accepts a GameLost event
  PASS: the GameLost message is logged, matching what begin()'s queued event would produce once CityView is the current stage
Case: genuine wipeout (no abort)
  PASS: loaded common and gamestate for the wipeout case
  PASS: exactly one player base after fillPlayerStartingProperty()
  PASS: Battle::beginBattle created a battle
  PASS: battle came up as BaseDefense
  PASS: player units spawned for the base defense battle
  PASS: spawned player unit has a tile position   [x25]
  PASS: player units present to kill
  PASS: playerWon == false once every unit is dead
  PASS: player_bases is empty after the base is lost
  PASS: current_base.id is empty after the base is lost
  PASS: eventFromBattle == GameLost after the base is lost
  PASS: CityView constructed with an empty current_base
  PASS: after-battle update still pending once CityView is merely constructed
  PASS: CityView::begin() realized the pending after-battle update
  PASS: CityView::handleGameStateEvent() accepts a GameLost event
  PASS: the GameLost message is logged, matching what begin()'s queued event would produce once CityView is the current stage
ALL CHECKS PASSED
```

`ctest` is green on this branch, unchanged from master on every pre-existing test.

andremmfaria:
Headless playtest, before and after

ABORT MISSION confirmation with zero units lost:
Abort mission confirmation

The debriefing screen whose OK button used to crash the game:
Debriefing screen

The DEFEAT message box that now appears on the city view instead:
DEFEAT message box

The main menu reached cleanly after clicking OK on DEFEAT:
Main menu after defeat

A fresh campaign started from that main menu, with no stale state from the dead game:
Fresh campaign after defeat

The same DEFEAT flow after a real fought battle with a casualty, not just a lossless abort:
Defeat after a fought battle

The same abort sequence on a master build ends in the segfault shown in the harness comment above, before the city view is drawn.

PR #1692, opened 2026-09-16
Skirmish: refuse a second goToBattle while a battle load is pending (#1690)

### LINKED ISSUE #1690: Skirmish: pressing OK again while the battle is loading re-enters goToBattle and kills the process
**Build:** master, code references against `3dc29d54`. Linux RelWithDebInfo build.

## What happens

Open Skirmish Mode from the in-game options, pick a base map, and press OK on the Customize Forces screens more than once while the first press is still being processed (each press returns to another forces or equip screen, so pressing OK again is the natural thing to do). Every press calls `Skirmish::goToBattle()` again. The log shows the setup repeating:

```
W ... Skirmish::goToBattle(...) Adding new agents to base BASE_SKIRMISH
W ... Skirmish::goToBattle(...) Resetting base inventory
W ... Skirmish::goToBattle(...) Adding new agents to base BASE_SKIRMISH
W ... Skirmish::goToBattle(...) Resetting base inventory
(four times in total)
E static void OpenApoc::Battle::beginBattle(...) Battle::beginBattle called while another battle is in progress!
```

and the process dies right after that line, with no further output and no clean shutdown.

Related: #1689.

## Cause

- `SelectForces` forwards every `BUTTON_OK` click straight to `skirmish.goToBattle(...)` ([`game/ui/skirmish/selectforces.cpp:338-433`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/ui/skirmish/selectforces.cpp#L338-L433)), and `Skirmish` does the same for its own OK ([`game/ui/skirmish/skirmish.cpp:725-746`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/ui/skirmish/skirmish.cpp#L725-L746)). Neither disables the button or checks whether a battle is already being created.
- `goToBattle()` rebuilds `BUILDING_SKIRMISH` and `BASE_SKIRMISH` from scratch on every call ([`skirmish.cpp:301-330`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/ui/skirmish/skirmish.cpp#L301-L330)), replacing the map entries that the in-flight load already holds `StateRef`s to, then queues another `BattleBriefing` whose loader thread calls `Battle::beginBattle()` ([`skirmish.cpp:29-70`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/ui/skirmish/skirmish.cpp#L29-L70), [`skirmish.cpp:657-661`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/ui/skirmish/skirmish.cpp#L657-L661)).
- [`Battle::beginBattle()`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/state/battle/battle.cpp#L2449) detects the second call, logs the error and returns without creating a battle ([`game/state/battle/battle.cpp:2449-2453`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/state/battle/battle.cpp#L2449-L2453), [`game/state/battle/battle.cpp:2473-2477`](https://github.com/OpenApoc/OpenApoc/blob/3dc29d54/game/state/battle/battle.cpp#L2473-L2477)), but the loader and briefing continue as if a battle existed, and the process terminates.

## Expected

The skirmish setup screens should ignore or disable OK while a battle load is pending (or `goToBattle()` should return early when `state.current_battle` or a pending load exists), and a rejected `beginBattle()` should fail the load gracefully instead of letting the briefing proceed into an undefined state.

andremmfaria:
Fixes #1690.

## Root cause

`Skirmish::goToBattle()` rebuilds `BUILDING_SKIRMISH`, `BASE_SKIRMISH` and the whole agent roster from scratch every time it runs, then arms a single `loadBattle` closure and pushes the equip screen. Nothing stops it running again while an earlier call's battle build is still in flight, and `Skirmish::resume()` never clears `loadBattle` after firing it, so a later `resume()` replays it. A second OK press therefore submits a second `fw().threadPoolEnqueue` battle build against the same `GameState`, which has no locking around `current_battle`, `buildings`, `player_bases` or `agents`; the only mutex in `GameState` guards the object-id counter. The two background tasks then race on all of it.

`Battle::beginBattle` does notice the collision, with a plain check-then-set on `state.current_battle` and a `Battle::beginBattle called while another battle is in progress!` log line, but it returned `void`. Both skirmish loader lambdas ignored that and went straight on to write their skirmish bookkeeping through `state->current_battle`, which on a rejected call points at whichever battle did start. The losing load silently overwrote the winning battle's `skirmish` flag, score snapshot and relationship snapshot.

Underneath both of those, the per-frame stage command loop in `Framework::run()` copied `stageCommands`, iterated the copy, and then cleared the original, so anything queued while that batch was running was discarded. `StageStack::pop()` and `push()` call `finish()`, `pause()` and `resume()` synchronously from inside that loop, so a transition raised from one of those never happened: `SelectForces::resume()`'s self-POP and the `REPLACEALL` to `BattleBriefing` that `Skirmish::resume()` ends up raising both vanished. That is what leaves the setup screen on screen with a battle already loading behind it, and it is what invites the second OK press in the first place. On a master-equivalent binary a patient single-OK sequence never reaches the battle at all.

## Fix

`Skirmish` latches a flag as soon as a load is armed and `goToBattle()` returns early while that flag is set or while `state.current_battle` exists. The stage is replaced when the battle starts, so a `Skirmish` still on screen with the flag set means the load is still in flight and OK has to stay inert. `resume()` now moves `loadBattle` into a local and clears the member before invoking it, so a later `resume()` cannot replay it. Both the Enter-key path and `SelectForces`' OK button reach `goToBattle()`, so they are covered by the same guard.

Both `Battle::beginBattle` overloads now return `bool`: false when another battle is already in progress, and false when `createBattle` produced nothing. The building and vehicle skirmish loaders check it and bail out before the bookkeeping block, so a rejected load can no longer corrupt the battle that actually started. The other callers already assumed success and are unchanged; they ignore the return value and behave exactly as before.

The command loop now swaps the batch into the local instead of copying it. `stageCommands` is left empty, so a cascaded command appends to a container nothing is iterating and is processed on the next frame rather than thrown away, and the loop still never walks a container its own body can append to. This affects every screen in the game, not just skirmish, so the cascade sites were audited: `SelectForces::resume()`'s POP and `NotificationScreen::resume()`'s conditional POP now take effect, and `WeeklyFundingScreen::finish()`'s PUSH of the score screen now survives alongside the POP that uncovered it. Commands raised during a `REPLACEALL` or `QUIT` teardown are still dropped, deliberately: `StageStack::clear()` finishes every stage and resumes each one it uncovers on the way down, so those commands target a stack that has just been destroyed and would otherwise land on the replacement screen, for example popping a freshly pushed `BattleBriefing` and emptying the stack.

## Verification

`ctest` is 16/16 green on this branch, including a new `test_skirmish_reentrancy` that calls `goToBattle()` twice and checks the second call leaves the temporary building, base, agent roster and city references untouched, then checks `beginBattle` reports failure and preserves an in-progress battle's bookkeeping; its before and after output is in the next comment. A headless playtest compared this branch against a master-equivalent binary: on master a patient single-OK skirmish never leaves the Skirmish screen, and nineteen OK presses produced nineteen `Adding new agents to base BASE_SKIRMISH` lines and eighteen `another battle is in progress` errors without ever reaching a battle, while on this branch both the patient run and a burst of ten rapid OK clicks produced exactly one setup, zero collision errors, and a running battlescape. Aborting the battle returned cleanly to the city view and a second skirmish started afterwards, so the latch is per instance and does not leave a later screen inert. For the framework change specifically, the Weekly Funding Assessment now hands off to the score screen, which is `WeeklyFundingScreen::finish()`'s PUSH surviving the same frame as its own POP, and boot, difficulty, city, options, save, load and quit were all exercised without a stuck or doubled screen.

Two unrelated defects turned up during that playtest and are out of scope here. A base-defense skirmish segfaults reproducibly inside `BattleMap::createBattle` (`battlemap.cpp:239`), which walks every entry of `state.agents` and dereferences `agent.second->homeBuilding` with no null check while the skirmish roster contains agents that have none; that file is untouched by this branch and the guard behaved correctly on the way in, with exactly one `goToBattle` and no collision error. Separately, one run died during a single legitimate Alien Farm load with one `goToBattle`, no collision error and the log ending inside `Battle::initialMapCheck`; it did not reproduce under gdb and no backtrace was captured, so it is recorded as an open observation rather than a claim.


andremmfaria:
Harness for the before/after check, committed as `tests/test_skirmish_reentrancy.cpp` and wired into ctest next to `test_city_dimension_gate`; it links `OpenApoc_GameUI` as well, because it drives the real `Skirmish` stage headlessly rather than a stand-in.

Abridged core, with the per-assertion logging collapsed into a `check`:

```cpp
// difficulty0 loaded headless, then fillPlayerStartingProperty(). Target is the first
// alien-owned human-city building that has a battle map.
auto skirmish = mksp<Skirmish>(state);
skirmish->setLocation(target);

skirmish->goToBattle();

// The first call has to have run to completion: it is what arms the pending load, and the
// guard is only meaningful if this state actually got built.
auto firstBuilding = state->buildings["BUILDING_SKIRMISH"];
auto firstBase = state->player_bases["BASE_SKIRMISH"];
auto agentsAfterFirst = state->agents.size();

skirmish->goToBattle(); // the second OK press, while the first load is still pending

// Refused: same temporary building and base, no extra agents, and the city did not collect
// a second BUILDING_SKIRMISH reference.
check(state->buildings["BUILDING_SKIRMISH"] == firstBuilding);
check(state->player_bases["BASE_SKIRMISH"] == firstBase);
check(state->agents.size() == agentsAfterFirst);
check(countSkirmishBuildingRefs(city) == 1);

// beginBattle's half. A battle is already in progress, so the call must report failure.
state->current_battle = existing;
bool started = Battle::beginBattle(*state, false, state->getAliens(), agents, nullptr,
                                   nullptr, nullptr, noVehicle, target);

// Mirrors what the skirmish loaders do once a battle really has started. Before beginBattle
// reported failure this block ran unconditionally.
if (started)
{
    state->current_battle->skirmish = true;
    state->current_battle->scoreBeforeSkirmish = state->totalScore.tacticalMissions;
}

check(!started);
check(state->current_battle == existing);
check(!existing->skirmish); // the rejected load must not touch the winner's bookkeeping
```

Output on master `2574d806` (fix reverted):
```
W Skirmish::goToBattle Adding new agents to base BASE_SKIRMISH
W Skirmish::goToBattle Resetting base inventory
W Skirmish::goToBattle Adding new agents to base BASE_SKIRMISH
W Skirmish::goToBattle Resetting base inventory
E test_second_gotobattle_is_refused Second goToBattle() replaced BUILDING_SKIRMISH while a load was pending
E main test_skirmish_reentrancy failed - 1 test(s) failed
```

The doubled `Adding new agents` / `Resetting base inventory` pair is the same shape the issue's own log repeats ten times over. With only `beginBattle` reverted, the second half fails on its own:
```
E Battle::beginBattle Battle::beginBattle called while another battle is in progress!
E test_beginbattle_refuses_while_a_battle_is_in_progress A rejected battle load clobbered the in-progress battle's skirmish bookkeeping
E main test_skirmish_reentrancy failed - 1 test(s) failed
```

Output on this branch:
```
1/1 Test #16: test_skirmish_reentrancy .........   Passed    1.03 sec
```

One caveat on that second leg: `beginBattle` returning `void` on master means the `bool started = ...` line does not compile there, so running it required dropping the return value and leaving the bookkeeping block unconditional, which is exactly the shape `loadBattleBuilding` had before this branch.

andremmfaria:
Headless playtest, master versus this branch.

On master, a patient single-OK sequence (Skirmish OK, equip OK, wait ten seconds) never leaves the Skirmish screen: the transition to the briefing was raised during a pop cascade and discarded.

master, stuck on the Skirmish screen

Every further OK re-enters goToBattle and lands on another Equip Agent screen with a duplicated roster and an empty inventory bar. This repeats for as long as the player keeps pressing.

master, second equip screen with a duplicated roster

```
=== master-equivalent binary (OpenApoc-1689, skirmish+framework code == master) ===
Patient sequence: Skirmish OK -> equip OK -> wait 10s = still on the Skirmish screen (1690-01).
Every further OK re-enters goToBattle. After 19 presses:
  19x Skirmish::goToBattle 'Adding new agents to base BASE_SKIRMISH'
  18x Battle::beginBattle 'called while another battle is in progress!'
  battle never reached; process stayed alive this run (crash is race-dependent).

249:W void OpenApoc::Skirmish::goToBattle(...) Adding new agents to base BASE_SKIRMISH
250:W void OpenApoc::Skirmish::goToBattle(...) Resetting base inventory
302:W void OpenApoc::Skirmish::goToBattle(...) Adding new agents to base BASE_SKIRMISH
303:W void OpenApoc::Skirmish::goToBattle(...) Resetting base inventory
304:E static void OpenApoc::Battle::beginBattle(...) Battle::beginBattle called while another battle is in progress!
317:W void OpenApoc::Skirmish::goToBattle(...) Adding new agents to base BASE_SKIRMISH
318:W void OpenApoc::Skirmish::goToBattle(...) Resetting base inventory
319:E static void OpenApoc::Battle::beginBattle(...) Battle::beginBattle called while another battle is in progress!
```

On this branch the same patient sequence reaches Mission Briefing with both confirm buttons visible.

this branch, Mission Briefing reached

Ten rapid OK clicks plus ten rapid Returns run straight through briefing and squad assignment into a live battle, with exactly one battle setup in the log.

this branch, battle running after OK spam

The patient run reaches the same place: units on the map, battle clock advancing, hostile unit spotted.

this branch, battle running

The Customise Forces route also works, including spamming OK on the Customize Forces screen itself: SelectForces pops correctly, the equip screen runs and the briefing appears.

this branch, briefing via Customise Forces

Unrelated to skirmish, the Weekly Funding Assessment now hands off to the score screen: its finish() PUSH and its own POP are raised in the same frame, which is the exact pattern the framework change restores.

weekly funding assessment handing off to the score screen

PR #1693, opened 2026-09-16
City: two-way roads with lanes, queueing, overtaking and collisions (#785)

### LINKED ISSUE #785: [city] Civilian (all) road vehicles from differ directions moves by one side of road
Screenshot from 2019-10-12 21-42-40

andremmfaria:
Fixes #785.
Refs #264 (items 19 and 20).

## Root cause

Ground vehicles are point objects driven from tile centre to tile centre, with no concept of a lane and no awareness of each other.

Every ground hop targets `Tile::getRestingPosition()` (`tile.cpp:29`), the exact tile centre, so two vehicles travelling in opposite directions down the same road share one line and pass through one another. `GroundVehicleTileHelper::canEnterTile` (`vehiclemission.cpp:3290`) checks scenery connections and hill flags only and never looks at other vehicles, unlike the flying helper, so a fast vehicle also drives straight through the back of a slow one. `Vehicle::handleCollision` (`vehicle.cpp:2654`) has a projectile branch and nothing else, so contact costs nothing either.

The tile data cannot supply lanes. A road tile carries `tile_type`, `road_type` (`StraightBend`, `Junction`, `Terminal`), a four way `connection` and a four way `hill` flag (`scenerytiletype.h`), and the road graph is undirected. There is no direction of travel anywhere in the scenery or the vehicle serialization, so a lane has to be derived at run time from the direction of the hop being taken.

## Changes

Four commits, each independently revertable and each gated so that with its option off the behaviour is what it is today.

**Lanes and driving side.** `OpenApoc.NewFeature.TwoWayRoads` (default on) and `OpenApoc.NewFeature.DriveOnLeft` (default off, so right hand traffic). Each hop's destination is nudged sideways from the tile centre by a quarter of a tile, on the side matching that hop's direction, so opposing traffic travels on two distinct lines. A quarter tile is small enough that the tile a vehicle is considered to occupy and every arrival check are unaffected. Building entrances and exits (`Terminal` road tiles) and off road ATV movement stay on the centre line, and the offset is skipped for any vehicle wider than one tile.

**Holding behind same-lane traffic.** Before committing to its next hop a ground vehicle looks 1.5 tiles ahead along its own direction of travel and holds position if another ground vehicle is there within 0.4 tiles sideways. The look ahead is derived, not picked: a follower released at distance L arrives L - 1 + slow/fast behind, and the gap has to stay at or above 1.0 for the two never to share a tile, which gives L >= 1.43 at the 4:7 speed ratio of the slowest and fastest road vehicles. The direction test uses velocity rather than facing, because a parked vehicle keeps whatever facing it last had; zero velocity means an obstacle and always blocks, non-zero blocks only on a positive dot product, so oncoming and crossing traffic are ignored. Wrecks are driven through exactly as before.

This cannot deadlock. A vehicle only waits on traffic strictly ahead of it along its own heading, so on a straight the wait relation is a strict order and whoever is at the head of a queue always has clear road. Vehicles crossing at a junction see a zero dot product and never wait on each other, oncoming vehicles see a negative one. The only cycle left is a ring of bumper to bumper traffic closing on itself, so the wait is also capped at 8 game seconds, comfortably above the 432 tick worst case legitimate wait (the slowest road vehicle clearing 1.5 tiles). That cap is what stops a parked or stuck blocker from killing a lane.

**Overtaking.** A vehicle held for 0.5 game seconds pulls out and passes. The threshold has to sit below the routine per tile wait of roughly 124 ticks, or a faster vehicle would pace the slower one forever and never pass. It only passes when it is genuinely faster and when the next four hops are an unbroken straight of `StraightBend` road tiles with nothing coming the other way; four hops covers the vehicle being passed plus the ground it covers during the pass. Those hops are driven with the lane offset mirrored, then it merges back. The speed comparison uses `Vehicle::getSpeed()` rather than `type->top_speed`, because `top_speed` is 0 for every road vehicle type in the base gamestate and all of their speed comes from the engine, so a type level comparison would never be true.

**Collisions.** `OpenApoc.NewFeature.VehicleCollisions`, default off. Two ground vehicles whose centres come within 0.4 tiles damage both sides for `min(10, closingSpeed * 0.5)`, with a one game second cooldown per vehicle so one contact costs one hit rather than one per tick. The contact distance sits below the 0.5 tile gap the lane offsets put between opposing lanes, so correctly separated traffic never touches; what is left to collide is head on traffic with lanes off, vehicles crossing at a junction, and a pass that goes wrong. Damage goes through the existing `applyDamage` path, so a vehicle driven far enough down crashes the way it already does. Detection reuses the tile object intersection bookkeeping rather than adding a spatial query. Faction relations are untouched: a collision is an accident, not an attack.

## Serialization

One new serialized field, `VehicleMission::overtakeTilesRemaining`, added to the `VehicleMission` object in `gamestate_serialize.xml`. Without it a save taken mid pass reloads with the vehicle merging back into the lane it is halfway past, on top of the vehicle it was overtaking.

The hold timer (`VehicleMission::trafficHoldStartTicks`) and the contact cooldown (`Vehicle::ticksContactAvailable`) are deliberately not serialized. Reloading mid hold restarts a couple of seconds of standing still, and reloading mid contact costs at worst one extra hit. No save version bump, and old saves converge within one tile of movement.

## Verification

`ctest` is green on this branch: 16 tests, master's 15 plus `tests/test_two_way_roads.cpp`. That test boots the real city, finds a straight road run using the game's own ground passability helper, places vehicles and drives them with real `gotoLocation` missions, and covers all four slices. Every case was also run with its own slice's sources reverted and fails there. Details and before/after output in the harness comment below.

Headless playtest of a fresh medium campaign, one save reloaded for every scenario with `SeedRng` on, so the frames are pixel comparable rather than anecdotal:

- Options surface correctly, all three under `Unsupported ground vehicles crash` in the cityscape list, defaults as stated, and `DriveOnLeft` persists across an OK and reopen.
- Lanes are visible on live civilian traffic with defaults, and the same save at the same in game second with `TwoWayRoads=0` puts the same vehicles back on the yellow centre line. That before/after pair is the direct evidence that the offset is live and is a quarter tile.
- `DriveOnLeft=1` at the same second mirrors both vehicles to the other side and changes nothing else.
- Save and load is clean: three loads, city intact, traffic resumes with no snapping and no stuck vehicle. No new log error or warning against the baseline run.

Not observed live: no hold, no pass and no contact happened in any observation window, because traffic density over the game time this harness could simulate never put two ground vehicles in the same lane close enough to trigger one. An X-COM ground vehicle order was not issued either, since map click input under that harness opens the building screen. The unit test remains the only evidence that queueing, overtaking and contact damage fire.

## Known limits

- The collision damage numbers are invented. No primary source describes contact damage between road vehicles in the original game, so they are chosen to be survivable rather than measured. That is exactly why `VehicleCollisions` ships off: the mechanism is here and tested, and someone with the original game can tune it and flip the default.
- A pass commits once it starts and does not re-check the opposing lane while it is under way, so oncoming traffic that appears after the window was measured is not avoided. With collisions on, that now costs damage.
- No city wide soak. Everything measured is two vehicles on a hand picked straight; how often passes fire in dense traffic, and whether the 8 second cap is ever reached there, is unmeasured.
- Turbo interaction is reasoned rather than tested. Game time jumps far past the 8 second cap in a single turbo update, so holds expire immediately, but no turbo scenario was built to confirm it.


andremmfaria:
Harness for the before/after checks, committed as `tests/test_two_way_roads.cpp` and wired into ctest alongside `test_serialize` and `test_city_dimension_gate`, since unlike a comparative measurement these are four pass/fail regression gates.

```cpp
// Every case reloads the gamestate, so none of them inherits the previous case's
// traffic or game time.
sp<GameState> loadState();

// Road-run discovery. Uses the game's own ground passability helper rather than
// reimplementing road connectivity, so the run found here is one real pathfinding
// would also accept. straightTilesOnly additionally demands StraightBend tiles,
// which is what an overtake window needs.
bool findStraightRoadRun(GameState &state, int minLength, StraightRun &out,
                         bool straightTilesOnly = false)
{
	GroundVehicleTileHelper helper(map, VehicleType::Type::Road);
	for (/* every tile, both axes */)
	{
		while ((int)run.size() < minLength)
		{
			if (!helper.canEnterTile(map.getTile(cur), map.getTile(next))) break;
			run.push_back(next);
		}
		if ((int)run.size() >= minLength) { out.tiles = run; out.dir = dir; return true; }
	}
	return false;
}

// Distance from the tile centre, perpendicular to the direction of the run: this is
// the lane offset, read straight off the live vehicle position.
float lateralOffset(const Vec3<float> &position, const Vec3<int> &tile, const Vec3<int> &dir)
{
	return dir.x != 0 ? position.y - (tile.y + 0.5f) : position.x - (tile.x + 0.5f);
}

// Case 1, lane separation. Two Autotrans sent down one straight run in opposite
// directions; record each one's lateral offset as it crosses the middle tile.
void testLaneSeparation()
{
	auto forward = city->placeVehicle(*state, vType, player, tile(first)->getRestingPosition());
	auto backward = city->placeVehicle(*state, vType, player, tile(last)->getRestingPosition());
	forward->setMission(*state, VehicleMission::gotoLocation(*state, *forward, last));
	backward->setMission(*state, VehicleMission::gotoLocation(*state, *backward, first));
	for (int i = 0; i < 400 && !(sawForward && sawBackward); i++)
	{
		state->update(TICKS_PER_SECOND / 4);
		if ((Vec3<int>)forward->position == midTile)
			forwardLateral = lateralOffset(forward->position, midTile, run.dir);
		// ... same for backward
	}
	check(std::abs(forwardLateral) > 0.1f && std::abs(backwardLateral) > 0.1f, "...");
	check((forwardLateral > 0.0f) != (backwardLateral > 0.0f), "...");
}

// Shared by cases 2 and 3. A slow leader (Autotrans, speed 4) one tile in front of a
// faster follower (Blazer Turbo Bike, speed 7), both sent to the far end of the run.
// Road vehicle types carry no top_speed of their own, all of it comes from the engine,
// so the pair is chosen by name and the speed gap checked once placed.
FollowerRun runFollowerScenario(bool twoWayRoads, int runLength, int updates,
                                bool straightTilesOnly = false)
{
	config().set("OpenApoc.NewFeature.TwoWayRoads", twoWayRoads);
	// ... place follower at run.tiles[0], leader at run.tiles[1], both goto run.tiles.back()
	for (int i = 0; i < updates; i++)
	{
		state->update(TICKS_PER_SECOND / 4);
		// Once either parks on the shared destination the other arrives on top of it,
		// which says nothing about how traffic behaves en route.
		if (follower->missions.empty() || leader->missions.empty()) break;

		float followerLateral = lateralOffset(follower->position, followerTile, run.dir);
		float leaderLateral = lateralOffset(leader->position, leaderTile, run.dir);
		bool opposingLanes = followerLateral * leaderLateral < -0.01f;
		bool leaderMoving = glm::length(leader->velocity) > 0.001f;

		if (opposingLanes) result.opposingLaneSamples++;
		else if (result.opposingLaneSamples > 0 && followerLateral * leaderLateral > 0.01f)
			result.mergedBack = true;
		// Sharing a moving leader's tile only counts as a failure to queue when both are
		// on the same side of the road; during a pass the follower is meant to be there.
		if (followerTile == leaderTile && leaderMoving && !opposingLanes)
			result.sameLaneOverlaps++;
		if (gap.x * axis.x + gap.y * axis.y > 0.5f) result.followerAhead = true;
	}
	return result;
}

// Case 2, en-route blocking. Option on: never shares the moving leader's tile in its
// lane. Option off: does, which is master's behaviour.
void testEnRouteBlocking()
{
	auto on = runFollowerScenario(true, 12, 600);
	check(on.closeSamples > 0, "the follower actually caught the leader");
	check(on.sameLaneOverlaps == 0, "never shares the moving leader's tile in its lane");
	auto off = runFollowerScenario(false, 12, 600);
	check(off.sameLaneOverlaps > 0, "with the option off the follower drives through the leader");
}

// Case 3, overtaking. Same setup on a longer StraightBend-only run.
void testOvertaking()
{
	auto on = runFollowerScenario(true, 16, 900, true);
	check(on.opposingLaneSamples > 0, "the follower used the opposing lane");
	check(on.mergedBack, "the follower merged back into its own lane");
	check(on.followerAhead, "the follower finished the pass in front of the leader");
}

// Case 4, collisions. Two Autotrans head on down one centre line with lanes off, which
// is the simplest way to make two ground vehicles actually touch. Count health drops.
HeadOnRun runHeadOnScenario(bool collisions)
{
	config().set("OpenApoc.NewFeature.TwoWayRoads", false);
	config().set("OpenApoc.NewFeature.VehicleCollisions", collisions);
	// ... place at opposite ends, send each to the other's tile
	for (int i = 0; i < 400; i++)
	{
		state->update(TICKS_PER_SECOND / 4);
		if (forward->getHealth() < forwardHealth) { result.forwardHits++; /* ... */ }
		if (backward->getHealth() < backwardHealth) { result.backwardHits++; /* ... */ }
	}
	return result;
}

void testVehicleCollisions()
{
	auto on = runHeadOnScenario(true);
	check(on.forwardHits == 1 && on.backwardHits == 1,
	      "both vehicles are damaged exactly once by the one contact");
	auto off = runHeadOnScenario(false);
	check(off.forwardHits == 0 && off.backwardHits == 0,
	      "with the option off neither vehicle is damaged");
}
```

Output with slice 1 reverted (that leg ran the same check before it was committed under its current name, hence the different message prefix):
```
runLaneSeparationCheck: forward lateral = 0, backward lateral = 0
Expected non-zero lane offsets, got forward=0 backward=0
exit=1
```

Output on this branch:
```
lane separation: forward lateral = 0.25, backward lateral = -0.25
PASS: lane separation: both vehicles are off the tile centre line
PASS: lane separation: opposing vehicles use opposite lanes
```

Output with slice 2 reverted:
```
en-route blocking: TwoWayRoads on -> sameLaneOverlaps = 11, closeSamples = 32
FAIL: en-route blocking: the follower never shares the moving leader's tile in its lane
en-route blocking: TwoWayRoads off -> sameLaneOverlaps = 11, closeSamples = 32
exit=1
```

Output on this branch:
```
follower scenario: leader speed 4, follower speed 7
en-route blocking: TwoWayRoads on -> sameLaneOverlaps = 0, closeSamples = 33
PASS: en-route blocking: the follower actually caught the leader
PASS: en-route blocking: the follower never shares the moving leader's tile in its lane
en-route blocking: TwoWayRoads off -> sameLaneOverlaps = 11, closeSamples = 32
PASS: en-route blocking: with the option off the follower drives through the leader
```

The option-off column reads 11 in both legs, which is the point: with `TwoWayRoads` off this branch reproduces master exactly.

Output with slice 3 reverted:
```
overtaking: TwoWayRoads on -> opposingLaneSamples = 0, mergedBack = false, followerAhead = false
FAIL: overtaking: the follower used the opposing lane
FAIL: overtaking: the follower merged back into its own lane
FAIL: overtaking: the follower finished the pass in front of the leader
exit=3
```

Output on this branch:
```
overtaking: TwoWayRoads on -> opposingLaneSamples = 19, mergedBack = true, followerAhead = true
PASS: overtaking: the follower used the opposing lane
PASS: overtaking: the follower merged back into its own lane
PASS: overtaking: the follower finished the pass in front of the leader
```

Output with slice 4 reverted:
```
collisions: VehicleCollisions on -> forwardHits = 0, backwardHits = 0
FAIL: collisions: both vehicles are damaged exactly once by the one contact
collisions: VehicleCollisions off -> forwardHits = 0, backwardHits = 0
exit=1
```

Output on this branch:
```
collisions: VehicleCollisions on -> forwardHits = 1, backwardHits = 1
PASS: collisions: both vehicles are damaged exactly once by the one contact
collisions: VehicleCollisions off -> forwardHits = 0, backwardHits = 0
PASS: collisions: with the option off neither vehicle is damaged
```

Each reverted leg restored only that slice's own source files, kept the test as committed, and was rebuilt and rerun before the slice was put back; `ctest` is green on the branch (16 tests after the rebase onto `d06cc102`) and `test_serialize_difficulty0` passes with the new serialized field.

andremmfaria:
Headless playtest, same save, same in-game second

Every run below reloads one save taken from a fresh medium campaign with `SeedRng` on, so frame N of any two runs carries the same in-game clock and the frames are directly pixel comparable rather than anecdotal.

The three new options in the cityscape list, directly under `Unsupported ground vehicles crash`, with their shipping defaults: `TwoWayRoads` on, `DriveOnLeft` and `VehicleCollisions` off.
Options list

Defaults, 12:45:29. A civilian car and a civilian bike each sit clearly off the yellow centre line, on the side matching their direction of travel.
Lanes on

Same frame, cropped to the two vehicles.
Lanes on, crop

Same save, same camera, same 12:45:29, with `TwoWayRoads=0`. Both vehicles are back on the centre line, the car straddling the dashes.
Lanes off

Same frame, cropped. Side by side with the crop above, this pair is the direct evidence that the offset is live and is a quarter tile.
Lanes off, crop

Same save, same camera, same 12:45:29, with `DriveOnLeft=1`. Both vehicles are mirrored to the other side of the centre line and nothing else about their positions changed.
Drive on left

